PR #4, opened 2019-09-17
Implemented player skins

Sandertv:
This merge request adds API to get the skin from a player. It provides a Skin and Cape type, both of which implement the image.Image interface so that they may be directly encoded as an image.


Saiini:
Looks good to me 👍

PR #6, opened 2019-09-19
Created basic scoreboard sending and removing

Saiini:
Basic scoreboard sending and removing, this has no API at the moment

### DIFF
--- a/dragonfly/session/session.go
+++ b/dragonfly/session/session.go
@@ -171,6 +171,7 @@ func (s *Session) SendScoreboard(displayName string, objName string){
 		SortOrder:     0,
 	})
 }
+
 // RemoveScoreboard ...
 func (s *Session) RemoveScoreboard(objName string){
 	s.writePacket(&packet.RemoveObjective{
--- a/dragonfly/session/session.go
+++ b/dragonfly/session/session.go
@@ -161,6 +161,23 @@ func (s *Session) SendJukeBoxPopup(message string) {
 	})
 }
 
+// SendScoreboard ...
+func (s *Session) SendScoreboard(displayName string, objName string){
+	s.writePacket(&packet.SetDisplayObjective{
+		DisplaySlot:   "sidebar",
+		ObjectiveName: objName,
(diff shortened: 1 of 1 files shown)


dragonfly/session/session.go
```diff
@@ -161,6 +161,23 @@ func (s *Session) SendJukeBoxPopup(message string) {
...
 }
 
+// SendScoreboard ...
+func (s *Session) SendScoreboard(displayName string, objName string){
+	s.writePacket(&packet.SetDisplayObjective{
+		DisplaySlot:   "sidebar",
+		ObjectiveName: objName,
+		DisplayName:   displayName,
+		CriteriaName:  "dummy",
+		SortOrder:     0,
```
Sandertv: You can remove this line due to the default value. Also, have you checked if the "dummy" criteria name is required?

dragonfly/session/session.go
```diff
@@ -161,6 +161,23 @@ func (s *Session) SendJukeBoxPopup(message string) {
...
+	})
+}
+// RemoveScoreboard ...
```
Sandertv: Please add a newline before this comment.

dragonfly/session/session.go
Saiini: looking at other implementations - they used the dummy criteria name, hence why i added it there

Sandertv: Yeah I understand. Have you tried to see if it works when keeping this field empty as well?

Saiini: Just tried it, removing the sort order does nothing and allows the scoreboard to be displayed, but having the criteria name to nothing (or removing it entirely) doesn't display the scoreboard at all

dragonfly/session/session.go
Saiini: fixed

dragonfly/session/session.go
Sandertv: Okay, thanks for checking 👍

PR #7, opened 2019-09-28
Implemented basic chunks and chunk sending

Sandertv:
This merge request adds a `chunk` package which implements chunks. Chunks themselves are rather low-level abstract structures: They don't work directly with blocks, but instead only with block runtime IDs.

The main structure of a chunk is as follows:
Chunk:
    0-16 Sub chunks:
        0-256 Block storages:
            Runtime ID palette
            Pointers to the palette, representing blocks

The block storages of each sub chunk hold a palette, which in turn holds all block runtime IDs used in that storage. The storage then has 4096 pointers to runtime IDs in that palette. The size of these pointers varies, depending on how many unique blocks there are in the palette. They will have a size of either 1, 2, 3, 4, 5, 6, 8 or 16 bits, where the sizes 3, 5 and 6 have additional padding.

### DIFF
--- a/dragonfly/session/session.go
+++ b/dragonfly/session/session.go
@@ -1,8 +1,10 @@
 package session
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/dragonfly-tech/dragonfly/dragonfly/player/chat"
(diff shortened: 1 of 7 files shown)


Saiini:
Wow... Awesome job sander! 😄

PR #8, opened 2019-09-30
Chunk loading and basic blocks

Sandertv:
This merge request implements the dynamic loading of chunks using generators. A basic block implementation is added which is based around blocks and block encoders, where the encoders translate Go types+values to ID, meta and NBT values.


Saiini:
All looks good, thanks =D

Saiini:
But, world generation doesn't seem to work for me

PR #10, opened 2019-10-13
Implemented player viewing, including moving around and the player list

Sandertv:
This merge request implements player viewing and the moving around of players. It also moves around some of the code in session.go to other files so that relevant code can be found more easily.


Saiini:
Looks good, thanks =D

PR #13, opened 2019-10-24
World I/O implementation

Sandertv:
This merge request implements world I/O. Dragonfly now uses the current world format used by Minecraft, so single-player worlds may be loaded by Dragonfly.
In addition to the world I/O implementation, this PR fixes a couple of hidden memory leaks and cleans up the shutdown of the server when using ctrl+c.


Saiini:
All looks fine

PR #14, opened 2019-10-25
Time implementation

Sandertv:
This PR implements time in Dragonfly: There is now API for time changing and stopping, and Dragonfly will, by default, have a working day-night cycle. These properties are also saved to the provider.


Saiini:
Hot

PR #19, opened 2020-06-30
Handle PlayerSkin packet

TwistedAsylumMC:
I have tested by myself and there are no issues when changing skin, this just needs to be pushed to another branch so it can be tested on the dev server with multiple people.


TwistedAsylumMC:
Closing since I now have push permissions and requires more changes

PR #28, opened 2020-07-13
[Feature] Kelp

DenielWorld:
Kelp block & item introduced, somewhat functional.

Existing Problems:
* Kelp only waterlogs at times. [FIXED]
* Breaking a kelp block also removes its liquid block. [FIXED]
* Breaking a kelp block at the bottom does not break upper kelp blocks. [FIXED]
* Breaking a kelp block in survival dupes the existing kelp in your inventory. [FIXED]


dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+	return 0 //Kelp grows without any light.
+}
+
+func (Kelp) RequiresHydration() bool {
+	return false //Kelp does not need hydration but rather just has to be placed in water.
+}
+
+func (Kelp) RequiresFarmland() bool {
+	return false //Kelp doesn't require a farmland, literally grows underwater.
+}
```
Sandertv, line 46: You can remove all this.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"github.com/df-mc/dragonfly/dragonfly/entity/physics"
+	"github.com/df-mc/dragonfly/dragonfly/item"
+	"github.com/df-mc/dragonfly/dragonfly/world"
+	"github.com/go-gl/mathgl/mgl64"
+	"math/rand"
+	"time"
+)
+
+type Kelp struct {
```
Sandertv, line 12: Make sure to document the struct.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+
+	//0-15, defines the age of the kelp block. If age is 15, kelp won't grow any further.
```
Sandertv, line 14: Make sure to start documentation with the name of the field/method/type and add a space after the `//`.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+func (Kelp) SideClosed(pos, side world.BlockPos, w *world.World) bool {
+	return false //Kelp can always be flowed through.
+}
+
+func (Kelp) AABB(world.BlockPos, *world.World) []physics.AABB {
+	return nil //Kelp can be placed even if someone is standing on its placement position.
+}
+
+//In Java Edition, the age value can be up to 25, but MCPE limits it to 15.
+func (k Kelp) SetRandomAge() Kelp {
```
Sandertv, line 62: Should make this an unexported method as it's quite useless outside of this package.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+	k.Age = rand.Intn(14)
+	return k
+}
+
+//The age must be 0-15.
+func (k Kelp) SetAge(age int) Kelp {
```
Sandertv: This method can probably be removed too, I don't think it's needed provided the field is exported.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (k Kelp) UseOnBlock(pos world.BlockPos, face world.Face, _ mgl64.Vec3, w *world.World, user item.User, ctx *item.UseContext) (used bool) {
+	pos, _, used = firstReplaceable(w, pos, face, k)
+	if !used {
+		return
+	}
+
+	blockBelow, _ := w.Block(pos.Add(world.BlockPos{0, -1})).EncodeBlock()
+	if blockBelow == "minecraft:air" || blockBelow == "minecraft:water" || blockBelow == "minecraft:flowing_water" { //Kelp cannot be placed on top of some blocks.
```
Sandertv, line 81: Please avoid this, just check against block types.
```go
switch w.Block(pos.Add(world.BlockPos{0, -1})).(type) {
    case Air, Water: return
}
```
Note: We need to add some sort of solid interface for this as well.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+
+// NeighborUpdateTick ...
+func (k Kelp) NeighbourUpdateTick(pos, changed world.BlockPos, w *world.World) {
+	if changed.Y()-1 == pos.Y() { //When a kelp block is broken above, the kelp block underneath it gets a new random age.
+		w.SetBlock(pos, k.SetRandomAge())
+	}
+
+	blockBelow, _ := w.Block(pos.Add(world.BlockPos{0, -1})).EncodeBlock()
+
+	if blockBelow == "minecraft:air" || blockBelow == "minecraft:water" || blockBelow == "minecraft:flowing_water" {
```
Sandertv, line 103: Same as above.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// ScheduledTick ...
+func (Kelp) ScheduledTick(pos world.BlockPos, w *world.World) {
+	w.BreakBlock(pos)
+}
+
+// RandomTick ...
+func (k Kelp) RandomTick(pos world.BlockPos, w *world.World, r *rand.Rand) {
+	if rand.Intn(100) < 15 && k.Age < 15 { //Every random tick, there's a 14% chance for Kelp to grow if its age is below 15.
```
Sandertv, line 115: Please use the instance of `*rand.Rand` parameter provided. The global `rand` instance has locks which slows it down considerably.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,128 @@
...
+		abovePos := pos.Add(world.BlockPos{0, 1})
+
+		liquid, liquidAboveExists := w.Liquid(abovePos)
+		block, _ := w.Block(abovePos).EncodeBlock()
+
+		//For kelp to grow, there must be only water above.
+		if liquidAboveExists && liquid.LiquidType() == "water" && (block == "minecraft:air" || block == "minecraft:water" || block == "minecraft:flowing_water") {
```
Sandertv, line 122: Same as above.

dragonfly/block/register.go
```diff
@@ -20,6 +20,7 @@ func init() {
 	world.RegisterBlock(Grass{}, Grass{Path: true})
 	world.RegisterBlock(Dirt{}, Dirt{Coarse: true})
 	world.RegisterBlock(Cobblestone{}, Cobblestone{Mossy: true})
+	world.RegisterBlock(Kelp{})
```
Sandertv, line 23: This requires you register kelp of all ages.

DenielWorld:
Kelp now functions properly. The only issue which persists is that kelp can still be placed on top of most non-solid blocks, but that can be resolved once there is some sort of a Solid interface present. I think it's ready for merge.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,144 @@
...
+	case Air, Water:
+		return false
+	}
+
+	liquid, liquidExists := w.Liquid(pos)
+	// A check for existent water at this position, kelp cannot be placed if none is found.
+	if !liquidExists || liquid.LiquidType() != "water" || liquid.LiquidDepth() < 8 {
+		// Water must be a source block to plant kelp.
+		return false
+	}
```
Sandertv, line 84: Could you change this to something like this:
```go
liquid, ok := w.Liquid(pos)
	if !ok { 
		return false
	} else if _, ok := liquid.(Water); ok || liquid.LiquidDepth() < 8 {
		return false
	}
```
Ideally we'd avoid using strings like that, LiquidType() is only really there for checking if two liquids are of the same type.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,144 @@
...
+// RandomTick ...
+func (k Kelp) RandomTick(pos world.BlockPos, w *world.World, r *rand.Rand) {
+	if r.Intn(100) < 15 && k.Age < 15 {
+		// Every random tick, there's a 14% chance for Kelp to grow if its age is below 15.
+		abovePos := pos.Add(world.BlockPos{0, 1})
+
+		liquid, liquidAboveExists := w.Liquid(abovePos)
+
+		// For kelp to grow, there must be only water above.
+		if liquidAboveExists && liquid.LiquidType() == "water" {
```
Sandertv, line 124: Same as above.

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,144 @@
...
+			switch w.Block(abovePos).(type) {
+			case Air, Water:
+				w.PlaceBlock(abovePos, Kelp{Age: k.Age + 1})
+				if liquid.LiquidDepth() < 8 {
+					// When kelp grows into a water block, the water block becomes a source block.
+					w.SetLiquid(abovePos, Water{true, 8, false})
```
Sandertv, line 130: Please use named struct fields here. (Still: true, Depth: 8, Falling: false)

dragonfly/block/kelp.go
```diff
@@ -0,0 +1,144 @@
...
+				}
+			}
+		}
+	}
+	w.ScheduleBlockUpdate(pos, time.Second/20)
+}
+
+// allKelp returns all possible states of a kelp block.
+func allKelp() (b []world.Block) {
+	for i := 15; i >= 0; i-- {
```
Sandertv, line 140: Might be a bit easier to read if you just do `for i := 0; i < 16; i++`

PR #30, opened 2020-07-21
Add functions to get a player by name

iAtomPlaza:


TwistedAsylumMC:
Personally I do not think we should follow pocketmine's api when it comes to this and we should only have builtin methods for full username & uuid searching. Sander may have a different opinion but that is mine.

iAtomPlaza:
> Personally I do not think we should follow pocketmine's api when it comes to this and we should only have builtin methods for full username & uuid searching. Sander may have a different opinion but that is mine.

sorry, i just thought it might be easier

dragonfly/server.go
```diff
@@ -187,6 +187,26 @@ func (server *Server) Player(uuid uuid.UUID) (*player.Player, bool) {
 	return nil, false
 }
 
+// GetPlayer returns a player is the name passed matches part of a name of an online player: (in-case sensitive)
+func (server *Server) GetPlayer(name string) (*player.Player, bool) {
```
T14Raptor, line 191: This is pointless in my opinion, only use for it is searching for usernames gotten from players running commands, but won't be needed since a "target" argument will most likely be added.

dragonfly/server.go
```diff
@@ -187,6 +187,26 @@ func (server *Server) Player(uuid uuid.UUID) (*player.Player, bool) {
...
+	for p := range server.players {
+		if strings.HasPrefix(strings.ToLower(p.Name()), strings.ToLower(name)) {
```
T14Raptor, line 193: Forgot to import the "strings" package.

dragonfly/server.go
```diff
@@ -187,6 +187,26 @@ func (server *Server) Player(uuid uuid.UUID) (*player.Player, bool) {
...
+			return p, true
+		}
+	}
+	return nil, false
+}
+
+// GetPlayerExact returns a player if the name passed matches an online players name: (in-case sensitive)
+func (server *Server) GetPlayerExact(name string) (*player.Player, bool) {
```
T14Raptor, line 201: I'd prefer the name "PlayerByName" instead of "GetPlayerExact"

iAtomPlaza:
I f***ed up, I thought I was doing something and everyone is against it but then I don't understand why 😭

PR #36, opened 2020-07-25
Control Server-Side XBOX Live Authentication

TwistedAsylumMC:


config.toml
```diff
@@ -14,6 +14,8 @@
   # The message shown to players when the server is shutting down. The message may be left empty to direct
   # players to the server list directly.
   ShutdownMessage = "Server closed."
+  # AuthEnabled controls whether or not players must be connected to Xbox Live in order to join the server.
+  AuthEnabled = True
```
Sandertv, line 18: Could you make this `true`?

PR #42, opened 2020-07-26
Add Beacon-Fed-Ores

DenielWorld:
Items Added:
* Emerald
* Gold Ingot
* Iron Ingot
* Netherite Ingot


dragonfly/item/netherite_ingot.go
```diff
@@ -0,0 +1,14 @@
+package item
+
+// NetheriteIngot is a rare mineral crafted with 4 pieces of netherite scrap and 4 gold ingots.
+type NetheriteIngot struct{}
+
+// EncodeItem ...
+func (NetheriteIngot) EncodeItem() (id int32, meta int16) {
+	return 742, 0
```
Sandertv, line 8: The actual ID of netherite ingots is negative.

Sandertv: Nevermind, this is only the case for blocks.

PR #46, opened 2020-07-27
Added various ores

DaPigGuy:
## Blocks
* Coal Ore
* Iron Ore
* Gold Ore
* Nether Gold Ore
* Lapis Lazuli Ore
* Diamond Ore
* Emerald Ore
* Ancient Debris

## Items
* Coal
* Gold Nuggets

### DIFF
--- a/dragonfly/block/break_info.go
+++ b/dragonfly/block/break_info.go
@@ -81,11 +81,8 @@ type BreakInfo struct {
 	XPDrops XPDropRange
 }
 
-// XPDropRange is a struct holding the min & max XP drop amounts of blocks.
-type XPDropRange struct {
-	Min,
-	Max int
-}
+// XPDropRange holds the min & max XP drop amounts of blocks.
+type XPDropRange [2]int
 
 // neverEffective is a convenience function for blocks that are mined the same by all tools.
 var neverEffective = func(t tool.Tool) bool {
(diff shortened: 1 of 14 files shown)


Sandertv:
We don't have XP yet, but it might be a good idea to introduce an `XPDropRange` with a min and max and have that as a field in `BreakInfo` so that is taken care of as soon as we actually implement xp.

dragonfly/item/coal.go
```diff
@@ -0,0 +1,9 @@
+package item
+
+// Coal is an item used as fuel & crafting torches.
+type Coal struct{}
```
Sandertv, line 4: Please add a TODO here for actually making these fuel.

dragonfly/block/break_info.go
```diff
@@ -77,6 +77,14 @@ type BreakInfo struct {
...
 	// item used to break the block is not a tool, a tool.None is passed.
 	Drops func(t tool.Tool) []item.Stack
+	// XPDrops is the range of XP a block can drop when broken.
+	XPDrops XPDropRange
+}
+
+// XPDropRange is a struct holding the min & max XP drop amounts of blocks.
+type XPDropRange struct {
+	Min,
+	Max int
```
Sandertv, line 87: I'm thinking it might be a good idea to make this a [2]int, so like
`type XPDropRange [2]int`

PR #47, opened 2020-07-27
Add all stone slabs

Aericio:


Sandertv:
I'm a bit concerned over this PR because of the `StoneSlab2`, `StoneSlab3`, `StoneSlab4` etc. I would much prefer seeing actual properly named slabs, like `SandstoneSlab`, `QuartzSlab` etc, because it's much more friendly to users of the API and doesn't expose those ugly numbers.

Aericio:
Is `StoneSlab` fine? All the rest should be split up, then?

Sandertv:
StoneSlab should probably be split up too. Sandstone and, say, stone bricks, are just completely different things.

Aericio:
Okay... I'm assuming these can stay like this right? https://github.com/df-mc/dragonfly/pull/47/commits/eeed1d42b5d7302e21fbffc97890412db5c706f3#diff-31d9d2d577baa1e58f3667fe580e8d00

Sandertv:
Might be better to just pull those out into their own block types, so that they're all unique blocks instead of just block types.

Aericio:
ah yeah. ok.

PR #50, opened 2020-07-27
Added brick blocks, brick items in progress.

ImDaRealAni:
Added brick blocks.


Sandertv:
Please feel free to add brick items in this PR too.

dragonfly/block/bricks.go
```diff
@@ -0,0 +1,39 @@
...
+type Bricks struct {
+	noNBT
+	solid
+}
+
+// BreakInfo ...
+func (b Bricks) BreakInfo() BreakInfo {
+	return BreakInfo{
+		Hardness: 2,
+		Harvestable: func(t tool.Tool) bool {
```
DaPigGuy, line 18: pickaxeHarvestable is the equivalent to this function

dragonfly/item/register.go
```diff
@@ -83,4 +83,5 @@ func init() {
 	world.RegisterItem("minecraft:clay_ball", ClayBall{})
 
 	world.RegisterItem("minecraft:stick", Stick{})
-}
+	world.RegisterItem("minecraft:bricks", Brick{})
```
Sandertv, line 86: I think this is supposed to be `minecraft:brick`

ImDaRealAni: Hm, it worked for me when I tested in the creative inventory.

Sandertv: Because this is used on disk and not over network.

dragonfly/block/bricks.go
```diff
@@ -0,0 +1,36 @@
...
+	solid
+}
+
+// BreakInfo ...
+func (b Bricks) BreakInfo() BreakInfo {
+	return BreakInfo{
+		Hardness: 2,
+		Harvestable: pickaxeHarvestable,
+		Effective: pickaxeEffective,
+		Drops: simpleDrops(item.NewStack(b, 1)),
```
Sandertv, line 19: Make sure to run go fmt over these files.

PR #55, opened 2020-07-29
Add methods to set the player mobile/immobile

TwistedAsylumMC:


dragonfly/player/player.go
```diff
@@ -767,6 +767,22 @@ func (p *Player) SetVisible() {
 	p.updateState()
 }
 
+// SetImmobile prevents the player from moving around, but still allows them to look around still.
```
Sandertv, line 770: Double `still`.

PR #58, opened 2020-07-31
retrieve player instance if you dont know their uuid

iAtomPlaza:


Sandertv:
I'm not sure if we should allow different casing in this method, maybe just checking if their exact name matches would be safer?

iAtomPlaza:
I'm not sure how that would be a problem. Perhaps you could elaborate? Cause it's something I use all the time in my commands where a parameter is a players name and I personally find it easier to not wory about case. As I mentioned, please enlighten me as to how this may be a problem?

Sandertv:
As commands will have a dedicated `Target` type that will take care of that stuff, we should probably keep this case sensitive as it is generally a good idea to be strict rather than lenient.

iAtomPlaza:
Alright, I'll make another commit

PR #62, opened 2020-08-04
Added Nether quartz, Quartz blocks, Quartz ore

mmm545:


dragonfly/block/quartz_ore.go
```diff
@@ -0,0 +1,35 @@
+package block
+
+import "github.com/df-mc/dragonfly/dragonfly/item"
+
+// Nether quartz ore is ore found in the Nether.
+type QuartzOre struct {
```
Sandertv, line 6: Might be better to call this NetherQuartzOre.

mmm545: Done

mmm545:
a

PR #64, opened 2020-08-06
Implemented potions

DaPigGuy:


dragonfly/item/item.go
```diff
@@ -79,6 +85,8 @@ type UseContext struct {
 	// NewItem is the item that is added after the item is used. If the player no longer has an item in the
 	// hand, it'll be added there.
 	NewItem Stack
+	// SurvivalDedicatedItem will add any new items only in survival mode.
+	SurvivalDedicatedItem bool
```
Sandertv, line 89: I feel like the name of this field is rather ambiguous. If I hadn't read the rest of the code, I wouldn't know what this did. Might be good to rename it to something more clear, even if the field name gets longer.

dragonfly/item/glass_bottle.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// GlassBottle is an item that can hold various liquids.
+type GlassBottle struct{}
+
+// UseOnBlock ...
+func (g GlassBottle) UseOnBlock(pos world.BlockPos, _ world.Face, _ mgl64.Vec3, w *world.World, _ User, ctx *UseContext) bool {
+	if liquid, ok := w.Liquid(pos); ok {
+		if item_internal.IsWater(liquid) && liquid.LiquidDepth() == 8 {
+			if liquid == w.Block(pos) {
```
Sandertv, line 17: Couldn't you just do `item_internal.IsWater(w.Block(pos))` and make that function accept a `world.Block`, then try to assert that to `block.Water` in that function?

dragonfly/item/glass_bottle.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/df-mc/dragonfly/dragonfly/world"
+	"github.com/go-gl/mathgl/mgl64"
+)
+
+// GlassBottle is an item that can hold various liquids.
+type GlassBottle struct{}
+
+// UseOnBlock ...
+func (g GlassBottle) UseOnBlock(pos world.BlockPos, _ world.Face, _ mgl64.Vec3, w *world.World, _ User, ctx *UseContext) bool {
+	if liquid, ok := w.Liquid(pos); ok {
```
Sandertv, line 15: You should be able to remove this check (?)

dragonfly/item/potion/potion.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// FireResistance ...
+func FireResistance() Potion {
+	return Potion{potion: 12} //TODO
+}
+
+// LongFireResistance ...
+func LongFireResistance() Potion {
+	return Potion{potion: 13} //TODO
```
Sandertv, line 81: Could you please add something like `Implement fire resistance` at these TODOs so we can easily find them once looking for fire resistance TODOs?

PR #65, opened 2020-08-07
Implemented block gravity

DaPigGuy:


dragonfly/entity/falling_block.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/df-mc/dragonfly/dragonfly/internal/item_internal"
+	"github.com/df-mc/dragonfly/dragonfly/item"
+	"github.com/df-mc/dragonfly/dragonfly/world"
+	"github.com/go-gl/mathgl/mgl64"
+	"sync/atomic"
+)
+
+// FallingBlock is the entity form of a block that appears when a gravity-affected block loses its support.
+type FallingBlock struct {
+	Block         world.Block
```
Sandertv, line 16: I wouldn't personally export this field, as the type of a FallingBlock probably shouldn't be changed once set in `NewFallingBlock`.

dragonfly/entity/falling_block.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Pitch ...
+func (f *FallingBlock) Pitch() float64 {
+	return 0
+}
+
+// State ...
+func (f *FallingBlock) State() []state.State {
+	id, _ := world.BlockRuntimeID(f.Block)
+	return []state.State{state.Varied{Variant: int(id)}}
```
Sandertv, line 87: I wouldn't put the block of the FallingBlock in the State as it doesn't change. Maybe an idea to send the right packet to set the block runtime ID for the FallingBlock right after sending the AddActor packet?

dragonfly/block/block.go
```diff
@@ -141,3 +143,30 @@ type transparent struct{}
 func (transparent) LightDiffusionLevel() uint8 {
 	return 0
 }
+
+// GravityAffected represents blocks affected by gravity.
+type GravityAffected interface {
+	// CanSolidify returns whether the falling block can return back to a normal block without being on the ground.
+	CanSolidify(pos world.BlockPos, w *world.World) bool
```
Sandertv, line 150: Is there any circumstance under which someone might want this to return true? I think it might be more useful to simply have a `GravityAffected` method so that different variants of a block could return different values.

dragonfly/block/block.go
```diff
@@ -141,3 +143,30 @@ type transparent struct{}
...
+
+// fall spawns a falling block entity at the given position.
+func (g gravityAffected) fall(b world.Block, pos world.BlockPos, w *world.World) {
+	_, air := w.Block(pos.Side(world.FaceDown)).(Air)
+	_, liquid := w.Liquid(pos.Side(world.FaceDown))
+	if air || liquid {
+		w.BreakBlock(pos)
+
+		e := entity.NewFallingBlock(b, pos.Vec3())
+		e.SetVelocity(mgl64.Vec3{})
```
Sandertv, line 169: This is already the velocity after calling `NewFallingBlock`.

PR #67, opened 2020-08-08
Added brick block and item

ImDaRealAni:


dragonfly/item/register.go
```diff
@@ -101,6 +101,8 @@ func init() {
 
 	world.RegisterItem("minecraft:apple", Apple{})
 
+	world.RegisterItem("minecraft:bricks", Brick{})
```
Sandertv, line 104: This is still the wrong string for a brick. (Should be `minecraft:brick`)

ImDaRealAni:
I fixed the namespace ID.

PR #68, opened 2020-08-10
Implemented fire spreading

DaPigGuy:
This PR includes the following changes:
* Added entity -> block collisions
* Added method world.BreakBlockWithoutParticles()
  * Usages of world.BreakBlock() in carpet, cocoa, crops, kelp, lanterns, & falling blocks have been migrated
* Implemented fire spreading
  * Added interface Flammable
    * Existing flammable blocks now implement this interface
  * Added struct FlammabilityInfo
  * Added fire & lava damage
     * Implemented fire resistance effect
  * Added fire & soul fire blocks
  * Added flint & steel
  * Lava now spreads fire to nearby flammable blocks
* Added Crimson & Warped variants of wooden blocks
* Added soul sand & soul soil
  * Bubble Columns are currently not implemented
* Miscellaneous additions
  * Added barrier blocks
  * Added blue ice
  * Added basalt & polished basalt
    * Lava flowing over soul soil while neighbouring blue ice will generate basalt
  * Added smooth stone
  * Added crying obsidian
  * Added gilded blackstone
  * Added sea lanterns
  * Added prismarine crystals
  * Added shroomlight
* Miscellaneous changes
  * Updated leave drops
* Miscellaneous bug fixes
  * Fixed buckets being added to inventory upon emptying while in creative mode
  * Fixed fence gates being unbreakable in survival mode
  * Fixed falling block entity AABB
  * Fixed cocoa block AABB for growth stages 1 & 2


dragonfly/block/wood/wood.go
```diff
@@ -6,41 +6,52 @@ import "fmt"
 // others carry one of these types.
 type Wood struct {
 	wood
+	Flammable bool
```
Sandertv, line 9: My only concern with bools like this is that is that the default `Wood` will not be flammable. This could cause some nasty bugs. Might be better to just have a method that returns whether it's flammable based on `wood`.

Sandertv: The same counts for the other types where it's like this, but that's for later.

dragonfly/entity/flammable.go
```diff
@@ -0,0 +1,15 @@
...
+	// FireProof is whether the entity is currently fire proof.
+	FireProof() bool
+	// FireDamage deals fire damage to the entity.
+	FireDamage(amount float64)
+	// LavaDamage deals lava damage to the entity.
+	LavaDamage(amount float64)
+	// FireTicks returns duration of fire in ticks.
+	FireTicks() int
+	// SetFireTicks sets the player on fire for the specified duration in ticks.
+	SetFireTicks(ticks int)
```
Sandertv, line 14: It's not very nice if we have to expose all these. A `SetOnFire(time.Duration)` method might be nicer, along with an `Extinguish()` method (and change `FireTicks()` to `OnFireDuration()`). As for the other methods, I'm guessing they're needed to get around circular import issues. Is there any way we can use an internal package to get around this?

dragonfly/player/player.go
```diff
@@ -1503,6 +1546,28 @@ func (p *Player) Tick(current int64) {
...
+	aabb := p.AABB().Translate(p.Position())
+	for x := int(aabb.Min().X()); x <= int(aabb.Max().X()); x++ {
+		for y := int(aabb.Min().Y()); y <= int(aabb.Max().Y()); y++ {
+			for z := int(aabb.Min().Z()); z <= int(aabb.Max().Z()); z++ {
+				if collide, ok := p.World().Block(world.BlockPos{x, y, z}).(block.EntityColliding); ok {
+					collide.EntityCollide(p)
+				}
+			}
+		}
+	}
```
Sandertv, line 1569: Could you add a TODO to investigate moving this to `Move()` once we get proper movement interpolation?

dragonfly/world/world.go
```diff
@@ -291,6 +291,18 @@ func (w *World) BreakBlock(pos BlockPos) {
 	}
 }
 
+// BreakBlockWithoutParticles breaks a block at the position passed. Unlike when setting the block at that position to air,
+// BreakBlockWithoutParticles will also update blocks around the position.
+func (w *World) BreakBlockWithoutParticles(pos BlockPos) {
```
Sandertv, line 296: The name of this method is a bit meh, but I can't think of anything better right off the bat.

Sandertv (changes requested):

dragonfly/block/fire.go
```diff
@@ -0,0 +1,257 @@
...
+// FlammableBlock returns true if a block is flammable.
+func FlammableBlock(block world.Block) bool {
+	if flammable, ok := block.(Flammable); ok && flammable.FlammabilityInfo().Encouragement > 0 {
+		return true
+	}
+	return false
+}
+
+// woodTypeFlammable returns whether a type of wood is flammable.
+func woodTypeFlammable(w wood.Wood) bool {
```
Sandertv, line 36: Might be better to move this into the wood.Wood an have it return the flammability info itself.

dragonfly/block/fire.go
```diff
@@ -0,0 +1,257 @@
...
+
+		for y := -1; y <= 4; y++ {
+			randomBound := 100
+			if y > 1 {
+				randomBound += (y - 1) * 100
+			}
+
+			for x := -1; x <= 1; x++ {
+				for z := -1; z <= 1; z++ {
+					if x != 0 || y != 0 || z != 0 {
```
Sandertv, line 151: Might be best to not nest this either, but continue immediately.

dragonfly/block/fire.go
```diff
@@ -0,0 +1,257 @@
...
+func (f Fire) EntityCollide(e world.Entity) {
+	if flammable, ok := e.(entity.Flammable); ok {
+		flammable.FireDamage(1)
+		flammable.SetOnFire(160)
+	}
+}
+
+// ScheduledTick ...
+func (f Fire) ScheduledTick(pos world.BlockPos, w *world.World) {
+	f.tick(pos, w)
```
Sandertv, line 196: This will make fire spread very quickly if you spam place blocks next to it. Is that intended?

dragonfly/block/leaves.go
```diff
@@ -29,8 +40,17 @@ func (l Leaves) BreakInfo() BreakInfo {
 		Effective: func(t tool.Tool) bool {
 			return t.ToolType() == tool.TypeShears || t.ToolType() == tool.TypeHoe
 		},
-		// TODO: Add saplings and apples and drop them here.
-		Drops: simpleDrops(),
+		Drops: func(t tool.Tool) (drops []item.Stack) {
+			if t.ToolType() == tool.TypeShears { // TODO: Silk Touch
+				drops = append(drops, item.NewStack(l, 1))
```
Sandertv, line 45: Could simply `return` here, then remove the `else`.

dragonfly/entity/flammable.go
```diff
@@ -0,0 +1,17 @@
+package entity
+
+// Flammable is an interface for entities that can be set on fire.
+type Flammable interface {
+	// FireProof is whether the entity is currently fire proof.
+	FireProof() bool
+	// FireDamage deals fire damage to the entity.
+	FireDamage(amount float64)
+	// LavaDamage deals lava damage to the entity.
+	LavaDamage(amount float64)
```
Sandertv, line 10: Can we somehow get rid of this method and the method above? If we can move this to an internal package to get around import cycles that would be great. The methods by themselves are pretty useless.

dragonfly/entity/flammable.go
```diff
@@ -0,0 +1,17 @@
...
+	// OnFireDuration returns duration of fire in ticks.
+	OnFireDuration() int
+	// SetOnFire sets the entity on fire for the specified duration.
+	SetOnFire(ticks int)
```
Sandertv, line 14: Ideally these two methods would operate on a `time.Duration`. Ticks are pretty awkward to work with in general for users. You can translate them to ticks in these methods.

Sandertv (changes requested):

dragonfly/player/player.go
```diff
@@ -666,6 +673,7 @@ func (p *Player) Respawn() {
 	p.addHealth(p.MaxHealth())
 	p.hunger.Reset()
 	p.sendFood()
+	p.SetOnFire(0)
```
Sandertv, line 676: Can use p.Extinguish() here.

dragonfly/item/flint_and_steel.go
```diff
@@ -0,0 +1,43 @@
...
+	return DurabilityInfo{
+		MaxDurability: 65,
+		BrokenItem:    simpleItem(Stack{}),
+	}
+}
+
+// UseOnBlock ...
+func (f FlintAndSteel) UseOnBlock(pos world.BlockPos, face world.Face, _ mgl64.Vec3, w *world.World, _ User, ctx *UseContext) bool {
+	ctx.DamageItem(1)
+	w.PlaySound(pos.Vec3(), sound.Ignite{})
```
Sandertv, line 31: I don't think a flint and steel is supposed to lose durability and play a sound when it doesn't actually do anything.

dragonfly/player/player.go
```diff
@@ -789,6 +797,29 @@ func (p *Player) SetMobile() {
...
+func (p *Player) SetOnFire(duration time.Duration) {
+	p.fireTicks.Store(int64(duration.Seconds() * 20))
+	p.updateState()
+}
+
+// Extinguish ...
+func (p *Player) Extinguish() {
+	if p.OnFireDuration() > 0 {
+		p.SetOnFire(0)
+	}
```
Sandertv, line 820: Might want to change this to just `p.SetOnFire(0).`

dragonfly/player/player.go
```diff
@@ -1503,6 +1539,29 @@ func (p *Player) Tick(current int64) {
 		p.Hurt(4, damage.SourceVoid{})
 	}
 
+	if p.OnFireDuration() > 0 {
+		if p.FireProof() {
+			p.SetOnFire(0)
```
Sandertv, line 1544: Can use p.Extinguish() here.

dragonfly/player/player.go
```diff
@@ -1503,6 +1539,29 @@ func (p *Player) Tick(current int64) {
...
+		} else {
+			p.SetOnFire(p.OnFireDuration() - time.Second/20)
```
Sandertv, line 1546: Could only subtract one from the player's current fire ticks and only call Extinguish() if they drop below 0 so you don't send packets everytime.

Sandertv (changes requested):

PR #73, opened 2020-09-29
basic compass

xerenahmed:


Sandertv:
Thanks for your PR!

Please make sure to resolve the build issues so I can review this.

dragonfly/item/compass.go
```diff
@@ -0,0 +1,9 @@
+package item
+
+// Compass is helps to find the spawn place.
```
Sandertv, line 3: Grammar isn't perfect here, something like "Compass is an item used to find the spawn position." would be better.

Thunder33345:
should there be functionalities of like changing where it points without setting the world spawn?
since in df there's not going to be underlying packet access, i feel like it's necessary to at least brought it up

Sandertv:
Changing the world spawn should result in a packet being sent to all viewers of the world. I'm not sure that's in the scope of this PR though.

PR #74, opened 2020-09-30
Implemented torches

DaPigGuy:


dragonfly/block/torch.go
```diff
@@ -0,0 +1,122 @@
...
+// Torch are non-solid blocks that emit light.
+type Torch struct {
+	noNBT
+	empty
+
+	// Facing is the direction from the torch to the block.
+	Facing world.Face
+	// Type is the type of fire lighting the torch.
+	Type fire.Fire
+}
```
Sandertv, line 19: Don't forget to embed the transparent struct so skylight doesn't get blocked.

Sandertv (changes requested):
One small nitpick again.

Sandertv (approved):
All good, thanks for the PR!

PR #77, opened 2020-09-30
Implemented dragon egg

DaPigGuy:


dragonfly/session/world.go
```diff
@@ -741,3 +759,8 @@ func vec32To64(vec3 mgl32.Vec3) mgl64.Vec3 {
 func vec64To32(vec3 mgl64.Vec3) mgl32.Vec3 {
 	return mgl32.Vec3{float32(vec3[0]), float32(vec3[1]), float32(vec3[2])}
 }
+
+// abs ...
+func abs(a int) int {
+	return int(math.Abs(float64(a)))
```
Sandertv, line 765: Might be better not to use floats here. A simple `if a > 0 { return a } return -a` might be better.

dragonfly/world/particle/block.go
```diff
@@ -31,6 +31,14 @@ type BlockForceField struct{ particle }
 // Bonemeal is a particle that shows up on bonemeal usage.
 type Bonemeal struct{ particle }
 
+// DragonEggTeleport is a particle that shows up when a dragon egg teleports.
+type DragonEggTeleport struct {
+	particle
+
+	// Diff is the distance teleported by the dragon egg.
```
Sandertv, line 38: Might be good to add a code sample here on what you would generally set to this field.

Sandertv (changes requested):
Thanks for the pull request! All looks good except for some very small nitpicks. I've added some comments to the diff.

Sandertv (approved):
Looks good! Thank you for the PR!

PR #79, opened 2020-09-30
Implemented note blocks

DaPigGuy:


dragonfly/block/instrument/instrument.go
```diff
@@ -0,0 +1,86 @@
+package instrument
+
+// Instrument represents a note block instrument.
+type Instrument struct {
+	MagicNumber int
```
Sandertv, line 5: Would be better to wrap this in an unexported `instrument` struct so new Instruments (with invalid values) cannot be constructed.

dragonfly/block/note_block.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// NoteBlock is a musical block that emits sounds when powered with redstone.
+type NoteBlock struct {
+	nbt
+	solid
+	bass
+
+	// Pitch is the current pitch the note block is set to. Value ranges from 0-24.
+	Pitch int32
```
Sandertv, line 18: Would be best to keep this an `int` for ease of use. Make sure to cast it to an int32 in the NBT when you change it.

dragonfly/player/player.go
```diff
@@ -1100,6 +1100,12 @@ func (p *Player) StartBreaking(pos world.BlockPos, face world.Face) {
 	ctx := event.C()
 	p.handler().HandleStartBreak(ctx, pos)
 	ctx.Continue(func() {
+		if punchable, ok := p.World().Block(pos).(block.Punchable); ok {
+			p.swingArm()
+			punchable.Punch(pos, face, p.World(), p)
+			return
+		}
+
```
Sandertv, line 1108: This might prevent breaking punchable blocks. I don't think you'd want to return here.

dragonfly/block/block.go
```diff
@@ -196,3 +205,41 @@ type EntityCollider interface {
...
+// snare is a struct that may be embedded for blocks that create a snare drum sound.
+type snare struct{}
+
+// Instrument ...
+func (snare) Instrument() instrument.Instrument {
+	return instrument.Snare()
+}
+
+// clicksandsticks is a struct that may be embedded for blocks that create a clicks and sticks sound.
+type clicksandsticks struct{}
```
Sandertv, line 232: Might be better named `clicksAndSticks`

dragonfly/block/block.go
```diff
@@ -196,3 +205,41 @@ type EntityCollider interface {
...
+
+// Instrument ...
+func (clicksandsticks) Instrument() instrument.Instrument {
+	return instrument.ClicksAndSticks()
+}
+
+// bassdrum is a struct that may be embedded for blocks that create a bass drum sound.
+type bassdrum struct{}
```
Sandertv, line 240: Might be better named `bassDrum`.

Sandertv (changes requested):
Thanks for the PR! I left some comments. Please resolve them when you get the time.

dragonfly/block/instrument/instrument.go
```diff
@@ -0,0 +1,88 @@
+package instrument
+
+// Instrument represents a note block instrument.
+type Instrument struct {
+	MagicNumber instrument
```
Sandertv, line 5: Might want to unexport this field and expose a Uint8() method. People can still assign integers to this.

Sandertv (changes requested):
One last small comment. You might want to make the Instrument type similar to something like, say, the colour.Colour struct.

dragonfly/block/instrument/instrument.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Instrument represents a note block instrument.
+type Instrument struct {
+	instrument
+}
+
+type instrument int32
+
+// MagicNumber ...
+func (i instrument) MagicNumber() int32 {
```
Sandertv, line 11: I'm not so much a fan of `MagicNumber`. Might be better to just call it `Int32()`.

Sandertv (approved):
Looks good! Thanks for the PR!

PR #82, opened 2020-10-03
Implemented ice and packed ice

provsalt:
Implemented ice.go and packed_ice.go

### DIFF
--- a/dragonfly/block/block.go
+++ b/dragonfly/block/block.go
@@ -124,7 +124,7 @@ func (noNBT) HasNBT() bool {
 	return false
 }
 
-// nbt may be embedded by blocks that do have NBT.
+// nbt may be embenbdded by blocks that do have NBT.
 type nbt struct{}
 
 // HasNBT ...
--- a/dragonfly/block/ice.go
+++ b/dragonfly/block/ice.go
@@ -1,0 +1,38 @@
+package block
+
+import "github.com/df-mc/dragonfly/dragonfly/item/tool"
+
+// TODO: Slipperiness and SilkTouch
+
+// Ice is a solid block similar to packed ice.
+type Ice struct {
+	noNBT
+	solid
+}
+
+// BreakInfo ...
+func (i Ice) BreakInfo() BreakInfo {
+	return BreakInfo{
+		Hardness: 0.5,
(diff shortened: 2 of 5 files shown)


dragonfly/block/block.go
```diff
@@ -124,7 +124,7 @@ func (noNBT) HasNBT() bool {
 	return false
 }
 
-// nbt may be embedded by blocks that do have NBT.
+// nbt may be embenbdded by blocks that do have NBT.
```
Sandertv, line 127: Typo.

dragonfly/block/ice.go
```diff
@@ -0,0 +1,44 @@
+package block
+
+import "github.com/df-mc/dragonfly/dragonfly/item/tool"
+
+// TODO: Slipperiness, melting and SilkTouch
```
Sandertv, line 5: Melting should be implementable using the current system. From the wiki:
"Ice can be used to create water either by its melting or being broken. If there is another block directly underneath the ice block, the ice reverts to water when broken. Ice also melts into water if the light level immediately next to it on any side is higher than 11, from light sources other than sunlight (and regardless of whether there's a block below)."

dragonfly/block/ice.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Ice is a solid block similar to packed ice.
+type Ice struct {
+	noNBT
+	transparent
```
Sandertv, line 10: `transparent` returns a LightDiffusionLevel of 0, but the method below returns 2 (and correctly so). I don't think you'd want to embed transparent.

dragonfly/block/packed_ice.go
```diff
@@ -0,0 +1,38 @@
+package block
+
+import "github.com/df-mc/dragonfly/dragonfly/item/tool"
+
+// TODO: Slipperiness, slowness, noteblock sound and SilkTouch
```
Sandertv, line 5: Can remove SilkTouch here, there's a TODO you added below.

Sandertv (changes requested):
Thanks for the PR! I left some comments below.

Sandertv:
Noteblocks were merged, so please make sure to merge those into your branch and fix conflicts so you can add instruments to the ice.

PR #85, opened 2020-10-16
Add method for entity types

T14Raptor:


dragonfly/session/world.go
```diff
@@ -189,15 +189,15 @@ func (s *Session) ViewEntity(e world.Entity) {
 			HeadYaw:         float32(e.Yaw()),
 		})
 	default:
+		entityType := e.EncodeEntity()
```
Sandertv, line 192: Feel free to just move this into the packet directly without extra assignment.

PR #92, opened 2020-11-03
Implement custom join/leave messages

AndreasHGK:
This PR adds a basic implementation of customizable join and leave messages.

You can set the messages both in the `config.toml` file or directly with code for each player.


dragonfly/session/text.go
```diff
@@ -134,3 +134,13 @@ func (s *Session) SendSubtitle(text string) {
 func (s *Session) SendActionBarMessage(text string) {
 	s.writePacket(&packet.SetTitle{ActionType: packet.TitleActionSetActionBar, Text: text})
 }
+
+// SetJoinMessage changes the message that appears when the player joins the server
+func (s *Session) SetJoinMessage(message string) {
+	s.joinMessage = message
```
Sandertv, line 140: I think these properties are vulnerable to race conditions. Consider changing them to an `atomic.String`

dragonfly/player/player.go
```diff
@@ -256,6 +256,16 @@ func (p *Player) RemoveBossBar() {
...
 
+// SetJoinMessage changes the message that appears when the player joins the server
+func (p *Player) SetJoinMessage(message string) {
+	p.s.SetJoinMessage(message)
+}
+
+// SetLeaveMessage changes the message that appears when the player leaves the server
+func (p *Player) SetLeaveMessage(message string) {
+	p.s.SetLeaveMessage(message)
+}
```
Sandertv, line 267: I'm not too fond of these messages and the one in Session. Maybe just SetJoinMessage and SetLeaveMessage in server.go would be better?

dragonfly/config.go
```diff
@@ -20,6 +20,12 @@ type Config struct {
 		ShutdownMessage string
 		// AuthEnabled controls whether or not players must be connected to Xbox Live in order to join the server.
 		AuthEnabled bool
+		// JoinMessage is the message that appears when a player joins the server. Leave this empty to disable it.
+		// %v is the placeholder for the username of the player
```
Sandertv, line 24: Make sure to add this documentation in the respective methods too.

Sandertv (changes requested):
Thanks for you PR! I've left some comments on the diff.

dragonfly/server.go
```diff
@@ -217,6 +221,30 @@ func (server *Server) SetName(a ...interface{}) {
 	server.name.Store(fmt.Sprint(a...))
 }
 
+// JoinMessage is the message that appears when a player joins the server. Leave this empty to disable it.
+// %v is the placeholder for the username of the player
+func (server *Server) JoinMessage() string {
```
Sandertv, line 226: Is there actually any use for this?

dragonfly/server.go
```diff
@@ -217,6 +221,30 @@ func (server *Server) SetName(a ...interface{}) {
...
+
+// SetJoinMessage changes the join message for all players on the server. Leave this empty to disable it.
+//// %v is the placeholder for the username of the player
+func (server *Server) SetJoinMessage(message string) {
+	server.joinMessage.Store(message)
+}
+
+// LeaveMessage is the message that appears when a player leaves the server. Leave this empty to disable it.
+// %v is the placeholder for the username of the player
+func (server *Server) LeaveMessage() string {
```
Sandertv, line 238: Same as above.

Sandertv (commented):
Two comments, looks pretty good now!

dragonfly/server.go
AndreasHGK: The use is just to return what the current join/leave message is

Sandertv: Yeah, but is there any reason anyone would like... Have any use for that?

AndreasHGK: I don't know lol, I just added it in case someone somewhere would need it.

Sandertv (approved):

PR #93, opened 2020-11-23
Add plugin system

py354:
Now you can add your server plugins. For instance coloring messages red:

**red_text_plugin.go**
`type RedText struct {}`
`func (r *RedText) HandleChat(p *player.Player, ctx *event.Context, message *string) {`
`	*message = text.Colourf("<red>%s</red>", *message)`
`}`

**main.go**
	`server := dragonfly.New(&config, log)`
	`server.AddPlugin(&RedText{})`

### DIFF
--- a/dragonfly/player/handler.go
+++ b/dragonfly/player/handler.go
@@ -15,81 +15,81 @@ import (
 type MoveHandler interface {
 	// HandleMove handles the movement of a player. ctx.Cancel() may be called to cancel the movement event.
 	// The new position, yaw and pitch are passed.
-	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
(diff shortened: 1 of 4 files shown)


Saiini:
DF doesn't support plugins because the user will need Go installed, then recompile the entire binary to use that very plugin. Not only does this put weight on an inexperienced user but this will cause an issue with contributors because they need to re-export everything they just did just to support the plugin API. **You need to use DF as a library.**

Sandertv:
Hi py543, thanks for your PR!
Unfortunately, right now, we're not really looking into adding this multihandler concept builtin in Dragonfly. This is a feature that can be added easily as an external library and we would prefer people to only use it when applicable. There are also some other points with this PR which don't exactly work the way they should, like this feature only working with player handlers and a big maintenance burden that comes with this PR.
I will close this PR for those reasons, but please do remember that you can always implement this as a separate go library that people can use.

PR #103, opened 2021-01-31
Implement grasses (grass, ferns, tall grass, large ferns, and nether sprouts)

JustTalDevelops:


dragonfly/block/grass_plant.go
```diff
@@ -0,0 +1,157 @@
...
+	"math/rand"
+)
+
+// GrassPlant is a transparent plant block which can be used to obtain seeds and as decoration.
+type GrassPlant struct {
+	noNBT
+	transparent
+	empty
+
+	UpperBit bool
```
Sandertv, line 18: This isn't a bit so this should probably be named something else. Maybe something like `UpperPart`.

dragonfly/block/grass_plant.go
```diff
@@ -0,0 +1,157 @@
...
+
+	grass.Grass
```
Sandertv, line 20: Not sure if this should be embedded. Might be better to just make it a normal field (`Grass` or `GrassType`).

dragonfly/block/grass_plant.go
```diff
@@ -0,0 +1,157 @@
...
+// BoneMeal attempts to affect the block using a bone meal item.
+func (g GrassPlant) BoneMeal(pos world.BlockPos, w *world.World) bool {
+	switch g.Grass {
+	case grass.SmallGrass():
+		w.SetBlock(pos, GrassPlant{Grass: grass.TallGrass()})
+		w.SetBlock(pos.Side(world.FaceUp), GrassPlant{Grass: grass.TallGrass(), UpperBit: true})
+		return true
+	case grass.Fern():
+		w.SetBlock(pos, GrassPlant{Grass: grass.LargeFern()})
+		w.SetBlock(pos.Side(world.FaceUp), GrassPlant{Grass: grass.LargeFern(), UpperBit: true})
```
Sandertv, line 66: Are these supposed to trigger block updates around it? If so this should use PlaceBlock.

Sandertv (changes requested):
Got a couple of comments. Looks good for the most part.

dragonfly/block/grass_plant.go
JustTalDevelops: It isn't, which is why I use SetBlock.

dragonfly/block/grass_plant.go
JustTalDevelops: Renamed to Type.

dragonfly/block/grass_plant.go
JustTalDevelops: Renamed to UpperPart.

Sandertv (approved):

PR #107, opened 2021-02-28
Set default spawn to the highest block at 0, 0 

JustTalDevelops:


dragonfly/server.go
```diff
@@ -381,6 +379,13 @@ func (server *Server) loadWorld() {
 		server.log.Fatalf("error loading world: %v", err)
 	}
 	server.world.Provider(p)
+	if _, ok := server.world.Generator().(world.NopGenerator); ok {
+		server.world.SetGenerator(generator.Flat{})
+	}
+	if p.FirstLoad() {
+		server.world.SetSpawn(world.BlockPos{0, server.world.HighestBlock(0, 0), 0})
+	}
```
TwistedAsylumMC, line 387: Are you forgetting to set FirstLoad to false?

dragonfly/world/mcdb/provider.go
```diff
@@ -85,6 +86,11 @@ func (p *Provider) initDefaultLevelDat() {
 	p.d.MultiPlayerGame = true
 }
 
+// FirstLoad returns true if this is the first time the world was loaded.
+func (p *Provider) FirstLoad() bool {
```
TwistedAsylumMC, line 90: The naming of this function breaks the pattern used in the rest of the file. It might be best to come up with a new name which matches the pattern to be consistent.

dragonfly/server.go
JustTalDevelops: No, FirstLoad is meant to be a way of checking if the level data existed before the the world was loaded, so as a way of checking if this is the first time the server is generating level data for the world.

dragonfly/world/mcdb/provider.go
JustTalDevelops: Would "LoadFirstLoad" be a better name?

dragonfly/server.go
JustTalDevelops: If the level.dat doesn't exist, default data is used and in it we set FirstLoad to true because we can assume that this is the first time the world data is being created. I could probably rename this to something less confusing.

TwistedAsylumMC: Okay that makes sense

PR #108, opened 2021-02-28
Set default spawn to the highest block at 0, 0

JustTalDevelops:


JustTalDevelops:
Updated to return the FirstLoad value in the return params for mcdb.New

dragonfly/world/mcdb/provider.go
```diff
@@ -31,25 +31,27 @@ const chunkVersion = 19
 // New creates a new provider reading and writing files to files under the path passed. If a world is present
 // at the path, New will parse its data and initialise the world with it. If the data cannot be parsed, an
 // error is returned.
-func New(dir string) (*Provider, error) {
+func New(dir string) (p *Provider, previouslyExisted bool, err error) {
```
Sandertv, line 34: I don't think we should add this here if it's not needed in the function. The caller can just do a os.Stat() beforehand to see if a level.dat exists.

dragonfly/world/mcdb/provider.go
```diff
@@ -117,12 +118,7 @@ func (p *Provider) SetWorldName(name string) {
 
 // WorldSpawn returns the spawn of the world as present in the level.dat.
 func (p *Provider) WorldSpawn() world.BlockPos {
-	y := p.d.SpawnY
-	if p.d.SpawnY > 256 {
```
Sandertv, line 121: Vanilla will have the SpawnY set to the maximum value by default and finds a safe spawn at the highest block. Removing this will spawn the player miles in the sky.

Sandertv (changes requested):

dragonfly/world/mcdb/provider.go
JustTalDevelops: Ah I see, that makes a lot more sense now. Would keeping the world spawn function how it is but adding a check to the initial load world sequence to see if the spawn Y is above 256, and if so, setting the spawn to the highest block be a good idea?

dragonfly/world/mcdb/provider.go
JustTalDevelops: Removed and ran an OS stat beforehand instead

dragonfly/world/mcdb/provider.go
Sandertv: Yeah that sounds fair enough.

JustTalDevelops: Alright, should be done then

JustTalDevelops:
Anything else that needs to be changed?

Sandertv:
Looks good now, thank you!

Sandertv (approved):

PR #111, opened 2021-03-16
Prefer normal namespace with meta over unique namespace

JustTalDevelops:
This should fix all item entries issues (when adding new items to Dragonfly)


Sandertv:
Not sure what this changes as it's such a massive file, but I'll merge it, assuming it's tested.

PR #115, opened 2021-03-28
Implement fall damage for players

JustTalDevelops:


Sandertv:
Thanks for the PR! This currently does not take into account liquids, where I think it should reset the fall distance. I'm just entirely not sure if simply checking if the player is inside of a liquid would be sufficient, because if the player was falling really quickly it might not have a movement where the position is inside of the liquid(?)

dragonfly/entity/damage/source.go
```diff
@@ -45,10 +45,18 @@ type SourceFireTick struct{}
...
+// SourceFall is a source that is used if the player fell.
+type SourceFall struct{}
+
 // SourceCustom is a cause used for dealing any kind of custom damage. Armour reduces damage of this source,
 // but otherwise no enchantments have an additional effect.
 type SourceCustom struct{}
 
+// ReducedByArmour ...
+func (SourceFall) ReducedByArmour() bool {
+	return true
```
Sandertv, line 57: Does all armour reduce damage from fall damage, or just boots with feather falling?

dragonfly/player/player.go
```diff
@@ -404,6 +406,37 @@ func (p *Player) Heal(health float64, source healing.Source) {
...
+func (p *Player) fall(fallDistance float64) {
+	fallDamage := fallDistance - 3
+	for _, e := range p.Effects() {
+		if _, ok := e.(effect.JumpBoost); ok {
+			fallDamage -= float64(e.Level())
+		}
+	}
+	fallDamage = math.Ceil(fallDamage)
+
+	if fallDamage < 0.5 {
```
Sandertv, line 433: It ceils first, then checks if it's smaller than 0.5, but that will never be the case unless it's 0 exactly. You'd probably want to move the ceil below this block.

dktapps:
> if the player was falling really quickly it might not have a movement where the position is inside of the liquid(?)

the 99% case would be landing in water on the ground, so the collision with the solid block would ensure it, no?

dragonfly/entity/damage/source.go
dktapps: Armour doesn't reduce fall damage by itself.

Sandertv:
In a lot of cases, yes, but there are cases where people have water with signs under it to drop through. I don't know to what extent it would happen even with that though.

dragonfly/entity/damage/source.go
JustTalDevelops: Would implementing reduced fall damage for feather falling be in the scope of this PR? Just wondering since things like PvP aren't affected by enchantments yet

dragonfly/player/player.go
JustTalDevelops: Moved.

dragonfly/entity/damage/source.go
Sandertv: Nah not, I'd put that in a PR of a feather falling enchantment.

JustTalDevelops: Ah okay, this is resolved then.

JustTalDevelops:
For now I'm just checking if the block at the player's position is equal to a liquid before attempting to apply any fall damage, and it seems to work. I don't think this should cause any issues?

Sandertv:
That would break if the player first dropped through water, then hit the ground later.

JustTalDevelops:
Implemented the method you suggested, let me know if I'm missing anything

Sandertv:
Looks good, thank you!

Sandertv (approved):

PR #119, opened 2021-04-13
(WIP) 1.16.220 support

JustTalDevelops:
Currently I'm really skeptical on the approach I took to block breaking, but everything else seems fine.
Feedback and testers would be appreciated.

### DIFF
--- a/dragonfly/player/player.go
+++ b/dragonfly/player/player.go
@@ -1212,6 +1212,11 @@ func (p *Player) AbortBreaking() {
 	}
 }
 
+// IsBreaking returns true if the player is currently breaking a block.
+func (p *Player) IsBreaking() bool {
+	return p.breaking.Load()
+}
+
 // ContinueBreaking makes the player continue breaking the block it started breaking after a call to
 // Player.StartBreaking().
 // The face passed is used to display particles on the side of the block broken.
--- a/dragonfly/player/player.go
+++ b/dragonfly/player/player.go
@@ -1212,8 +1212,8 @@ func (p *Player) AbortBreaking() {
 	}
 }
 
-// IsBreaking returns true if the player is currently breaking a block.
-func (p *Player) IsBreaking() bool {
+// Breaking returns true if the player is currently breaking a block.
+func (p *Player) Breaking() bool {
 	return p.breaking.Load()
 }
 
--- a/dragonfly/player/player.go
+++ b/dragonfly/player/player.go
@@ -1212,11 +1212,6 @@ func (p *Player) AbortBreaking() {
 	}
 }
 
-// Breaking returns true if the player is currently breaking a block.
-func (p *Player) Breaking() bool {
-	return p.breaking.Load()
-}
-
 // ContinueBreaking makes the player continue breaking the block it started breaking after a call to
 // Player.StartBreaking().
 // The face passed is used to display particles on the side of the block broken.
--- a/dragonfly/server.go
+++ b/dragonfly/server.go
@@ -335,7 +335,7 @@ func (server *Server) handleConn(conn *minecraft.Conn) {
 		GameRules:                    map[string]interface{}{"naturalregeneration": false},
 		Difficulty:                   2,
 		Items:                        server.itemEntries(),
-		PlayerMovementSettings:       protocol.PlayerMovementSettings{MovementType: protocol.PlayerMovementModeServer},
+		PlayerMovementSettings:       protocol.PlayerMovementSettings{MovementType: protocol.PlayerMovementModeServer, ServerAuthoritativeBlockBreaking: true},
 		ServerAuthoritativeInventory: true,
 	}
 	if err := conn.StartGame(data); err != nil {
--- a/dragonfly/session/handler_player_action.go
+++ b/dragonfly/session/handler_player_action.go
@@ -3,7 +3,6 @@ package session
 import (
 	"fmt"
 	"github.com/df-mc/dragonfly/dragonfly/world"
-	"github.com/df-mc/dragonfly/dragonfly/world/gamemode"
 	"github.com/go-gl/mathgl/mgl64"
 	"github.com/sandertv/gophertunnel/minecraft/protocol"
 	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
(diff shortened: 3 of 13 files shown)


dragonfly/player/player.go
```diff
@@ -1212,6 +1212,11 @@ func (p *Player) AbortBreaking() {
 	}
 }
 
+// IsBreaking returns true if the player is currently breaking a block.
+func (p *Player) IsBreaking() bool {
```
Sandertv, line 1216: Would call this `Breaking`

dragonfly/server.go
```diff
@@ -335,7 +335,7 @@ func (server *Server) handleConn(conn *minecraft.Conn) {
 		GameRules:                    map[string]interface{}{"naturalregeneration": false},
 		Difficulty:                   2,
 		Items:                        server.itemEntries(),
-		PlayerMovementSettings:       protocol.PlayerMovementSettings{MovementType: protocol.PlayerMovementModeServer},
+		PlayerMovementSettings:       protocol.PlayerMovementSettings{MovementType: protocol.PlayerMovementModeServer, ServerAuthoritativeBlockBreaking: true},
```
Sandertv, line 338: Do we actually need this? If we don't use this we don't need any changing of the block breaking.

dragonfly/world/world.go
```diff
@@ -82,7 +82,7 @@ type World struct {
 	positionCache       []ChunkPos
```
Sandertv, line 82: I don't think these changes are supposed to be here. Did you revert a commit on accident?

Sandertv (changes requested):

dragonfly/world/entity.go
```diff
@@ -4,14 +4,13 @@ import (
...
 	"github.com/go-gl/mathgl/mgl64"
-	"io"
 )
 
 // Entity represents an entity in the world, typically an object that may be moved around and can be
 // interacted with by other entities.
 // Viewers of a world may view an entity when near it.
 type Entity interface {
-	io.Closer
+	Close() error
```
Sandertv, line 13: Also any reason to change this?

JustTalDevelops: Uhhh no, I have no idea where that came from.

dragonfly/world/world.go
JustTalDevelops: No idea, I don't think I did that though

dragonfly/server.go
JustTalDevelops: It seems to make block breaking feel a lot less laggy. Even if we reverted this, continue destroying seems to have been changed so it would still need some block breaking changes anyways.

dragonfly/player/player.go
JustTalDevelops: Renamed

dragonfly/world/world.go
JustTalDevelops: Removed those changes

dragonfly/world/entity.go
JustTalDevelops: Switched back to io.Closer

dragonfly/server.go
Sandertv: I thought the block breaking was fixed in the previous version, so it'd be weird if that broke now. Either way, if it actually is less laggy I suppose we should look into that. I do think the current way isn't going to cut it though.

dragonfly/session/handler_player_action.go
```diff
@@ -55,11 +56,26 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
...
 
 		s.c.ContinueBreaking(world.Face(face))
+	case protocol.PlayerActionPredictDestroyBlock:
+		s.swingingArm.Store(true)
+		defer s.swingingArm.Store(false)
+
+		s.c.StartBreaking(world.BlockPos{int(pos[0]), int(pos[1]), int(pos[2])}, world.Face(face))
+		s.c.FinishBreaking()
+		if _, ok := s.c.GameMode().(gamemode.Survival); ok {
+			s.c.StartBreaking(world.BlockPos{int(pos[0]), int(pos[1]), int(pos[2])}.Side(world.Face(face).Opposite()), world.Face(face))
```
Sandertv, line 77: I'm a bit concerned over this part. We can't just take the opposite face, because if a player is first destroying the top of a block at an angle, when that block breaks it'll target the one on the side of it, not under it.

dragonfly/session/handler_player_action.go
```diff
@@ -55,11 +56,26 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
 		s.c.AbortBreaking()
 	case protocol.PlayerActionStopBreak:
 		s.c.FinishBreaking()
+	case protocol.PlayerActionContinueDestroyBlock:
+		if s.c.Breaking() {
+			s.c.AbortBreaking()
+			s.c.StartBreaking(world.BlockPos{int(pos[0]), int(pos[1]), int(pos[2])}, world.Face(face))
+		}
```
Sandertv, line 63: This also looks a bit... off, are you sure this isn't just supposed to do the exact same thing as the case below?

JustTalDevelops: Changed

dragonfly/session/handler_player_action.go
JustTalDevelops: Changed

dragonfly/server.go
JustTalDevelops: For sure

dragonfly/session/handler_player_action.go
```diff
@@ -46,13 +46,17 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
 		}
 	case protocol.PlayerActionStopSwimming:
 		s.c.StopSwimming()
+	case protocol.PlayerActionContinueDestroyBlock:
+		fallthrough
```
Sandertv, line 50: I think this is supposed to above the protocol.PlayerActionCrackBreak case, or is it sent as a sort of 'start break' action?

JustTalDevelops: Yeah it's used to start breaking the next block as far as I can tell, hence why I fall through there

Sandertv: Got it, that makes sense.

JustTalDevelops:
Updated gophertunnel to v1.12.1. Anything else needed to be changed?

Sandertv (approved):
Looks good, thank you!

PR #121, opened 2021-04-22
Added minecraft version on startup message

mmm545:


Sandertv:
Maybe adding `for` before the version would be a little better!

PR #123, opened 2021-04-26
Makes several RandomTicker block implementations use the rand.Rand instance provided

Heisenburger69:
Should've gotten everything, let me know if I missed everything. 
Also set the rand.Rand instance to a _ when not required.


Sandertv (approved):
Looks perfect, thank you!

PR #144, opened 2021-05-31
Fix torches and chests not showing up in creative

JustTalDevelops:


Sandertv:
Thank you for the PR! Looks good.

PR #149, opened 2021-06-03
block/sandstone.go: Implemented sandstone blocks.

T14Raptor:


server/block/sandstone_type.go
```diff
@@ -0,0 +1,94 @@
...
+
+type sandstone uint8
+
+// NormalSandstone is the normal variant of sandstone.
+func NormalSandstone() SandstoneType {
+	return SandstoneType{sandstone(0)}
+}
+
+// CutSandstone is the chiseled variant of sandstone.
+func CutSandstone() SandstoneType {
```
Sandertv, line 18: Wrong docs.

server/block/sandstone_type.go
```diff
@@ -0,0 +1,94 @@
...
+	return SandstoneType{sandstone(1)}
+}
+
+// ChiseledSandstone is the cut variant of sandstone.
```
Sandertv, line 22: Also wrong docs.

Issue #146, opened 2021-05-31
Clean up tall grass block types


Sandertv:
The tall grass block is a bit of a mess. The awkward fusion of blocks that can have a double block version and can't just does not work and is rather confusing. This should probably be split up into a couple of other types.

PR #153, opened 2021-06-05
Implemented double plants

DaPigGuy:
- Implemented Sunflower, Lilac, Rose Bush, & Peony blocks
- Moved tall grass & large ferns to its own block (#146)
- Moved nether sprouts to its own block (#146)
- Fixed being able to overwrite blocks by bonemealing grass/ferns
- Fixed being able to overwrite blocks by placing tall grass/large ferns
- Fixed being unable to place grass/ferns/nether sprouts on dirt & coarse dirt
- Fixed grass & ferns not dropping when mined with shears
- Fixed grass/ferns/nether sprouts not being replaceable
- Fixed incorrect flammability values for tall grass & ferns
- Fixed issue with placing doors within replaceable blocks
- Made checks for whether tall grass/large ferns can stay stricter
  - Upper/lower bit must be same type
  - If lower bit, above block must be an upper bit plant, and if upper bit, below block must be a lower bit plant


Sandertv:
This is an improvement to grass, but I'd rather like to see it even less mixed up. If we had a TallGrass and DoubleTallGrass block, both with a `GrassType` field, that would probably be ideal. Then a separate type for (double) flowers. As for the nether sprouts, they should probably be a different block entirely.

server/block/tall_grass.go
```diff
@@ -0,0 +1,115 @@
...
+			w.SetBlock(pos.Side(cube.FaceUp), upper)
+			return true
+		}
+	case Fern():
+		upper := DoubleTallGrass{Type: Fern(), UpperPart: true}
+		if replaceableWith(w, pos.Side(cube.FaceUp), upper) {
+			w.SetBlock(pos, DoubleTallGrass{Type: Fern()})
+			w.SetBlock(pos.Side(cube.FaceUp), upper)
+			return true
+		}
```
Sandertv, line 58: These two cases can be merged by using `g.Type`. That's why the `GrassType` is useful!

server/block/tall_grass.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// FlammabilityInfo ...
+func (g TallGrass) FlammabilityInfo() FlammabilityInfo {
+	return newFlammabilityInfo(30, 100, false)
+}
+
+// BreakInfo ...
+func (g TallGrass) BreakInfo() BreakInfo {
+
```
Sandertv, line 29: Bit of a weird empty line here.

Sandertv (commented):
Looks good! Got two very small nitpicks.

Sandertv (approved):
Looks good, thank you!

PR #156, opened 2021-06-05
Signs

aimjel:
Implementing signs


server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+type WoodSign struct {
+	transparent
+	bass//might be wrong
+
+
+	// Wood is the type of wood of the door. This field must have one of the values found in the material
+	// package.
+	Wood WoodType
+	// Pos is the position of the sign and the block entity
+	Pos cube.Pos
```
Sandertv, line 24: Fields that aren't saved in block NBT or states won't persist properly, so this probably won't as expected. See my hint above. You'll probably want to remove this field.

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+	// Facing is the direction the sign is facing.
+	Facing cube.Direction
+	// Standing is whether the sign is on a block or against
+	Standing bool
+	// Text is an array with the text
+	Text string
+	// SignTextColor no idea what this does
+	SignTextColor int32
```
Sandertv, line 32: You don't need to add fields if you don't know what they do 😉. As for its functionality, it's probably for dyeing signs by clicking them with a dye.

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+	// TextOwner
+	TextOwner string
```
Sandertv, line 34: Don't forget to document this. Is it an XUID?

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Wood is the type of wood of the door. This field must have one of the values found in the material
+	// package.
+	Wood WoodType
+	// Pos is the position of the sign and the block entity
+	Pos cube.Pos
+	// Facing is the direction the sign is facing.
+	Facing cube.Direction
+	// Standing is whether the sign is on a block or against
+	Standing bool
```
Sandertv, line 28: I'm not entirely sure if standing and wall signs should be merged because the facing is different. If you have any thoughts on that let me know!

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// EncodeNBT ...
+func (s WoodSign) EncodeNBT() map[string]interface{} {
+	fmt.Printf("encode sign nbt %#v\n", s)
+	return map[string]interface{}{
+		"id": "Sign",
+		"x": int16(s.Pos.X()),
+		"y": int16(s.Pos.Y()),
+		"z": int16(s.Pos.Z()),
```
Sandertv, line 93: See the comment above for positions. For saving to disk, these are added automatically. I don't think they're necessary for sending over network, but if they are, the position should probably be added automatically for every block entity. This would be done by adding x y and z in the `(*session.Session).ViewBlockUpdate` method.

server/session/handler_block_actor_data.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+)
+
+type BlockActorDataHandler struct {}
+
+func (b BlockActorDataHandler) Handle(p packet.Packet, s *Session) error {
+	pk := p.(*packet.BlockActorData)
+	x, y, z := int(pk.Position.X()), int(pk.Position.Y()), int(pk.Position.Z())
+	switch pk.NBTData["id"]{
```
Sandertv, line 15: Might want to check if "id" exists first so we don't get any potential crashes.

server/session/handler_block_actor_data.go
```diff
@@ -0,0 +1,29 @@
...
+
+	case "Sign":
+		blockPosition := cube.Pos{x, y, z}
+		if v, ok := s.c.World().Block(blockPosition).(block.WoodSign); ok {
+			v.SignTextColor = pk.NBTData["SignTextColor"].(int32)
+			v.Text = pk.NBTData["Text"].(string)
+			v.TextOwner = pk.NBTData["TextOwner"].(string)
```
Sandertv, line 22: Same for these fields here as above.

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+
+// DecodeNBT ...
+func (s WoodSign) DecodeNBT(data map[string]interface{}) interface{} {
+	fmt.Printf("decode sign nbt %#v\n", data)
+	pos := cube.Pos{int(data["x"].(int32)), int(data["y"].(int32)), int(data["z"].(int32))}
+	return WoodSign{Wood: s.Wood, Facing: s.Facing, Standing: s.Standing,
+		Pos: pos,
+		Text: data["Text"].(string),
+		SignTextColor: data["SignTextColor"].(int32),
+		TextOwner: data["TextOwner"].(string),
```
Sandertv, line 82: These could also potentially end up panicking if the fields don't exist.

server/session/handler_block_actor_data.go
```diff
@@ -0,0 +1,29 @@
...
+func (b BlockActorDataHandler) Handle(p packet.Packet, s *Session) error {
+	pk := p.(*packet.BlockActorData)
+	x, y, z := int(pk.Position.X()), int(pk.Position.Y()), int(pk.Position.Z())
+	switch pk.NBTData["id"]{
+
+	case "Sign":
+		blockPosition := cube.Pos{x, y, z}
+		if v, ok := s.c.World().Block(blockPosition).(block.WoodSign); ok {
+			v.SignTextColor = pk.NBTData["SignTextColor"].(int32)
+			v.Text = pk.NBTData["Text"].(string)
```
JustTalDevelops, line 21: Make sure to validate this field and make sure the size is valid.
I think PocketMine checks if the Text is less than 1000, and won't update the block if it is over that threshold.

server/block/wood_sign.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	if s.Standing {
+		return "minecraft:" + woodType + "standing_sign", map[string]interface{}{"ground_sign_direction": int32(s.Facing)}
+	}
+
+	return "minecraft:" + woodType + "wall_sign", map[string]interface{}{"facing_direction": int32(s.Facing)}
+}
+
+func AllSigns() (signs []world.Block) {
```
JustTalDevelops, line 121: Might want to unexport this as it will only be used in register.go

server/block/wood_sign.go
Sandertv: We still need to figure out more thoroughly how this works and its details. We can't really keep a field lingering in the API of which we don't know how it works exactly.

server/block/wood_sign.go
```diff
@@ -0,0 +1,119 @@
...
+	bass //might be wrong
+
+	// Wood is the type of wood of the door. This field must have one of the values found in the material
+	// package.
+	Wood WoodType
+	// Facing is the direction the sign is facing.
+	Facing cube.Direction
+	// Standing is whether the sign is on a block or against
+	Standing bool
+	// Text is an array with the text
```
Sandertv, line 25: Make sure to fix the docs here.

server/block/wood_sign.go
```diff
@@ -0,0 +1,119 @@
...
+	return water
+}
+
+func (s WoodSign) UseOnBlock(pos cube.Pos, face cube.Face, clickPos mgl64.Vec3, w *world.World, user item.User, ctx *item.UseContext) (used bool) {
+	pos = pos.Side(face)
+	if face != cube.FaceDown {
+
+		if face == cube.FaceUp {
+			s.Standing = true
+			s.Facing = cube.Direction(int16(math.Floor((((user.Yaw()+180)*16/360)+0.5)*100)/100) & 15)
```
Sandertv, line 60: I don't think it's a good idea to merge this into the Facing field. The cube.Direction type can't have up to 16 values, it only knows of north, east, south and west. I think we need a separate type for this.

server/block/wood_sign.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return placed(ctx)
+}
+
+// DecodeNBT ...
+func (s WoodSign) DecodeNBT(data map[string]interface{}) interface{} {
+	return WoodSign{Wood: s.Wood, Facing: s.Facing, Standing: s.Standing,
+		Text:          data["Text"].(string),
+		SignTextColor: data["SignTextColor"].(int32),
+		TextOwner:     data["TextOwner"].(string),
```
Sandertv, line 74: This can still panic if any of these fields aren't present in the NBT.

server/session/handler_block_actor_data.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/df-mc/dragonfly/server/block"
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+	"strings"
+	"unicode/utf8"
+)
+
+type BlockActorDataHandler struct{}
+
+func (b BlockActorDataHandler) Handle(p packet.Packet, s *Session) error {
```
Sandertv, line 13: Make sure to add docs over these.

server/session/handler_block_actor_data.go
```diff
@@ -0,0 +1,48 @@
...
+						if len(text) >= 100 {
+							text = text[:100]
+						}
+						v.Text = text
+					}
+				}
+
+				if pkTextOwner, ok := pk.NBTData["TextOwner"]; ok {
+					v.TextOwner = pkTextOwner.(string)
+				}
```
Sandertv, line 35: Should probably not use the data the user provides here. Set this field using the XUID of the player.

Sandertv (changes requested):

Sandertv:
Thanks a lot for the PR! Great work!
I'll merge this into a dedicated branch as discussed on Discord to make sure the last small things are fixed up. It'll be merged before the next release!

PR #168, opened 2021-06-19
player/player.go: Implemented runtime player skin changing.

### LINKED ISSUE #135: Implement runtime player skin changing
Skin changes are currently not handled, so changing skin in-game won't allow other people to see it. `*packet.PlayerSkin` should have a handler to handle skin changes.

Skin changing should probably have a `HandleSkinChange(old, new)` method in the `player.Handler` so that people can listen for skin changes. 

One thing to consider is that skin changes should be propagated to all players. We could either send the new skin to all connected sessions, or we could send a `*packet.PlayerSkin` only to viewers of the player that changes skin and always send a `*packet.PlayerSkin` when a player sees another player after it entering the view distance.

T14Raptor:
Closes #135


server/player/player.go
```diff
@@ -175,9 +176,33 @@ func (p *Player) Addr() net.Addr {
 // is shown to.
 // If the player was not connected to a network session, a default skin will be set.
 func (p *Player) Skin() skin.Skin {
+	p.skinMu.RLock()
+	defer p.skinMu.RUnlock()
 	return p.skin
 }
 
+// ChangeSkin ...
```
T14Raptor, line 184: Documentation is needed here

server/player/player.go
```diff
@@ -175,9 +176,33 @@ func (p *Player) Addr() net.Addr {
...
+func (p *Player) ChangeSkin(skin skin.Skin) {
```
Sandertv, line 185: Might be an idea to change this method name to `SetSkin` instead?

T14Raptor:
This is ready for review now.

Sandertv (approved):
Looks good to me!

PR #170, opened 2021-06-20
Added logger interface

### LINKED ISSUE #169: Make Dragonfly accept any logger that implements an interface
Currently Dragonfly only accepts Logrus loggers for logging. This isn't ideal for obvious reasons. We should probably have a logger interface somewhere and accept any value that implements it.

Thunder33345:
Added logger interface to be able to accept any logger
Closes #169


Sandertv (approved):
Looks good, thank you!

PR #180, opened 2021-06-23
Add all 16 dyes

provsalt:


Sandertv:
Would probably be a better idea to merge these into a single `Dye` with a colour field. Currently the colour type is in the block package though, so maybe move that into the item package?

provsalt:
Changed.

Sandertv (approved):

PR #181, opened 2021-06-23
Send world spawn to players when necessary

### LINKED ISSUE #150: Send world spawn to players when necessary
Currently the world spawn isn't sent to the player after its initial spawn. The spawn position should be updated on world change and when the position of the world that the player is in is changed.

AndreasHGK:
Closes #150

### DIFF
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -750,7 +750,7 @@ func (s *Session) ViewWorldSpawn(pos cube.Pos) {
 	s.writePacket(&packet.SetSpawnPosition{
 		SpawnType:     packet.SpawnTypeWorld,
 		Position:      blockPos,
-		Dimension:     1,
+		Dimension:     packet.DimensionOverworld,
 		SpawnPosition: blockPos,
 	})
 }
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -744,6 +744,17 @@ func (s *Session) ViewEmote(player world.Entity, emote uuid.UUID) {
 	})
 }
 
+// ViewWorldSpawn ...
+func (s *Session) ViewWorldSpawn(pos cube.Pos) {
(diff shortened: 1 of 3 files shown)


AndreasHGK:
Should I also add a method for this in the Player struct so people can update the spawn position for instance to point a compass somewhere? If so I'll do that otherwise you can review this PR

server/world/world.go
```diff
@@ -1451,6 +1454,7 @@ func (w *World) addViewer(c *chunkData, viewer Viewer) {
 	}
 
 	viewer.ViewTime(w.Time())
+	viewer.ViewWorldSpawn(w.Spawn())
```
Sandertv, line 1457: You might want to move this to the end of `(*World*).addWorldViewer()`. The `ViewTime` call is actually misplaced too, as this function is called for every chunk loaded. Would you mind moving that to the same function too?
Thanks!

server/session/world.go
```diff
@@ -744,6 +744,17 @@ func (s *Session) ViewEmote(player world.Entity, emote uuid.UUID) {
 	})
 }
 
+// ViewWorldSpawn ...
+func (s *Session) ViewWorldSpawn(pos cube.Pos) {
+	blockPos := protocol.BlockPos{int32(pos[0]), int32(pos[1]), int32(pos[2])}
+	s.writePacket(&packet.SetSpawnPosition{
+		SpawnType:     packet.SpawnTypeWorld,
+		Position:      blockPos,
+		Dimension:     1,
```
Sandertv, line 753: I think 1 is the nether, not sure if that's right? Maybe this works too though, I don't know.

AndreasHGK: You are right, I was looking at PM for this but I looked at the wrong file, woops 😅

server/world/world.go
AndreasHGK: Done!

Sandertv (approved):

PR #187, opened 2021-06-23
Implemented ladders

DaPigGuy:


server/player/player.go
```diff
@@ -840,6 +841,27 @@ func (p *Player) SetMobile() {
...
+// SetCanClimb allows a player to climb ladders & vines.
+func (p *Player) SetCanClimb() {
+	if !p.canClimb.CAS(false, true) {
+		return
+	}
+	p.updateState()
+}
+
+// SetCantClimb disallows a player from climbing ladders & vines.
+func (p *Player) SetCantClimb() {
```
Sandertv, line 858: These have rather awkward names. Maybe `EnableClimbing` and `DisableClimbing` would be better? I'm actually not sure if exposing these methods has any real use at all. Is there a good example where you'd want to disable climbing for players?

server/block/ladder.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Model ...
+func (l Ladder) Model() world.BlockModel {
+	return model.Ladder{Facing: l.Facing}
+}
+
+// allLadders ...
+func allLadders() (b []world.Block) {
+	for i := cube.Direction(0); i <= 3; i++ {
```
Sandertv, line 85: Could use `cube.Directions()` for this too.

Sandertv (approved):
Looks good, thank you!

PR #188, opened 2021-06-24
Implement basic player data provider

### LINKED ISSUE #132: Add player data storing to dragonfly

AndreasHGK:
EDIT:
closes #132 

This implements a player data provider interface, a NopProvider and a player provider that uses a LevelDB database in order to save and load player data.


AndreasHGK:
Edit: this PR isnt done, I decided I should probably not split it as there are currently some bugs and some things I'd like to still improve

AndreasHGK:
Currently everything works, however I cannot find a good way to load all the playerdata. The places where the player gets playerdata is very spread out and if one thing works in a certain place something else seems to not work. In addition I can't add items to the inventories when the sessions hasn't started yet. Does anyone have any suggestions?

server/player/player.go
```diff
@@ -1909,7 +1913,82 @@ func (p *Player) close() {
...
+// returns false.
+func (p *Player) Load() {
+	data, ok := p.provider.Load(p.UUID())
+	if !ok {
+		return
+	}
+
+	//p.pos.Store(data.Position)
+	p.Teleport(data.Position.Sub(mgl64.Vec3{0.5, 0, 0.5}))
+	p.Rotate(data.Yaw-p.yaw.Load(), data.Pitch-p.pitch.Load())
```
JustTalDevelops, line 1930: Perhaps this yaw and pitch should be sent in the start game packet, including the position, so you don't have to teleport the player?

AndreasHGK: Thanks, that does work!

AndreasHGK:
This PR should now be fully functional, I have tested pretty much everything now :D

server/player/player.go
```diff
@@ -90,11 +90,13 @@ type Player struct {
...
 
 	hunger *hungerManager
+
+	provider Provider
 }
 
 // New returns a new initialised player. A random UUID is generated for the player, so that it may be
 // identified over network.
-func New(name string, skin skin.Skin, pos mgl64.Vec3) *Player {
+func New(name string, skin skin.Skin, pos mgl64.Vec3, data *Data) *Player {
```
JustTalDevelops, line 99: Does this data need to be a pointer?

server/player/provider/leveldb.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &DBProvider{db: db}, nil
+}
+
+// Save ...
+func (p *DBProvider) Save(d player.Data) {
+	data := toJson(d)
+	jsondata, err := json.Marshal(data)
```
JustTalDevelops, line 32: Should this data really be in JSON? For blocks, their states might load incorrectly since we wouldn't know the value types

server/server.go
```diff
@@ -71,15 +73,21 @@ func New(c *Config, log internal.Logger) *Server {
...
+		p:              make(map[uuid.UUID]*player.Player),
+		name:           *atomic.NewString(c.Server.Name),
+		playerProvider: player.NopProvider{},
 	}
 	s.JoinMessage(c.Server.JoinMessage)
 	s.QuitMessage(c.Server.QuitMessage)
+	if p, err := provider.NewDBProvider("players"); err == nil {
+		s.PlayerProvider(p)
+	} else {
+		log.Errorf("Error starting player provider: %v", err)
```
JustTalDevelops, line 89: Perhaps you should return this error or panic? I imagine player data saving would be a pretty critical component

server/player/player.go
AndreasHGK: My reasoning to make the data a pointer is that players could pass nil if they want to use default values, for example since it seems you can make a player without a session I can imagine people might want to use the default values and when a new player joins they dont have any data.

An alternative to this is I could have an exported function like `player.DefaultData()` or something like that and instead pass that on if no data was found.

server/player/provider/leveldb.go
AndreasHGK: That is indeed a problem, I wasn't really sure how else to store it. If you know a good way what do you think would be a way to store it that keeps the value types?

JustTalDevelops: NBT should store types, we already use it for creative items anyways

server/server.go
AndreasHGK: Okie! I added a panic and I also added an option in the config to disable the default provider, that way players who dont want it can explicitly say so, and it can prevent a database being created for servers that don't use it.

server/player/data.go
```diff
@@ -0,0 +1,69 @@
...
+	// These are currently not implemented in DF.
+	XpLevel, XpTotal int
+	// XpPercentage is the player's current progress towards the next level.
+	// This is currently not implemented in DF.
+	XpPercentage float64
+	// XpSeed is the random seed used to determine the next enchantment in enchantment tables.
+	// This is currently not implemented in DF.
+	XpSeed int
+	// Gamemode is the last gamemode the user had, like creative or survival.
+	Gamemode world.GameMode
```
T14Raptor, line 45: GameMode*

server/player/data.go
```diff
@@ -0,0 +1,69 @@
...
+	FallDistance float64
+}
+
+// InventoryData is a struct that contains all data of the player inventories.
+type InventoryData struct {
+	// Items contains all the items in the player's main inventory.
+	// This excludes armor and offhand.
+	Items []item.Stack
+	// Armour contains all armor items the player is wearing.
+	Armour [4]item.Stack
```
T14Raptor, line 63: Maybe separate this into 4 fields (Helmet, Chestplate, etc.)? It isn't obvious what order it's in.

server/player/data.go
```diff
@@ -0,0 +1,69 @@
...
+	// Offhand is what the player is carrying in their non-main hand, like a shield or arrows.
+	Offhand item.Stack
```
T14Raptor, line 65: OffHand*

server/player/data.go
```diff
@@ -0,0 +1,69 @@
...
+	// Mainhand saves the slot in the hotbar that the player is currently switched to.
+	// Should be between 0-8.
+	Mainhand uint32
```
T14Raptor, line 68: MainHand*

server/player/provider/gamemode.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	survival = uint8(iota)
+	creative
+	adventure
+	spectator
+)
+
+func gamemodeToData(mode world.GameMode) uint8 {
+	gm := survival
+	if _, ok := mode.(world.GameModeCreative); ok {
```
T14Raptor, line 14: A switch statement could be used here.

server/player/provider/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if b, ok := stack.Item().(world.Block); ok {
+		data["block"], data["block_properties"] = b.EncodeBlock()
+	} else {
+		data["item"], data["item_meta"] = stack.Item().EncodeItem()
+	}
+
+	data["count"] = stack.Count()
+	if len(stack.Values()) > 0 {
+		data["nbt"] = stack.Values()
```
T14Raptor, line 66: Could this key be "dragonflyData" instead of "nbt"? Would be more consistent with nbtconv.

server/player/provider/leveldb.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return
+	}
+	_ = p.db.Put([]byte(data.UUID), jsondata, nil)
+}
+
+// Load ...
+func (p *DBProvider) Load(UUID uuid.UUID) (player.Data, bool) {
+	jsondata, err := p.db.Get([]byte(UUID.String()), nil)
```
T14Raptor, line 41: Raw UUID could be used here. (`UUID[:]`)

server/player/player.go
AndreasHGK: I asked around and this should be fine according to Sander so ill leave this one as resolved 😄

server/player/player.go
```diff
@@ -1912,6 +1920,76 @@ func (p *Player) close() {
...
+	_ = p.offHand.SetItem(1, data.OffHand)
+	p.Armour().SetBoots(data.Boots)
+	p.Armour().SetLeggings(data.Leggings)
+	p.Armour().SetChestplate(data.Chestplate)
+	p.Armour().SetHelmet(data.Helmet)
+}
+
+// GetSaveData returns the player data that needs to be saved. This is used when the player
+// gets disconnected and the player provider needs to save the data.
+func (p *Player) GetSaveData() Data {
```
Sandertv, line 1960: Would call this just `Data` or maybe `SaveData`

server/player/data.go
```diff
@@ -0,0 +1,72 @@
...
+	// Boots, Leggings, Chestplate, Helmet are armor pieces that belong to the slot corresponding to the name.
+	Boots      item.Stack
+	Leggings   item.Stack
+	Chestplate item.Stack
+	Helmet     item.Stack
+	// OffHand is what the player is carrying in their non-main hand, like a shield or arrows.
+	OffHand item.Stack
+	// MainHand saves the slot in the hotbar that the player is currently switched to.
+	// Should be between 0-8.
+	MainHand uint32
```
Sandertv, line 71: Would maybe change this to be called `MainHandSlot`

server/player/data.go
```diff
@@ -0,0 +1,72 @@
...
+	// XpLevel is the current xp level the player has, XpTotal is the total amount of xp the
+	// player has collected during their lifetime, which is used to display score upon player death.
+	// These are currently not implemented in DF.
+	XpLevel, XpTotal int
+	// XpPercentage is the player's current progress towards the next level.
+	// This is currently not implemented in DF.
+	XpPercentage float64
+	// XpSeed is the random seed used to determine the next enchantment in enchantment tables.
+	// This is currently not implemented in DF.
+	XpSeed int
```
Sandertv, line 43: Would change all these to `XP` instead of `Xp`

server/player/provider/gamemode.go
```diff
@@ -0,0 +1,38 @@
...
+	gm := survival
+	switch mode.(type) {
+	case world.GameModeCreative:
+		gm = creative
+	case world.GameModeAdventure:
+		gm = adventure
+	case world.GameModeSpectator:
+		gm = spectator
+	}
+	return gm
```
DaPigGuy:
```suggestion
	switch mode.(type) {
	case world.GameModeCreative:
		return creative
	case world.GameModeAdventure:
		return adventure
	case world.GameModeSpectator:
		return spectator
	}
	return survival
```

server/player/player.go
```diff
@@ -122,20 +123,27 @@ func New(name string, skin skin.Skin, pos mgl64.Vec3) *Player {
...
+// you can leave the data as nil to use default data.
+func NewWithSession(name, xuid string, uuid uuid.UUID, skin skin.Skin, s *session.Session, pos mgl64.Vec3, data *Data) *Player {
+	p := New(name, skin, pos, data)
 	p.s, p.uuid, p.xuid, p.skin = s, uuid, xuid, skin
 	p.inv, p.offHand, p.armour, p.heldSlot = s.HandleInventories()
 	p.locale, _ = language.Parse(strings.Replace(s.ClientData().LanguageCode, "_", "-", 1))
-
 	chat.Global.Subscribe(p)
+	if data != nil {
+		p.loadInventory(data.Inventory)
```
Sandertv, line 145: Might want to document that this is overwritten by s.HandleInventories(). Someone might remove this on accident if it's not clear why this is here.

server/player/playerdb/leveldb.go
```diff
@@ -0,0 +1,58 @@
...
+	err = json.Unmarshal(jsondata, &d)
+	if err != nil {
+		return player.Data{}, err
+	}
+
+	return fromJson(d), nil
+}
+
+// Close ...
+func (p *Provider) Close() {
```
Sandertv, line 56: Would return an error here.

server/player/provider.go
```diff
@@ -0,0 +1,32 @@
...
+// writing of the player data so that the Player may use it.
+type Provider interface {
+	// Save is called when the player leaves the server and passes on the current player data.
+	Save(UUID uuid.UUID, data Data) error
+	// Load is called when the player joins and passes the UUID of the player.
+	// It expects to the player data, and a bool that indicates whether or not the player has played before.
+	// If this bool is false the player will use default values and you can use an empty Data struct.
+	Load(UUID uuid.UUID) (Data, error)
+	// Close is called when the server closes and is useful to safely close your database.
+	Close()
```
Sandertv, line 15: Would return an error here to implement `io.Closer` (or you can embed `io.Closer` to do that)

server/server.go
```diff
@@ -365,16 +396,25 @@ func (server *Server) checkNetIsolation() {
 // handleSessionClose handles the closing of a session. It removes the player of the session from the server.
 func (server *Server) handleSessionClose(controllable session.Controllable) {
 	server.playerMutex.Lock()
+	if p, ok := server.p[controllable.UUID()]; ok {
+		err := server.playerProvider.Save(controllable.UUID(), p.Data())
+		if err != nil {
+			server.log.Errorf("Error while saving data: %v", err)
+		}
```
Sandertv, line 403: Would consider moving this outside of the mutex lock so you don't have to keep that locked while saving data.

Sandertv (commented):
Just a couple more comments!

server/server.go
```diff
@@ -364,17 +398,26 @@ func (server *Server) checkNetIsolation() {
 
 // handleSessionClose handles the closing of a session. It removes the player of the session from the server.
 func (server *Server) handleSessionClose(controllable session.Controllable) {
+	if p, ok := server.p[controllable.UUID()]; ok {
```
Sandertv, line 401: This map access does need to be protected by the Mutex lock. If you just change it to get p, ok := server.p[...] inside of the mutex lock and save the data outside of it, you should be good to go.

AndreasHGK: woops ur right 😅

server/player/player.go
```diff
@@ -93,8 +93,9 @@ type Player struct {
 }
 
 // New returns a new initialised player. A random UUID is generated for the player, so that it may be
-// identified over network.
-func New(name string, skin skin.Skin, pos mgl64.Vec3) *Player {
+// identified over network. You can either pass on player data you want to load or
+// you can leave the data as nil to use default data.
+func New(name string, skin skin.Skin, pos mgl64.Vec3, data *Data) *Player {
```
Sandertv, line 98: Now that I think of it, it might be a good idea to just keep the Data out of the New function and only have it in NewWithSession. I can't think of a time where this would be useful for just a normal player entity.

Sandertv: That way the thing with having to load the inventory twice is cleaned up as well.

server/player/player.go
```diff
@@ -122,20 +123,29 @@ func New(name string, skin skin.Skin, pos mgl64.Vec3) *Player {
...
 
 // NewWithSession returns a new player for a network session, so that the network session can control the
 // player.
 // A set of additional fields must be provided to initialise the player with the client's data, such as the
-// name and the skin of the player.
-func NewWithSession(name, xuid string, uuid uuid.UUID, skin skin.Skin, s *session.Session, pos mgl64.Vec3) *Player {
-	p := New(name, skin, pos)
+// name and the skin of the player. You can either pass on player data you want to load or
+// you can leave the data as nil to use default data.
+func NewWithSession(name, xuid string, uuid uuid.UUID, skin skin.Skin, s *session.Session, pos mgl64.Vec3, data *Data) *Player {
```
Sandertv, line 138: Can remove the pointer here now as well because NewWithSession will generally always have Data.

server/player/playerdb/leveldb.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"encoding/json"
+	"github.com/df-mc/dragonfly/server/player"
+	"github.com/df-mc/goleveldb/leveldb"
+	"github.com/google/uuid"
+	"os"
+)
+
+// Provider ...
```
Sandertv, line 11: Please document this! Just describe how data is stored (leveldb, json data etc.)

Sandertv (changes requested):
I've got a couple (for real, this time) last comments. Sorry for not spotting these before!

Sandertv:
Thanks a lot for the PR! 😄

PR #194, opened 2021-06-29
Implemented silk touch

DaPigGuy:


server/block/leaves.go
```diff
@@ -75,8 +75,8 @@ func (l Leaves) FlammabilityInfo() FlammabilityInfo {
 func (l Leaves) BreakInfo() BreakInfo {
 	return newBreakInfo(0.2, alwaysHarvestable, func(t tool.Tool) bool {
 		return t.ToolType() == tool.TypeShears || t.ToolType() == tool.TypeHoe
-	}, func(t tool.Tool) []item.Stack {
-		if t.ToolType() == tool.TypeShears { // TODO: Silk Touch
+	}, func(stack item.Stack) []item.Stack {
+		if _, ok := stack.Item().(item.Shears); ok || hasSilkTouch(stack) {
```
Sandertv, line 79: I'm not sure if this change from `tool.Tool` to `item.Stack` is a good one. This change right here makes it impossible to have functional custom shears. Is there some way we can check for silk touch on tools? Maybe we can even pass an additional bool to this drop function which indicates silk touch vs no silk touch?

server/item/enchantment/silk_touch.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// CompatibleWith ...
+func (e SilkTouch) CompatibleWith(s item.Stack) bool {
+	it := s.Item()
+	_, pickaxe := it.(item.Pickaxe)
+	_, axe := it.(item.Axe)
+	_, shovel := it.(item.Shovel)
+	_, hoe := it.(item.Hoe)
+	_, shears := it.(item.Shears)
```
Sandertv, line 30: Might want to simply check if it's a tool.Tool and check if its tool type is one of these 5. That way custom tools can have silk touch too.

Sandertv: The protection enchantments should probably be changed to check if they implement the armour interfaces instead of concrete types too.

Sandertv (changes requested):
One last comment, looks good otherwise!

Sandertv (approved):

PR #201, opened 2021-07-02
add a SwingArm event

Prim69:


Sandertv:
Make sure to run go fmt over your changes.

Prim69:
going to replace this with an event for punching air

PR #210, opened 2021-07-19
Introduce player handlers for the opening and closing of block containers

DenielWorld:
Added HandleContainerOpen() and HandleContainerClose() methods to the player Handler interface.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -1581,7 +1581,17 @@ func (p *Player) OpenBlockContainer(pos cube.Pos) {
 	if p.session() == session.Nop {
 		return
 	}
-	p.session().OpenBlockContainer(pos)
+	ctx := event.C()
+	p.handler().HandleContainerOpen(ctx, pos)
+
+	ctx.Continue(func() {
+		p.session().OpenBlockContainer(pos)
+	})
+}
+
+// CloseCurrentBlockContainer handles the closing of a block container by the player at the position passed.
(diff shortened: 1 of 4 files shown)


Sandertv:
Could you give a concrete example of when these would be useful?

server/player/player.go
```diff
@@ -1581,7 +1581,17 @@ func (p *Player) OpenBlockContainer(pos cube.Pos) {
...
+	p.handler().HandleContainerOpen(ctx, pos)
+
+	ctx.Continue(func() {
+		p.session().OpenBlockContainer(pos)
+	})
+}
+
+// CloseCurrentBlockContainer handles the closing of a block container by the player at the position passed.
+func (p *Player) CloseCurrentBlockContainer(pos cube.Pos) {
+	p.handler().HandleContainerClose(pos)
```
Sandertv, line 1594: This method does nothing by itself. Code in `session/player.go` should probably be moved here or this method name will be inaccurate.

T14Raptor:
@DenielWorld Any update?

Sandertv:
Any update?

UnknownOre:
> Could you give a concrete example of when these would be useful?
this could be useful if we wanted to forbid the player from opening a specific  container

JustTalDevelops:
Closing this because aforementioned functionality can be achieved the same way using a UseOnBlock handler.

PR #215, opened 2021-07-20
Fixed block picking retaining all block states (#209)

### LINKED ISSUE #209: Block picking retains block states
Picking any blocks retains block states it should not retain

Notable examples are:
- Upper Parts of Double Plants/Doors
- Open Doors/Trapdoors/Fence Gates
- Crops at various growth stages
- Sign Text

Only reproducible when item is not already in inventory

DaPigGuy:
Closes #209

### DIFF
--- a/server/block/hash.go
+++ b/server/block/hash.go
@@ -65,6 +65,7 @@ const (
 	hashIronBlock
 	hashIronOre
 	hashKelp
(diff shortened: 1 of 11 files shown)


Sandertv (approved):
Looks good to me.

PR #231, opened 2021-08-07
Animation a block breating after hitting a block with a sword 

marfush:
The problem is described in the title

The whole problem is the extra "!" :clown_face:


Sandertv:
Good catch, thank you!

PR #232, opened 2021-08-10
Ray trace API

Sandertv:


server/entity/physics/trace/result.go
```diff
@@ -0,0 +1,19 @@
...
+// Result represents the result of a ray-trace collision with a bounding box.
+type Result interface {
+	// AABB returns the bounding box collided with.
+	AABB() physics.AABB
+	// Position ...
+	Position() mgl64.Vec3
+	// Face ...
+	Face() cube.Face
+
+	__()
```
Sandertv, line 18: I think you can remove this. It's okay if other people implement this interface because that doesn't allow anything that doesn't work.

server/entity/physics/trace/result.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Result represents the result of a ray-trace collision with a bounding box.
+type Result interface {
+	// AABB returns the bounding box collided with.
+	AABB() physics.AABB
+	// Position ...
+	Position() mgl64.Vec3
+	// Face ...
+	Face() cube.Face
```
Sandertv, line 16: You should probably document both these functions.

server/entity/physics/trace/entity.go
```diff
@@ -0,0 +1,54 @@
...
+// Entity returns the entity that was collided with.
+func (r EntityResult) Entity() world.Entity {
+	return r.entity
+}
+
+// EntityIntercept performs a ray trace and calculates the point on the entities bounding box's edge nearest to the start position
+// that the ray trace collided with.
+// EntityIntercept returns a EntityResult with the entity collided with and with the colliding vector closest to the start position,
+// if no colliding point was found, it returns nil.
+func EntityIntercept(e world.Entity, start, end mgl64.Vec3) Result {
```
Sandertv, line 43: Consider returning (EntityResult, bool) instead.

server/entity/physics/trace/block.go
```diff
@@ -0,0 +1,75 @@
...
+// BlockPosition ...
+func (r BlockResult) BlockPosition() cube.Pos {
+	return r.blockPos
+}
+
+// BlockIntercept performs a ray trace and calculates the point on the block model's edge nearest to the start position
+// that the ray-trace collided with.
+// BlockIntercept returns a BlockResult with the block collided with and with the colliding vector closest to the start position,
+// if no colliding point was found, it returns nil.
+func BlockIntercept(pos cube.Pos, w *world.World, b world.Block, start, end mgl64.Vec3) Result {
```
Sandertv, line 44: Consider returning `(BlockResult, bool)` instead.

server/entity/physics/trace/aabb.go
```diff
@@ -0,0 +1,149 @@
...
+// Face ...
+func (r AABBResult) Face() cube.Face {
+	return r.face
+}
+
+// Intercept performs a ray trace and calculates the point on the AABB's edge nearest to the start position that the ray trace
+// collided with.
+// Intercept returns a AABBResult with the colliding vector closest to the start position, if no colliding point was found,
+// it returns nil.
+func Intercept(bb physics.AABB, start, end mgl64.Vec3) Result {
```
Sandertv, line 36: Consider returning `(AABBResult, bool)`.

Sandertv (commented):
I left a couple of comments. Could you give an example of how you'd use this API to, say, traverse over a distance of 10 blocks in some direction? Do you reckon it's possible and reasonable to export a function that combines everything?

server/entity/physics/trace/entity.go
T14Raptor: I kind of don't like the fact that a zero EntityResult will have to be returned.

server/entity/physics/trace/trace.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// divideVec3 ...
+func divideVec3(v1, v2 mgl64.Vec3) mgl64.Vec3 {
+	return mgl64.Vec3{divide(v1[0], v2[0]), divide(v1[1], v2[1]), divide(v1[2], v2[2])}
+}
+
+// divide ...
+func divide(f1, f2 float64) float64 {
```
Sandertv, line 61: Would maybe call this `safeDivide` or something like that. (Same for the vec3 version above)

server/entity/physics/trace/trace.go
```diff
@@ -0,0 +1,103 @@
...
+	return f1 / f2
+}
+
+// boundaryVec3 ...
+func boundaryVec3(v1, v2 mgl64.Vec3) mgl64.Vec3 {
+	return mgl64.Vec3{boundary(v1[0], v2[0]), boundary(v1[1], v2[1]), boundary(v1[2], v2[2])}
+}
+
+// boundary ...
+func boundary(s, d float64) float64 {
```
Sandertv, line 74: Would document this briefly.

server/entity/physics/trace/aabb.go
```diff
@@ -0,0 +1,147 @@
...
+	f := (y - a[1]) / yDiff
+	if f < 0 || f > 1 {
+		return nil
+	}
+
+	return &mgl64.Vec3{a[0] + (b[0]-a[0])*f, y, a[2] + (b[2]-a[2])*f}
+}
+
+// intermediateZ ...
+func intermediateZ(a, b mgl64.Vec3, z float64) *mgl64.Vec3 {
```
Sandertv, line 135: Would document (one of) these, it's not clear right away what these do.

Sandertv (commented):
Just a couple last comments for documentation, looks good otherwise!

Issue #134, opened 2021-05-18
Add a method to strike lightning in the world


Sandertv:
Lightning has traditionally been an entity, but it would probably be better if there was a simple method to strike lightning in a world. Something like `(*world.World).StrikeLightning(pos)` might work.

T14Raptor:
Pushing this back.

Sandertv:
Postponing to v0.3.0.

Sandertv:
Random thoughts: Making this an entity might be easier after all. It needs to have access to specific blocks (fire) and needs to damage and set on fire entities around it. This would come with quite some hacks if added to the `world` package.

PR #86, opened 2020-10-23
implemented basic lightning bolt entity

xerenahmed:

PR #233, opened 2021-08-10
Added lightning entity and handling

Lucaskyy:
This PR closes: #134

Tasks:
- [x] Lightning Entity
- [x] Entity Handling
- [x] Setting blocks on fire on struck
- [x] Testing the entity

If there's anything out of the ordinary, or you want it changed, please let me know!
Also note, this is my first time working in the Bedrock environment and my first PR here, so please do let me know if anything is wrong so I can learn from them in the future :) While making this, I looked at CloudBurst's code to get a better understanding of thunder in Bedrock edition.


server/entity/lightning.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if li.state >= 0 {
+		bb := li.AABB().Grow(3)
+		bb.Extend(mgl64.Vec3{bb.Max().X() + 6})
+
+		for _, e := range li.World().CollidingEntities(bb) {
+			if l, ok := e.(Living); ok && l.Health() <= 0 { // there's no point in damaging entities that are already dead
+				l.Hurt(5, damage.SourceLightning{})
+				if f, ok := e.(Flammable); ok && f.OnFireDuration() < 8*20 {
+					f.SetOnFire(8)
```
Sandertv, line 114: These functions accept a `time.Duration` which is not equivalent to ticks or seconds, but nanoseconds. If you meant for these to be 8 seconds, use `time.Second * 8`.

Lucaskyy:
Also, when this PR does get merged, make sure to close #86 because it is not needed anymore.

server/block/fire.go
```diff
@@ -1,12 +1,16 @@
...
+
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/entity"
 	"github.com/df-mc/dragonfly/server/entity/damage"
 	"github.com/df-mc/dragonfly/server/world"
+	"github.com/go-gl/mathgl/mgl64"
 	"math/rand"
 	"time"
+	_ "unsafe" // Imported for compiler directives.
```
Sandertv, line 13: I would like to to refrain from using any more go:linkname compiler directives. We had a bit of a journey to try and get rid of them all a while back and I think it's best if we don't go down that rabbit hole again. Compiler directives make it very difficult to refactor things.

Do you reckon we can approach this in some other way? I understand the circular dependencies are a bit of a problem, but if we can avoid using compiler directives that would be very helpful in the future.

server/entity/lightning.go
```diff
@@ -0,0 +1,123 @@
...
+	return w
+}
+
+// AABB ...
+func (Lightning) AABB() physics.AABB {
+	return physics.NewAABB(mgl64.Vec3{}, mgl64.Vec3{})
+}
+
+// State ...
+func (li *Lightning) State() int {
```
Sandertv, line 50: Is there any use for this to be exported? If not, just remove the method.

server/entity/lightning.go
```diff
@@ -0,0 +1,123 @@
...
+	return li.state
+}
+
+// Close closes the lighting.
+func (li *Lightning) Close() error {
+	if li.World() != nil {
```
Sandertv, line 56: Can remove this check as `*world.World(nil).RemoveEntity` checks for it too.

server/entity/lightning.go
```diff
@@ -0,0 +1,123 @@
...
+
+			setBlocksOnFire(li.World(), li.Position())
+		}
+	}
+}
+
+func nextInt(max int) int {
+	rand.Seed(time.Now().UnixNano())
+	return rand.Intn(max)
+}
```
Sandertv, line 121: This function has a couple of problems, the first being that `nextInt` doesn't make clear that it is random (it's a bit of a garbage name from Java edition). The other being that the seeding everytime is a problem. If you strike lightning, say, 1000 times in a second, all of those will have the same int. I'd just use `rand.Intn(max)` everywhere and remvoe this function, then make sure `rand.Seed(time.Now().UnixNano())` is called somewhere in an `init` function.

server/entity/lightning.go
```diff
@@ -0,0 +1,123 @@
...
+		li.World().PlaySound(li.Position(), sound.Explosion{})
+
+		bb := li.AABB().Grow(9).Extend(mgl64.Vec3{6})
+		for _, e := range li.World().CollidingEntities(bb) {
+			if l, ok := e.(Living); ok && l.Health() > 0 { // there's no point in damaging entities that are already dead
+				l.Hurt(5, damage.SourceLightning{})
+				if f, ok := e.(Flammable); ok && f.OnFireDuration() < 8*20 {
+					f.SetOnFire(time.Second * 8)
+				}
+				// TODO Check if the entity is a creeper, if so, make it an supercharged creeper
```
Sandertv, line 96: I think you can remove this TODO here. The creepers can do this check themselves because of `damage.SourceLightning`.

server/entity/lightning.go
Lucaskyy: No, I'll remove it.

server/entity/lightning.go
Lucaskyy: I understand, will do that.

server/block/fire.go
```diff
@@ -232,3 +236,26 @@ func allFire() (b []world.Block) {
...
+//go:linkname block_setBlocksOnFire github.com/df-mc/dragonfly/server/entity.setBlocksOnFire
+//noinspection ALL
+var block_setBlocksOnFire func(w *world.World, lPos mgl64.Vec3)
+
+func init() {
+	block_setBlocksOnFire = func(w *world.World, lPos mgl64.Vec3) {
+		_, isNormal := w.Difficulty().(world.DifficultyNormal)
+		_, isHard := w.Difficulty().(world.DifficultyHard)
+		if isNormal || isHard { // difficulty >= 2
+			bPos := cube.Pos{int(lPos.X()), int(lPos.Y()), int(lPos.Z())}
```
Sandertv, line 249: There should be a `cube.PosFromVec3()` function you can use for this.

Sandertv (approved):

I'm merging this to feature/lightning where I'll try to get rid of the compiler directives and approach it some other way.

Sandertv:
Will close #134 as soon as it's merged into master.

PR #234, opened 2021-08-10
Allow a way to change the respawn world in HandleRespawn

JustTalDevelops:


server/player/handler.go
```diff
@@ -41,7 +41,7 @@ type Handler interface {
 	HandleDeath(src damage.Source)
 	// HandleRespawn handles the respawning of the player in the world. The spawn position passed may be
 	// changed by assigning to *pos.
-	HandleRespawn(pos *mgl64.Vec3)
+	HandleRespawn(w *world.World, pos *mgl64.Vec3)
```
Sandertv, line 44: I think you'll need a `**world.World` here, or it'll make a copy of your world when you change it.

Issue #218, opened 2021-07-24
Weather


Lucaskyy:
I'll try to see if I can work on this after I've finished #134 :)

PR #236, opened 2021-08-11
Weather Implementation

Lucaskyy:
This PR closes: #218 

The whole implementation is done and tested, rain works.
I believe thunder also works fine since it uses the same system anyway.

Just need to add lightning handling for the weather implementation, but this is going to be hard due to import cycling.
Hopefully, I don't have to use another compiler directive ._.

### DIFF
--- a/server/block/fire.go
+++ b/server/block/fire.go
@@ -1,12 +1,16 @@
 package block
 
+//lint:file-ignore ST1022 Exported variables in this package have compiler directives. These variables are not otherwise exposed to users.
+
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/entity"
 	"github.com/df-mc/dragonfly/server/entity/damage"
 	"github.com/df-mc/dragonfly/server/world"
+	"github.com/go-gl/mathgl/mgl64"
 	"math/rand"
 	"time"
+	_ "unsafe" // Imported for compiler directives.
 )
 
 // Fire is a non-solid block that can spread to nearby flammable blocks.
@@ -231,4 +235,27 @@ func allFire() (b []world.Block) {
 		b = append(b, Fire{Age: i, Type: SoulFire()})
 	}
 	return
+}
+
+//go:linkname block_setBlocksOnFire github.com/df-mc/dragonfly/server/entity.setBlocksOnFire
+//noinspection ALL
+var block_setBlocksOnFire func(w *world.World, lPos mgl64.Vec3)
+
+func init() {
+	block_setBlocksOnFire = func(w *world.World, lPos mgl64.Vec3) {
+		_, isNormal := w.Difficulty().(world.DifficultyNormal)
+		_, isHard := w.Difficulty().(world.DifficultyHard)
+		if isNormal || isHard { // difficulty >= 2
+			bPos := cube.PosFromVec3(lPos)
+			b := w.Block(bPos)
+			_, isAir := b.(Air)
+			_, isTallGrass := b.(TallGrass)
+			if isAir || isTallGrass {
+				below := w.Block(bPos.Side(cube.FaceDown))
+				if below.Model().FaceSolid(bPos, cube.FaceUp, w) || neighboursFlammable(bPos, w) {
+					w.PlaceBlock(bPos, Fire{})
+				}
+			}
+		}
+	}
 }
--- a/server/world/world.go
+++ b/server/world/world.go
@@ -655,7 +655,7 @@ func (w *World) StartTime() {
 	w.enableTimeCycle(true)
 }
 
-// enableTimeCycle enables or disables the time cycling of tthe World.
+// enableTimeCycle enables or disables the time cycling of the World.
 func (w *World) enableTimeCycle(v bool) {
 	if w == nil {
 		return
@@ -663,6 +663,78 @@ func (w *World) enableTimeCycle(v bool) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	w.set.TimeCycle = v
+}
+
+// ToggleWeatherCycle enables or disables weather of the World.
+func (w *World) ToggleWeatherCycle(v bool) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	w.set.WeatherCycle = v
+}
+
+// setRainTime sets the rain time.
+func (w *World) setRainTime(new int) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	w.set.RainTime = int32(new)
+	w.mu.Unlock()
+}
+
+// setRainLevel sets the rain level.
+func (w *World) setRainLevel(new int) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	w.set.RainLevel = float32(new)
+	w.mu.Unlock()
+}
+
+// setThunderTime sets the thunder time.
+func (w *World) setThunderTime(new int) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	w.set.ThunderTime = int32(new)
+	w.mu.Unlock()
+}
+
+// setThunderLevel sets the thunder level.
+func (w *World) setThunderLevel(new int) {
+	if w == nil {
+		return
+	}
+	w.mu.Lock()
+	w.set.ThunderLevel = float32(new)
+	w.mu.Unlock()
+}
+
+// IsRaining returns a bool that decides whether it is raining or not.
+func (w *World) IsRaining() bool {
+	if w == nil {
+		return false
+	}
+	w.mu.Lock()
+	a := w.set.RainLevel > 0
+	w.mu.Unlock()
+	return a
+}
+
+// IsThundering returns a bool that decides whether it is thundering or not.
+func (w *World) IsThundering() bool {
+	if w == nil {
+		return false
+	}
+	w.mu.Lock()
+	a := w.set.ThunderLevel > 0
+	w.mu.Unlock()
+	return a
 }
 
 // AddParticle spawns a particle at a given position in the world. Viewers that are viewing the chunk will be
@@ -1149,6 +1221,23 @@ func (w *World) tick() {
 		w.set.Time++
 	}
 	t := int(w.set.Time)
+
+	if w.set.WeatherCycle {
+		// Tick Weather
+		// NOTE: All of the following numbers/calculations you see are from default Minecraft.
+
+		// Raining
+		w.set.RainTime--
+		if w.set.RainTime <= 0 {
+			w.SetRaining(w.set.RainLevel <= 0)
+		}
+
+		// Thunder
+		w.set.ThunderTime--
+		if w.set.ThunderTime <= 0 {
+			w.SetThunder(w.set.ThunderLevel <= 0)
+		}
+	}
 	w.mu.Unlock()
 
 	if tick%20 == 0 {
@@ -1156,6 +1245,13 @@ func (w *World) tick() {
 			viewer.ViewTime(t)
 		}
 	}
+
+	//if w.IsThundering() {
+	//	for pos := range w.chunks {
+	//		// TODO Handle lightning
+	//		// w.AddEntity(entity.NewLightning())
+	//	}
+	//}
 
 	w.tickEntities(tick)
 	w.tickRandomBlocks(viewers, tick)
@@ -1432,6 +1528,52 @@ func (w *World) tickEntities(tick int64) {
 		ticker.Tick(tick)
 	}
 	w.entitiesToTick = w.entitiesToTick[:0]
+}
+
+// SetRaining toggles raining depending on the raining argument.
+func (w *World) SetRaining(raining bool) {
+	var level int
+	if raining {
+		level = 1
+	} else {
+		level = 0
+	} // Go developers, can you please add the ternary operator? :)
+	w.setRainLevel(level)
+
+	for _, v := range w.allViewers() {
+		v.ViewRain(raining)
+	}
+
+	if raining {
+		w.setRainTime(rand.Intn(12000) + 12000)
+	} else {
+		w.setRainTime(rand.Intn(168000) + 12000)
+	}
+}
+
+// SetThunder toggles raining depending on the thundering argument.
+func (w *World) SetThunder(thundering bool) {
+	if thundering && !w.IsRaining() {
+		w.SetRaining(true)
+	}
+
+	var level int
+	if thundering {
+		level = 1
+	} else {
+		level = 0
+	} // Go developers, can you please add the ternary operator? :)
+	w.setThunderLevel(level)
+
+	for _, v := range w.allViewers() {
+		v.ViewThunder(thundering)
+	}
+
+	if thundering {
+		w.setThunderTime(rand.Intn(12000) + 3600)
+	} else {
+		w.setThunderTime(rand.Intn(168000) + 12000)
+	}
 }
 
 // allViewers returns a list of all viewers of the world, regardless of where in the world they are viewing.
--- a/server/entity/damage/source.go
+++ b/server/entity/damage/source.go
@@ -48,6 +48,9 @@ type SourceLava struct{}
 // SourceFall is a source that is used if the player fell.
 type SourceFall struct{}
 
+// SourceLightning is used for damage caused by being struck by lightning.
+type SourceLightning struct{}
+
 // SourceCustom is a cause used for dealing any kind of custom damage. Armour reduces damage of this source,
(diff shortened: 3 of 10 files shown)


Lucaskyy:
Also, please note that because I am basing off on the feature/lightning branch, merging this to master will cause the lightning branch to also be merged. So watch out when doing that.

server/world/world.go
```diff
@@ -1157,6 +1246,13 @@ func (w *World) tick() {
 		}
 	}
 
+	//if w.IsThundering() {
+	//	for pos := range w.chunks {
+	//		// TODO Handle lightning
+	//		// w.AddEntity(entity.NewLightning())
+	//	}
+	//}
```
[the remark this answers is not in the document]
Lucaskyy: > Just need to add lightning handling for the weather implementation

Yes, it is.

server/world/world.go
```diff
@@ -1434,6 +1530,52 @@ func (w *World) tickEntities(tick int64) {
 	w.entitiesToTick = w.entitiesToTick[:0]
 }
 
+// SetRaining toggles raining depending on the raining argument.
+func (w *World) SetRaining(raining bool) {
+	var level int
+	if raining {
+		level = 1
+	} else {
```
JustTalDevelops, line 1538: This else statement is useless, since level will be 0 if it's not raining.

server/world/world.go
```diff
@@ -1434,6 +1530,52 @@ func (w *World) tickEntities(tick int64) {
...
+// SetThunder toggles raining depending on the thundering argument.
+func (w *World) SetThunder(thundering bool) {
+	if thundering && !w.IsRaining() {
+		w.SetRaining(true)
+	}
+
+	var level int
+	if thundering {
+		level = 1
+	} else {
```
JustTalDevelops, line 1563: Same here.

Lucaskyy:
Moved both the lightning fire block code to internal and the lightning striking code to internal.

Lucaskyy:
Fixed the formatting :)

Lucaskyy:
Any updates on this PR?

server/internal/entity_internal/lightning.go
```diff
@@ -0,0 +1,84 @@
...
+		_, flowingWater := blockType.(block.Water)
+		if !tallGrass && !flowingWater {
+			vec = vec.Add(mgl64.Vec3{0, 1, 0})
+		}
+
+		w.AddEntity(entity.NewLightning(vec))
+	}
+}
+
+func canBlockSeeSky(w *world.World, pos mgl64.Vec3) bool {
```
Sandertv, line 37: This isn't an accurate name as blocks that can see the sky through glass don't matter. This should probably be called `highestBlock`.

server/internal/entity_internal/lightning.go
```diff
@@ -0,0 +1,84 @@
...
+	"math/rand"
+)
+
+//go:linkname world_performThunder github.com/df-mc/dragonfly/server/world.performThunder
+//noinspection ALL
+var world_performThunder func(w *world.World, pos world.ChunkPos)
+
+func init() {
+	world_performThunder = func(w *world.World, pos world.ChunkPos) {
+		LCG := int32(w.GetUpdateLGC() >> 2)
```
Sandertv, line 20: Would make this all lowercase for consistency.

server/block/fire.go
```diff
@@ -30,8 +33,8 @@ func FlammableBlock(block world.Block) bool {
 	return false
 }
 
-// neighboursFlammable returns true if one a block adjacent to the passed position is flammable.
-func neighboursFlammable(pos cube.Pos, w *world.World) bool {
+// NeighboursFlammable returns true if one a block adjacent to the passed position is flammable.
+func NeighboursFlammable(pos cube.Pos, w *world.World) bool {
```
Sandertv, line 37: Any reason to export this?

server/block/fire.go
```diff
@@ -1,12 +1,15 @@
...
+//lint:file-ignore ST1022 Exported variables in this package have compiler directives. These variables are not otherwise exposed to users.
+
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/entity"
 	"github.com/df-mc/dragonfly/server/entity/damage"
 	"github.com/df-mc/dragonfly/server/world"
 	"math/rand"
 	"time"
+	_ "unsafe" // Imported for compiler directives.
```
Sandertv, line 12: No reason for this and the lint directive above I think.

server/world/world.go
```diff
@@ -210,6 +213,13 @@ func (w *World) HighestBlock(x, z int) int {
 	return int(v)
 }
 
+const LCGConstant = 1013904223
+
+func (w *World) GetUpdateLGC() int {
```
Sandertv, line 218: Can we not just get a random int everytime? I get this probably matches vanilla but to what extent does that matter?

server/world/world.go
```diff
@@ -665,6 +675,78 @@ func (w *World) enableTimeCycle(v bool) {
 	w.set.TimeCycle = v
 }
 
+// ToggleWeatherCycle enables or disables weather of the World.
+func (w *World) ToggleWeatherCycle(v bool) {
```
Sandertv, line 679: Would split this into a separate StartWeatherCycle and StopWeatherCycle. (Or whatever naming to be analogue with the time cycle)

server/world/world.go
```diff
@@ -1134,6 +1216,14 @@ func (w *World) startTicking() {
 	}
 }
 
+var (
+	shouldUpdateRain = false
+	updateRain       = false
+
+	shouldUpdateThunder = false
+	updateThunder       = false
+)
```
Sandertv, line 1225: This breaks with multiple worlds.

Sandertv (changes requested):
Thanks for the PR! I've got a couple of comments

Lucaskyy:
Thanks, Sander for the review. I am currently finalizing #239 and after that, I'll get back to work on this PR again and fix all the requested changes.

Sandertv:
Could you merge conflicts when you get the time? Thanks!

Lucaskyy:
Yep, will do it now. I forgot, hehe.

server/block/fire.go
Lucaskyy: Yes, it's needed in the lightning code.

server/world/world.go
Lucaskyy: We could do this, but I'm not entirely sure which calculations we should use otherwise, since I'm not that good at maths xP

Lucaskyy:
Everything should be fixed now, when you get the time could you please check? :)

Lucaskyy:
Alright, this PR is completely f'd. Some things just don't make sense, so I'm thinking about just reverting everything, basing off of master, and putting everything of this actual PR in place again.

server/world/world.go
Lucaskyy: This is all that's left to do then. I'll check it now.

server/world/world.go
```diff
@@ -1581,22 +1581,22 @@ func (w *World) setRaining(raining bool) {
 	}
 
 	if raining {
-		w.setRainTime(rand.Intn(12000) + 12000)
+		w.setRainTime(int(x.Seconds() / 20)) // rand.Intn(12000) + 12000
```
Sandertv, line 1584: I think this is supposed to be * 20. You could also do `x / (time.Second*20)` to avoid floats.

server/world/world.go
```diff
@@ -1610,7 +1610,7 @@ func (w *World) setThunder(thundering bool) {
 	}
 
 	if thundering {
-		w.setThunderTime(rand.Intn(12000) + 3600)
+		w.setThunderTime(int(x.Seconds() / 20)) // rand.Intn(12000) + 3600
```
Sandertv, line 1613: Same as above.

Sandertv: Also you can remove the comment here.

server/world/world.go
Sandertv: Can remove the comment here too.

server/world/world.go
```diff
@@ -1258,13 +1258,13 @@ func (w *World) tick() {
 		// Raining
 		w.set.RainTime--
 		if w.set.RainTime <= 0 {
-			w.setRaining(w.set.RainLevel <= 0)
+			w.setRaining(w.set.RainLevel <= 0, time.Duration(rand.Intn(600) + 600))
```
Sandertv: Don't think this is accurate. time.Duration is in nanoseconds.

server/world/world.go
```diff
@@ -1258,13 +1258,13 @@ func (w *World) tick() {
...
 		}
 
 		// Thunder
 		w.set.ThunderTime--
 		if w.set.ThunderTime <= 0 {
-			w.setThunder(w.set.ThunderLevel <= 0)
+			w.setThunder(w.set.ThunderLevel <= 0, time.Duration(rand.Intn(600) + 180))
```
Sandertv: Don't think this is accurate. time.Duration is in nanoseconds.

server/world/world.go
Lucaskyy: An alternative to this is just a random number between int max and min, since it most likely underflows the integer since the equations are `add` and it caused the int to go into the negative. I might be wrong though. So there's the question, what would be better? If that one integer really minds you, then yes we could do that but this is closest to vanilla MC.

Lucaskyy: Nevermind, figured it out.
LCG is like any other RNG, but it is known to be the best one out there. It doesn't have duplicates or whatever, so Mojang chose that. Go's random can obviously be used, but if that has duplicates then this is better.

Lucaskyy:
This PR is finished. :) You can merge it whenever you want.

server/internal/entity_internal/lightning.go
```diff
@@ -0,0 +1,61 @@
...
+	"math"
+	"math/rand"
+)
+
+//go:linkname world_performThunder github.com/df-mc/dragonfly/server/world.performThunder
+//noinspection ALL
+var world_performThunder func(w *world.World, pos world.ChunkPos)
+
+func init() {
+	world_performThunder = func(w *world.World, pos world.ChunkPos) {
```
Sandertv, line 19: Would pass the lcg rand instance to this function so you can get rid of w.ThunderLCG().

server/world/world.go
```diff
@@ -672,6 +676,88 @@ func (w *World) enableTimeCycle(v bool) {
...
+// setThunderLevel sets the thunder level.
+func (w *World) setThunderLevel(new int) {
+	if w == nil {
+		return
+	}
+	w.set.ThunderLevel = float32(new)
+}
+
+// IsRaining returns a bool that decides whether it is raining or not.
+func (w *World) IsRaining() bool {
```
Sandertv, line 732: This should probably be named Raining.

server/world/world.go
```diff
@@ -672,6 +676,88 @@ func (w *World) enableTimeCycle(v bool) {
...
+	if w == nil {
+		return false
+	}
+	w.mu.Lock()
+	a := w.set.RainLevel > 0
+	w.mu.Unlock()
+	return a
+}
+
+func (w *World) isRaining() bool {
```
Sandertv, line 742: Same as above.

server/world/world.go
```diff
@@ -672,6 +676,88 @@ func (w *World) enableTimeCycle(v bool) {
...
+	if w == nil {
+		return false
+	}
+	a := w.set.RainLevel > 0
+	return a
+}
+
+// IsThundering returns a bool that decides whether it is thundering or not.
+func (w *World) IsThundering() bool {
```
Sandertv, line 751: This should probably be called Thundering.

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
...
 }
 
+// SetRaining toggles raining depending on the raining argument.
+func (w *World) SetRaining(raining bool, x time.Duration) {
+	w.mu.Lock()
+	w.setRaining(raining, x)
+	w.mu.Unlock()
+}
+
+func (w *World) setRaining(raining bool, x time.Duration) {
```
Sandertv, line 1576: Might want to document this and indicate that, as opposed to SetRaining, setRaining does not lock.

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
...
+
+	if raining {
+		w.setRainTime(int(x.Seconds() * 20))
+	} else {
+		w.setRainTime(rand.Intn(168000) + 12000)
+	}
+}
+
+// SetThunder toggles raining depending on the thundering argument.
+func (w *World) SetThunder(thundering bool, x time.Duration) {
```
Sandertv, line 1595: For consistency, might want to call this `SetThundering`.

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
...
+	w.mu.Lock()
+	w.setThunder(thundering, x)
+	w.mu.Unlock()
+}
+
+func (w *World) setThunder(thundering bool, x time.Duration) {
```
Sandertv, line 1601: Same as above. Also document this like setRaining.

Sandertv (commented):
Got some last comments, mostly for consistency of naming. The compiler directives are a problem but I can deal with that after I merge it into a different branch.

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
...
+	if thundering && !w.isRaining() {
+		w.setRaining(true, time.Duration(rand.Intn(600)+600))
```
Sandertv, line 1603: This duration still seems wrong.

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
...
+	w.setThunderLevel(level)
+
+	for _, v := range w.allViewers() {
+		v.ViewThunder(thundering)
+	}
+
+	if thundering {
+		w.setThunderTime(int(x.Seconds() * 20))
+	} else {
+		w.setThunderTime(rand.Intn(168000) + 12000)
```
Sandertv, line 1619: This duration might be wrong?

server/world/world.go
```diff
@@ -1449,6 +1566,60 @@ func (w *World) tickEntities(tick int64) {
 	w.entitiesToTick = w.entitiesToTick[:0]
 }
 
+// SetRaining toggles raining depending on the raining argument.
+func (w *World) SetRaining(raining bool, x time.Duration) {
```
Sandertv, line 1570: Would maybe split this up into StartRaining(time.Duration) and StopRaining(). The duration argument isn't useful here. Same for thunder.

Sandertv (commented):
Got another couple of comments, these should be the last.

server/internal/entity_internal/lightning.go
Lucaskyy: How would that work if we have a rand instance per world? We can also just do a global one.

Sandertv: Just have the caller pass its rand instance to this function.

server/world/world.go
Lucaskyy: It is not. `setThunderTime` is in ticks, so this is valid.
The ones that require a `time.Duration` are in seconds, so these have lower values.

server/world/world.go
Lucaskyy: I see, I can agree with that. I will do that then.

Lucaskyy: Though, I think it is only best to do that for the exported methods since it would become a huge mess if we'd also do it for the unexported ones.

Sandertv: Oh yeah I wouldn't do it for the unexported ones.

Lucaskyy:
Fixed all requested changes, should be all done now, for real this time xP

server/world/world.go
```diff
@@ -672,6 +676,90 @@ func (w *World) enableTimeCycle(v bool) {
...
+// setThunderTime sets the thunder time.
+func (w *World) setThunderTime(new int) {
+	if w == nil {
+		return
+	}
+	w.set.ThunderTime = int32(new)
+}
+
+// setThunderLevel sets the thunder level.
+func (w *World) setThunderLevel(new int) {
```
Sandertv, line 724: Any reason this is an int while the ThunderLevel field is a float32? I have no idea what value this should hold?

Lucaskyy: If I recall correctly, the values used by vanilla is always an int so that's why I did that. And I don't think floats have a use for it, not sure though.

Sandertv: I would still probably either make it a bool or a float32. If it's not used as a float32 and always has either 0 or 1, it makes more sense to make it a bool.

PR #238, opened 2021-08-11
Fix Fall Damage/Distance

Lucaskyy:
onGround was never set to false, so this PR fixes it.


Lucaskyy:
Raptor is going to make a better fix for this, so I closed this PR.

Issue #212, opened 2021-07-20
Implement farmland trampling


Sandertv:
Farmland isn't currently turned into dirt when jumped on.

Sandertv:
Being worked on in #239. I imagine this will be finished fairly soon. Postponing to v0.4.0.

PR #239, opened 2021-08-11
Farmland Trampling

Lucaskyy:
This PR closes: #212


Lucaskyy:
Fixed the conflicts, you can merge this PR :)

Lucaskyy:
One thing though, the farmland successfully goes away but it does NOT give your items back. This should be fixed outside of farmland since there might be other things that also have the same behavior as farmland when getting destroyed.

server/block/block.go
```diff
@@ -216,13 +216,15 @@ func newFlammabilityInfo(encouragement, flammability int, lavaFlammable bool) Fl
 // EntityCollider is an interface for blocks with special behaviors on entity collision.
 type EntityCollider interface {
 	// EntityCollide is called on entity collision.
-	EntityCollide(e world.Entity)
+	EntityCollide(e world.Entity, pos cube.Pos)
```
T14Raptor, line 219: Maybe put `pos` before `e` for consistency reasons? Other block functions typically have `pos` as the first argument.

Lucaskyy: I agree. I'll fix that tomorrow.

Lucaskyy: Actually, I'll fix it now,

Lucaskyy:
This PR is done and mergeable if everything is alright.
## Preview:
https://imgur.com/a/zatqrTW

server/player/player.go
```diff
@@ -1728,7 +1733,7 @@ func (p *Player) checkBlockCollisions() {
 	w := p.World()
 
 	aabb := p.AABB().Translate(p.Position())
-	grown := aabb.Grow(0.25)
+	grown := aabb.GrowVec3(mgl64.Vec3{0, 0.25})
```
Sandertv, line 1736: Maybe this 0.25 needs to be changed. This is a rather big extension and I don't think we really need that. Maybe just 0.01 would even be enough.

Lucaskyy: I'll try to see if that works too.

Lucaskyy:
Note to self: use EntityLander

T14Raptor:
I quickly finished this up and is now ready to be merged, thanks for the PR!

PR #249, opened 2021-08-21
Fixed form close detection not working

Endermanbugzjfc:
Sorry, I did a small mistake. The code checks if a `form.Form` struct is a Closer, instead of the submittable of it.

### DIFF
--- a/server/player/form/form.go
+++ b/server/player/form/form.go
@@ -15,7 +15,6 @@ import (
 type Form interface {
 	json.Marshaler
 	SubmitJSON(b []byte, submitter Submitter) error
-	HandleFormClose(submitter Submitter)
 	__()
 }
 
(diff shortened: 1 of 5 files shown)


server/player/form/form.go
```diff
@@ -78,6 +78,13 @@ func (f Custom) Elements() []Element {
 // If the values are valid and can be parsed properly, the Submit() method of the form's Submittable is called
 // and the fields of the Submittable will be filled out.
 func (f Custom) SubmitJSON(b []byte, submitter Submitter) error {
+	if b == nil {
```
Sandertv, line 81: This needs to be in all 3 of the forms.

Sandertv (approved):

Sandertv:
Looks good now, thanks!

PR #260, opened 2021-08-26
Adds the option to enable vanilla instant respawn

Prim69:
Vanilla's instant respawn is client-side, it just removes the respawn screen which is why this works.


Sandertv:
Could you add the proper punctuation in all documentation? Full stops most notably.

Sandertv:
Looks good, thank you!

PR #264, opened 2021-09-02
Fixed animation of block breaking in creative mode

arfush:
I found a bug with animation of block breaking in creative mode when a player extinguishes fire.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -1279,6 +1279,10 @@ func (p *Player) StartBreaking(pos cube.Pos, face cube.Face) {
 
 		p.breaking.Store(true)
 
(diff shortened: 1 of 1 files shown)


server/player/player.go
```diff
@@ -1279,6 +1279,10 @@ func (p *Player) StartBreaking(pos cube.Pos, face cube.Face) {
 
 		p.breaking.Store(true)
 
+		if p.GameMode().CreativeInventory() {
+			return
+		}
```
Sandertv, line 1284: The positioning of this just looks incorrect. Is there any reason this return happens after the `p.breaking.Store(true)`? And how about the `p.SwingArm()`? Why does that not happen?

PR #265, opened 2021-09-02
Implement *Player.Cps()

RestartFU:
in this PR I want to implement an easy way to get the current Clicks per second for a player. I know this is probably not well done but I still want to pr in case anyone have any idea of how to make it mergable

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -89,6 +89,7 @@ type Player struct {
 	breakParticleCounter atomic.Uint32
 
 	hunger *hungerManager
+	cps    int
 }
 
 // New returns a new initialised player. A random UUID is generated for the player, so that it may be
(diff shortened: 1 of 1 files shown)


server/player/player.go
```diff
@@ -1939,6 +1941,20 @@ func (p *Player) Breathing() bool {
...
+
+//addCps updates the Cps of the player and removes it 1 second later
+func (p *Player) addCps() {
+	go func() {
+		p.cps += 1
+		time.Sleep(1 * time.Second)
+		p.cps -= 1
+	}()
+}
+
```
TwistedAsylumMC, line 1957: Please use proper punctuation and formatting. Running ``go fmt`` over your code should fix the formatting issues. Acronyms should be all capitals.

Issue #255, opened 2021-08-25
Enable resource pack loading in Dragonfly


Sandertv:
This should be a fairly trivial task: Some configuration in the config.toml, loading resource packs probably from some designated new folder. Gophertunnel already has the API necessary for loading and sending resource packs.

SALLYPEMDAS:
PR open! let me know if there's anything to fix!!!

PR #274, opened 2021-09-13
Enables resource pack loading in Dragonfly

SALLYPEMDAS:
https://github.com/df-mc/dragonfly/issues/255

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -45,3 +45,7 @@
   # Folder controls where the player data will be stored by the default LevelDB
   # player provider if it is enabled.
   Folder = "players"
+
+[Resources]
+  # Folder configures the directory used by the server to load resource packs.
+  Folder = "resources"
--- a/server/config.go
+++ b/server/config.go
@@ -49,6 +49,11 @@ type Config struct {
 		// player provider if it is enabled.
 		Folder string
 	}
+
+	Resources struct {
+		// Folder controls the location where resource packs will be loaded from.
+		Folder string
+	}
 }
 
 // DefaultConfig returns a configuration with the default values filled out.
@@ -66,5 +71,6 @@ func DefaultConfig() Config {
 	c.Players.MaximumChunkRadius = 32
 	c.Players.SaveData = true
 	c.Players.Folder = "players"
+	c.Resources.Folder = "resources"
 	return c
 }
--- a/server/server.go
+++ b/server/server.go
@@ -5,6 +5,17 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"math/rand"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"syscall"
+	"time"
+	_ "unsafe" // Imported for compiler directives.
+
 	_ "github.com/df-mc/dragonfly/server/block"
 	"github.com/df-mc/dragonfly/server/cmd"
 	"github.com/df-mc/dragonfly/server/internal"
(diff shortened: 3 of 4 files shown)


server/listener.go
```diff
@@ -1,9 +1,11 @@
 package server
 
 import (
+	"io"
+
 	"github.com/df-mc/dragonfly/server/session"
 	"github.com/sandertv/gophertunnel/minecraft"
-	"io"
+	"github.com/sandertv/gophertunnel/minecraft/resource"
```
SALLYPEMDAS, line 8: reverting these changes

server/server.go
```diff
@@ -94,6 +98,28 @@ func New(c *Config, log internal.Logger) *Server {
...
+	if _, err := os.Stat(c.Resources.Folder); os.IsNotExist(err) {
+		_ = os.Mkdir(c.Resources.Folder, 0777)
+	}
+	resources, err := os.ReadDir(c.Resources.Folder)
+	if err != nil {
+		panic(err)
+	}
+	for _, entry := range resources {
+		fmt.Println(" ", entry.Name(), entry.IsDir())
+		fmt.Println(entry.Name())
```
DaPigGuy, line 110:
```suggestion
```

SALLYPEMDAS: will fix

server/server.go
```diff
@@ -94,6 +98,26 @@ func New(c *Config, log internal.Logger) *Server {
...
 
+	if _, err := os.Stat(c.Resources.Folder); os.IsNotExist(err) {
+		_ = os.Mkdir(c.Resources.Folder, 0777)
+	}
+	resources, err := os.ReadDir(c.Resources.Folder)
+	if err != nil {
+		panic(err)
+	}
+	for _, entry := range resources {
+		if entry.IsDir() {
```
TwistedAsylumMC, line 109: Why ignore directories here? Gophertunnel's resource pack implementation can parse directories too.

SALLYPEMDAS: Thanks for the review! Will fix!

Sandertv (changes requested):
Thanks for the PR!

Currently this doesn't have an entry added in the config.toml. Could you add it there with the appropriate documentation? Your PR looks really good otherwise, you could also consider moving the resource pack loading in server.New() to a separate function.

SALLYPEMDAS:
Thanks for the reviews!!!

PR updated:

Let me know if there are any more changes :)

server/server.go
```diff
@@ -547,3 +553,23 @@ func (server *Server) itemEntries() (entries []protocol.ItemEntry) {
 	}
 	return
 }
+
+// loadResources loads resource packs from path of specifed directory
```
Sandertv, line 557: `.` at the end here please!

server/config.go
```diff
@@ -49,6 +49,11 @@ type Config struct {
 		// player provider if it is enabled.
 		Folder string
 	}
+
+	Resources struct {
+		// Folder controls the location where resource packs will be loaded from
```
Sandertv, line 54: `.` at the end here please!

config.toml
```diff
@@ -45,3 +45,7 @@
   # Folder controls where the player data will be stored by the default LevelDB
   # player provider if it is enabled.
   Folder = "players"
+
+[Resources]
+  # Folder configures the directory used by the server to load resource packs
```
Sandertv, line 50: `.` at the end here please!

Sandertv (commented):
Last couple of documentation nitpicks, ready to go after this!

config.toml
SALLYPEMDAS: certainly!

server/config.go
SALLYPEMDAS: Yes!

server/server.go
SALLYPEMDAS: 👍

Sandertv (approved):
Looks good, thanks again for your PR!

SALLYPEMDAS:
<3

PR #286, opened 2021-09-25
Ignore embed fields in form submittable structs

Endermanbugzjfc:


Sandertv:
This has a problem too as it means the fields of the embedded struct, when they are actually exported too, cannot be used. In go 1.17 there's the reflect.VisibleFields() function that also returns all exported fields of the embedded structs, but we're not yet on that version.

PR #291, opened 2021-09-30
Implement PlaySound/StopSound

tigerza117:


server/player/player.go
```diff
@@ -367,6 +367,18 @@ func (p *Player) Transfer(address string) (err error) {
...
 }
 
+// SendSound send play a sound to the player. If sound to have in vanilla or custom resource pack will be not work
+func (p *Player) SendSound(soundName string, volume, pitch float32) (err error) {
+	p.session().SendSound(soundName, p.Position(), volume, pitch)
+	return
+}
+
+// StopSound send stop play a sound by sound name while sound playing to player
+func (p *Player) StopSound(soundName string, stopAll bool) (err error) {
```
Lucaskyy, line 377: I think it would be better to have a `StopSound(soundName string)` and a `StopAllSounds()` since the `soundName` parameter seems useless when `stopAll` is true.

Lucaskyy (changes requested):
Everything seems to be okay, but I have 1 little thing that might need changes.

Sandertv:
There already is a PlaySound method in players. I think it's more appropriate to merge this with that, and just expose sounds based on strings as a Sound implementation.

server/player/player.go
tigerza117: OK, I refactor to be a StopSound(soundName string) and a StopAllSounds()

server/player/player.go
```diff
@@ -1911,8 +1911,34 @@ func (p *Player) EyeHeight() float64 {
...
 // PlaySound plays a world.Sound that only this Player can hear. Unlike World.PlaySound, it is not broadcast
 // to players around it.
-func (p *Player) PlaySound(sound world.Sound) {
-	p.session().ViewSound(entity.EyePosition(p), sound)
+// if passing world.Sound argument it should be use LevelSoundEvent, if passing string argument it should be use PlaySound
+func (p *Player) PlaySound(sound ...interface{}) {
+	switch so := sound[0].(type) {
+	case world.Sound:
+		p.session().ViewSound(entity.EyePosition(p), so)
+	case string:
```
Sandertv, line 1919: Just implement world.Sound with a new type. I wouldn't add a special case for this.

tigerza117: Like world.PlaySound ?

Sandertv: Just implement it like any other sound is currently implemented. Make it a struct in the world/sound package and have fields in it like the name of the particle, pitch etc. Then make session send it appropriately in the `ViewSound` method.

tigerza117: OK

tigerza117:
Example....

`p.PlaySound(sound.PlaySound{SoundName: "random.toast", Volume: 1, Pitch: 1})`

PR #295, opened 2021-10-02
Implement red and brown mushrooms

JustTalDevelops:


server/block/brown_mushroom.go
```diff
@@ -0,0 +1,67 @@
...
+	return 1
+}
+
+// BreakInfo ...
+func (b BrownMushroom) BreakInfo() BreakInfo {
+	return BreakInfo{
+		Harvestable: alwaysHarvestable,
+		Effective:   nothingEffective,
+		Drops:       simpleDrops(item.NewStack(b, 1)),
+	}
```
DaPigGuy, line 28: Use the newBreakInfo helper method

server/block/red_mushroom.go
```diff
@@ -0,0 +1,62 @@
...
+	transparent
+}
+
+// BreakInfo ...
+func (r RedMushroom) BreakInfo() BreakInfo {
+	return BreakInfo{
+		Harvestable: alwaysHarvestable,
+		Effective:   nothingEffective,
+		Drops:       simpleDrops(item.NewStack(r, 1)),
+	}
```
DaPigGuy, line 23: Same here

Sandertv:
I'd merge the two mushrooms and have a MushroomType type to specify red/brown.

JustTalDevelops:
> 
> 
> I'd merge the two mushrooms and have a MushroomType type to specify red/brown.

The thing is, brown mushrooms are light emitters while red mushrooms aren't. By adding the LightEmissionLevel function to the Mushroom block, I'd make RedMushrooms a light block too. Is that an issue, or can I just set the light emission level for red mushrooms to 0?

JustTalDevelops:
> 
> 
> I'd merge the two mushrooms and have a MushroomType type to specify red/brown.

This should be done. Let me know if anything else should be changed.

server/block/leaves.go
```diff
@@ -13,10 +13,10 @@ import (
 type Leaves struct {
 	leaves
 
-	// Wood is the type of wood of the leaves. This field must have one of the values found in the material
+	// Wood is the type of mushroom of the leaves. This field must have one of the values found in the material
```
Sandertv, line 16: Your refactor of wood -> mushroom went a bit crazy. Could you fix these?

server/block/mushroom_type.go
```diff
@@ -0,0 +1,28 @@
...
+
+// MushroomTypes returns a list of all mushroom types.
+func MushroomTypes() []MushroomType {
+	return []MushroomType{BrownMushroom(), RedMushroom()}
+}
+
+type mushroom uint8
+
+// Uint8 returns the mushroom as an uint8.
+func (m mushroom) Uint8() uint8 {
```
Sandertv, line 26: Could maybe replace this with a `String` method instead and return "red" or "brown" so you can use it in EncodeBlock and EncodeItem.

server/block/mushroom.go
```diff
@@ -0,0 +1,81 @@
...
+	if !used {
+		return false
+	}
+
+	blockBelow := w.Block(pos.Side(cube.FaceDown))
+	if diffuser, ok := blockBelow.(LightDiffuser); ok {
+		if diffuser.LightDiffusionLevel() == 0 {
+			return false
+		}
+	}
```
Sandertv, line 45: If I understand this logic right, this would allow a mushroom to be placed on, say, a torch. I don't think that's right.

PR #297, opened 2021-10-02
Implement bone meal functionality on grass

JustTalDevelops:


server/block/grass.go
```diff
@@ -78,3 +92,30 @@ func (g Grass) Till() (world.Block, bool) {
 func (g Grass) Shovel() (world.Block, bool) {
 	return DirtPath{}, true
 }
+
+// plantSelection returns the selection of plants that can be grown by bone meal.
+func plantSelection() []world.Block {
```
Sandertv, line 97: You could simply store this in a package level variable so you don't need to recreate this slice every time bone meal is used on grass.

Sandertv (commented):
When I first looked at this PR I wanted to ask you to add a TODO for biomes, but you did that already so I think this looks good. Got one small improvement.

JustTalDevelops:
This is ready for review, by the way.

PR #298, opened 2021-10-02
Add a check, if 's' is blank, remove the line instead of replacing it in scoreboard…

RestartFU:
I'm not sure if we should really add that check in case you actually want to replace a line by nothing. In any case, I still think that we should have an easy to understand way to remove a line from the scoreboard

### DIFF
--- a/server/player/scoreboard/scoreboard.go
+++ b/server/player/scoreboard/scoreboard.go
@@ -52,8 +52,13 @@ func (board *Scoreboard) Set(index int, s string) (err error) {
 		return fmt.Errorf("index out of range %v", index)
 	}
 
-	// Remove new lines from the string
-	board.lines[index] = strings.TrimSuffix(strings.TrimSuffix(s, "\n"), "\n")
+	if s == "" {
+		// If s is nothing, remove the line
+		board.lines = append(board.lines[:index], board.lines[index+1:]...)
+	} else {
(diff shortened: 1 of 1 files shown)


Thunder33345:
-1 this feels "magical" and functionally overloading set to do more then setting
i am in favor of having something more straightforward like a remove function that takes in the int as the line to remove

Sandertv:
I'd just introduce a Remove method.

RestartFU:
I just made a new commit to my PR. I added a Remove method.

server/player/scoreboard/scoreboard.go
```diff
@@ -51,9 +51,19 @@ func (board *Scoreboard) Set(index int, s string) (err error) {
...
 		return fmt.Errorf("index out of range %v", index)
 	}
-
 	// Remove new lines from the string
 	board.lines[index] = strings.TrimSuffix(strings.TrimSuffix(s, "\n"), "\n")
+
+	return nil
+}
+
+// Remove removes a specific line from the scoreboard
```
Sandertv, line 60: If you add a `.` at the end here this PR is ready to be merged.

RestartFU: done

Sandertv (approved):

Sandertv:
Looks good, thank you!

PR #299, opened 2021-10-02
Optional console implementation

RestartFU:
This will implement console which can be enabled or disabled by setting the Console setting to true or false in the server config


Sandertv:
As discussed on Discord, we feel this is better left in a separate library. That said though, please do move this to a library and add the #df-mc tag so people can find it.

Thanks for your PR!

PR #301, opened 2021-10-02
Implemented sea pickle block

DaPigGuy:


server/block/sea_pickle.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+// SeaPickle is a small stationary underwater block that emits light, and is typically found in colonies of up to
+// four sea pickles.
+type SeaPickle struct {
+	empty
+	transparent
+
+	// ClusterCount is the amount of additional sea pickles clustered together.
+	ClusterCount int
```
Sandertv, line 19: Would maybe just name this `Count`.

server/block/sea_pickle.go
```diff
@@ -0,0 +1,165 @@
...
+	// Alive is whether the sea pickle is alive.
```
Sandertv, line 20: When is it alive? What does it mean practically?

Sandertv: (don't tell me, I know, but documenting it might be a good idea :P)

server/block/sea_pickle.go
Sandertv: That said, it seems this is 0 when there's 1 pickle. We should either document that very well or change the name of this field in a way that it indicates additional pickles.

server/block/sea_pickle.go
Sandertv: It might also be more consistent to make this field `Dead` as that's the convention with coral too.

Sandertv (approved):

PR #304, opened 2021-10-02
Implement HandleOpenContainer

RestartFU:
Ok I'm not sure that's the best name so feel free to suggest better names but I still think that this handler can be useful for many things

### DIFF
--- a/server/player/handler.go
+++ b/server/player/handler.go
@@ -1,6 +1,8 @@
 package player
 
 import (
+	"net"
+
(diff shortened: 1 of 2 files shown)


Sandertv:
Any place for this where simply handling interaction does not suffice? Also see #210.

RestartFU:
We wouldn't have to check if the player interacted with a container. Also if you change your mind merge #210 and not mine

Lucaskyy:
It seems that #210 is stale, so I think it's best to close that PR and continue here.

PR #305, opened 2021-10-03
adds a HandleToggleSprint in the player handler

Prim69:
This can be used to cancel sprinting or other cases for minigames.

### DIFF
--- a/server/player/handler.go
+++ b/server/player/handler.go
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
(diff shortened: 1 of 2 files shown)


server/player/handler.go
```diff
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
+	// HandleToggleSprint handles when the player starts or stops sprinting
+	HandleToggleSprint(ctx *event.Context, toggle bool)
```
Sandertv, line 26: The name 'toggle' doesn't make it clear what value this holds. To make it extra clear you could have a `before, after bool`

Sandertv:
Cancelling these events doesn't actually cancel the sprinting right now, because the Compare And Swap is done before the event call. That doesn't seem right?

PR #306, opened 2021-10-03
Adds a HandleToggleSneak function to the player handler

Prim69:
This would probably mainly be used for minigames

ex: in Hive's sonic mode you held down sneak to charge up a booster

### DIFF
--- a/server/player/handler.go
+++ b/server/player/handler.go
@@ -24,6 +24,8 @@ type Handler interface {
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
 	// HandleToggleSprint handles when the player starts or stops sprinting
 	HandleToggleSprint(ctx *event.Context, toggle bool)
+	// HandleToggleSneak handles when the player starts or stops sneaking
+	HandleToggleSneak(ctx *event.Context, toggle bool)
 	// HandleChat handles a message sent in the chat by a player. ctx.Cancel() may be called to cancel the
 	// message being sent in chat.
 	// The message may be changed by assigning to *message.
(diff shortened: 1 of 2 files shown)


server/player/handler.go
```diff
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
+	// HandleToggleSneak handles when the player starts or stops sneaking
+	HandleToggleSneak(ctx *event.Context, toggle bool)
```
Sandertv, line 26: Same with this 'toggle' field as in the HandleToggleSprint PR. This is quite an ambiguous variable name.

Lucaskyy: I think something like `toggled` would be better.

server/player/handler.go
```diff
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
+	// HandleToggleSneak handles when the player starts or stops sneaking
```
Lucaskyy, line 25: Please fix the punctuations here by putting a dot (.) at the end.

Lucaskyy (changes requested):
LGTM besides the 2 suggested changes mentioned.

server/player/handler.go
Sandertv: I think that's still ambiguous. Does it mean the old or the new value?

Lucaskyy: The new value. Is something like `isToggled` or `currentlyToggled` better? I'm not entirely sure what would be good to tell it's a new value without making it paragraphs long.

Sandertv: See the toggle sprint PR, I think that's the most straightforward way to do it.

Sandertv: Or maybe a `sneakingAfter` field.

Lucaskyy: `old, new` is better than `before, after` I think.

Lucaskyy (approved):
LGTM!

Sandertv:
Cancelling these events doesn't actually cancel the sneaking right now, because the Compare And Swap is done before the event call. That doesn't seem right?

PR #307, opened 2021-10-03
add command.Allow() function

AndreasHGK:
As mentioned in the documentation of the function, this function essentially checks if the source can use at least one of the Runnables of the command.
An important use for this is for example checking whether or not a command should be included in a /help command (if you implement a custom version of this)
Another use could be for example if you have a commandSpy feature and you want to exclude commands that the player does not have access to.

Currently there is already a way to check this that does essentially the same by checking if `len(command.Params(source)) == 0` but I do not think that is very intuitive in my opinion.

### DIFF
--- a/server/cmd/command.go
+++ b/server/cmd/command.go
@@ -106,6 +106,17 @@ func (cmd Command) Aliases() []string {
 	return cmd.aliases
 }
 
+// Allow returns whether the provided source is allowed to run the command. It will check each Runnable, and
+// will return true if one or more Runnables allow the source to run the command.
+func (cmd Command) Allow(source Source) bool {
+	for _, runnable := range cmd.v {
+		if allower, ok := runnable.Interface().(Allower); !ok || allower.Allow(source) {
+			return true
+		}
+	}
+	return false
+}
+
 // Execute executes the Command as a source with the args passed. The args are parsed assuming they do not
 // start with the command name. Execute will attempt to parse and execute one Runnable at a time. If one of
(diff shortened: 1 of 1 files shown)


Sandertv:
While I see the use for such a thing, I'm not sure how useful this is in practice when different runnables in a command can have different permissions. In that case you'd want to iterate over all runnables anyway, and this doesn't really assist with that?

AndreasHGK:
There currently is not a method to iterate over rumnables outside of the command package as far as i have seen. If the user has access to at least one runnable, then they can at least run the command right?

Sandertv:
Not necessarily, that just means it's able to run one of the usages of a command... which isn't very useful to know.

AndreasHGK:
It tells whether the user is able to see the command, which is what I'd use it for in for example /help
But yeah, just using `cmd.Params(source)` for this would work fine too (altho maybe you can't be certain if this behaviour will always be the same in the future?)

Sandertv:
The problem is that in /help you'd always display all runnables of a command separately. That's why I'm not sure how useful this specific method would be. I do think we should have something more user-friendly than the current cmd.Params() though.

AndreasHGK:
the thing is in my /help i only show command names and description as  i personally think this is more readable thanhaving every single usage listed
you already see those when you type the command (with autofill) and you can still get the usages when doing /help <commandName>
but i do agree that this has a pretty nice use case. I think command.Visible(source) would describe this function better

PR #321, opened 2021-10-20
cmd/argument.go: Added back support for single Target parameters.

T14Raptor:


T14Raptor:
Now that I think about it, this is kind of useless

PR #329, opened 2021-11-07
Send available entities to player on session start

NeutronicMC:


server/session/session.go
```diff
@@ -406,3 +415,19 @@ func (s *Session) closePlayerList() {
...
 	sessionMu.Unlock()
 }
+
+// sendAvailableEntities sends all registered entities to the player.
+func (s *Session) sendAvailableEntities() {
+	entities := world.Entities()
+	var entityData []actorIdentifier
+	for _, entity := range entities {
+		id := entity.EncodeEntity()
+		s.log.Debugf("entity %v\n", id)
```
TwistedAsylumMC, line 425: Guessing this isn't supposed to be here

NeutronicMC: Yep, lemme fix

server/world/entity.go
```diff
@@ -67,3 +67,8 @@ func EntityByName(name string) (SaveableEntity, bool) {
 	e, ok := entities[name]
 	return e, ok
 }
+
+// Entities returns all registered entities.
+func Entities() map[string]SaveableEntity {
```
Sandertv, line 72: I'd make this grab the values from the map and return a slice. We don't want people to be able to modify this map directly.

server/session/session.go
```diff
@@ -104,6 +105,12 @@ type Conn interface {
 	StartGame(data minecraft.GameData) error
 }
 
+// actorIdentifier represents the structure of an actor identifier sent over the network.
+type actorIdentifier struct {
```
Sandertv, line 109: I'd also move this down to where it's used. You could even have it inside of the sendAvailableEntities method.

Sandertv (approved):
Looks good and has been tested.

Sandertv:
Thanks for the PR!

PR #330, opened 2021-11-10
Implement entity riding

kaedendev:
Entities with the ability to ride other entities implement the Rider interface

Entities with the ability to be ridden implement the Rideable interface
- Entities have a configurable number of seats, and the seats have individually configurable positions
- When a Rider dismounts, all other current riders are reseated and a new driver is selected based on the order that they mounted the entity

Player implements the Rider interface

I've tested all of these changes


Sandertv (commented):
Thanks for your PR! I've got a couple of comments. It might also be desirable to have a mount/dismount event handler in player/handler.go.

server/entity/rider.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// Rider is an interface for entities that can ride other entities.
+type Rider interface {
+	// SeatPosition returns the Rider's current seat position.
+	SeatPosition() mgl32.Vec3
+	// RideEntity links the Rider to an entity if the entity is Rideable and if there is a seat available.
+	RideEntity(e world.Entity)
+	// DismountEntity unlinks the Rider from an entity.
+	DismountEntity(e world.Entity)
```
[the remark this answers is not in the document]
NeutronicMC: I was thinking that any Rider would have a function to dismount, but if it isn't necessary I'll remove it.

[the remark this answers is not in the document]
Sandertv: Yeah that's fair enough. You can keep it too if you think that's better.

server/player/player.go
```diff
@@ -2050,6 +2057,110 @@ func (p *Player) PunchAir() {
 	})
 }
 
+// MountEntity mounts the player to an entity if the entity is rideable and if there is a seat available.
+func (p *Player) MountEntity(r entity.Rideable) {
+	if e, ok := r.(world.Entity); ok {
```
Sandertv, line 2062: Would add a (cancellable) event for this.

server/player/player.go
```diff
@@ -2050,6 +2057,110 @@ func (p *Player) PunchAir() {
...
+		} else {
+			// Check and update seat position
+			p.checkSeats(e)
+		}
+	}
+}
+
+// DismountEntity dismounts the player from an entity.
+func (p *Player) DismountEntity() {
+	e, _ := p.RidingEntity()
```
Sandertv, line 2089: Would add a cancellable event for this.

Sandertv (commented):
Looking pretty good! My last comments are to add mount/dismount handlers in the player.Handler that are cancellable. Should be good to go after that if it's all been tested.

server/player/player.go
```diff
@@ -2050,6 +2057,127 @@ func (p *Player) PunchAir() {
 	})
 }
 
+// MountEntity mounts the player to an entity if the entity is rideable and if there is a seat available.
+func (p *Player) MountEntity(r entity.Rideable) {
+	ctx := event.C()
+	p.handler().HandleMount(ctx, r)
+	ctx.Continue(func() {
+		if e, ok := r.(world.Entity); ok {
```
Sandertv, line 2065: Would check for this at the start of this method. Also maybe make `entity.Rideable` implement `world.Entity` to allow omission of this check entirely?

server/player/player.go
```diff
@@ -2050,6 +2057,115 @@ func (p *Player) PunchAir() {
...
+			riders := r.Riders()
+			seat := len(riders)
+			positions := r.SeatPositions()
+			if len(positions) >= seat {
+				p.seatPosition.Store(positions[seat-1])
+				p.updateState()
+				for _, v := range p.viewers() {
+					v.ViewEntityMount(p, r, seat-1 == 0)
+				}
+			}
```
Sandertv, line 2077: Can return after this so you can remove the `} else {` on the next line.

server/player/player.go
```diff
@@ -2050,6 +2057,115 @@ func (p *Player) PunchAir() {
...
+		}
+	})
+}
+
+// DismountEntity dismounts the player from an entity.
+func (p *Player) DismountEntity() {
+	ctx := event.C()
+	e, seat := p.RidingEntity()
+	if e != nil {
+		if rideable, ok := e.(entity.Rideable); ok {
```
Sandertv, line 2090: This is always an `entity.Rideable`, no need to check if this is `ok`.

server/player/player.go
```diff
@@ -2050,6 +2057,115 @@ func (p *Player) PunchAir() {
...
+				for _, r := range rideable.Riders() {
+					r.MountEntity(rideable)
+				}
+			})
+		}
+	}
+}
+
+// checkSeats moves a player to the seat corresponding to their current index within the slice of riders.
+func (p *Player) checkSeats(e world.Entity) {
```
Sandertv, line 2110: Would just accept an `entity.Rideable` here.

server/player/player.go
```diff
@@ -2050,6 +2057,115 @@ func (p *Player) PunchAir() {
...
+	}
+}
+
+// SeatPosition returns the position of the player's seat.
+func (p *Player) SeatPosition() mgl32.Vec3 {
+	return p.seatPosition.Load().(mgl32.Vec3)
+}
+
+// seat returns the index of a player within the slice of riders.
+func (p *Player) seat(e world.Entity) int {
```
Sandertv, line 2134: Would accept an entity.Rideable here.

server/player/player.go
```diff
@@ -70,6 +71,10 @@ type Player struct {
 	armour       *inventory.Armour
 	heldSlot     *atomic.Uint32
 
+	seatPosition atomic.Value
+	ridingMu     sync.Mutex
+	riding       world.Entity
```
Sandertv, line 76: Would store this as a RideableEntity.

server/player/player.go
```diff
@@ -2050,6 +2057,115 @@ func (p *Player) PunchAir() {
...
+
+// setRiding saves the entity the Rider is currently riding.
+func (p *Player) setRiding(e world.Entity) {
+	p.ridingMu.Lock()
+	p.riding = e
+	p.ridingMu.Unlock()
+}
+
+// RidingEntity returns the entity the player is currently riding and the player's seat index.
+func (p *Player) RidingEntity() (world.Entity, int) {
```
Sandertv, line 2154: Would return an entity.Rideable here.

server/session/controllable.go
```diff
@@ -58,6 +60,11 @@ type Controllable interface {
 	Flying() bool
 	StopFlying()
 
+	SeatPosition() mgl32.Vec3
+	MountEntity(e entity.Rideable)
+	DismountEntity()
+	RidingEntity() (world.Entity, int)
```
Sandertv, line 66: Would return an entity.Rideable here.

server/entity/rider.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// Rider is an interface for entities that can ride other entities.
+type Rider interface {
+	// SeatPosition returns the Rider's current seat position.
+	SeatPosition() mgl32.Vec3
+	// MountEntity mounts the Rider to an entity if the entity is Rideable and if there is a seat available.
+	MountEntity(e Rideable)
+	// RidingEntity returns the entity the player is currently riding and the player's seat index.
+	RidingEntity() (world.Entity, int)
```
Sandertv, line 15: Would return a Rideable here.

Sandertv (commented):
Got some last comments just to clean up some of the world.Entity vs entity.Rideable. Should be good to go once you resolve those.

Sandertv:
So most of this PR looks good now, I just have one concern, which is `Rideable.RemoveRider` vs `Player.DismountEntity`. We expose two functions that (should) do the same, but one of them (`Rideable.RemoveRider`) is incomplete in the sense that it doesn't actually properly dismount the entity that is riding it. The same problem exists with `Rideable.AddRider` and `Rider.MountEntity`. I'm not personally sure how to fix this, do you have any ideas?

NeutronicMC:
I'm not sure how I would change that other than making the method names and descriptions more specific. The player needs to contain some sort of function to send packets to link entities as well as store the entity that it is mounted to. The entity needs to store all of the currently mounted players. I can't move most of the functionality into either method because it's specific to the scope of each interface. If you have any better ideas I'd be willing to try them out, but as far as I can tell this implementation works the best.

JustTalDevelops:
Any update on this?

JustTalDevelops:
As Sander suggested on Discord, a good alternative is probably some sort of `entity.Mount(some Entity, on Entity)` function.

PR #331, opened 2021-11-14
Implement crafting (2x2 and 3x3 systems)

JustTalDevelops:
This includes both regular and automatic crafting support.


server/item/recipe/item.go
```diff
@@ -0,0 +1,94 @@
...
+// InputItem is a recipe item. It is an item stack inherently, but it also has an extra value for if it applies to all types.
+type InputItem struct {
+	item.Stack
+	// Variants is true if the item applies to all of its possible variants. This is so it can be rendered in the recipe book
+	// for alternative types.
+	Variants bool
+}
+
+// inputItem is an item that is inputted to a crafting menu.
+type inputItem struct {
```
Sandertv, line 17: Not a huge fan of how this and InputItem are named the same. Maybe suffix it with `Data`?

server/item/recipe/item.go
```diff
@@ -0,0 +1,94 @@
...
+	// Name is the name of the item being inputted.
+	Name string `nbt:"name"`
+	// MetadataValue is the meta of the item. This can change the item almost completely, or act as durability.
+	MetadataValue int32 `nbt:"meta"`
+	// Count is the amount of the item.
+	Count int32 `nbt:"count"`
+}
+
+// toInputItem converts an inputItem to an InputItem.
+func (i inputItem) toInputItem() (InputItem, bool) {
```
Sandertv, line 27: Would remove the `to` here.

server/item/recipe/item.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return InputItem{Stack: item.NewStack(it, int(i.Count)), Variants: i.MetadataValue == 32767}, true
+}
+
+// inputItems is an array of input items.
+type inputItems []inputItem
+
+// toInputItems converts inputItems into an array of InputItems.
+func (i inputItems) toInputItems() (s []InputItem, ok bool) {
```
Sandertv, line 43: Would remove the `to` here.

server/item/recipe/item.go
```diff
@@ -0,0 +1,94 @@
...
+		Name       string                 `nbt:"name"`
+		Properties map[string]interface{} `nbt:"states"`
+		Version    int32                  `nbt:"version"`
+	} `nbt:"block"`
+	// NBTData contains extra NBTData which may modify the item in other, more discreet ways.
+	NBTData map[string]interface{} `nbt:"data"`
+}
+
+// ToStack converts an output item to an item stack.
+func (o outputItem) ToStack() (item.Stack, bool) {
```
Sandertv, line 73: Would remove the `To`.

server/item/recipe/item.go
```diff
@@ -0,0 +1,94 @@
...
+	if len(o.Name) == 0 {
```
Sandertv, line 74: This check is also unnecessary.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,57 @@
+package recipe
+
+import "github.com/df-mc/dragonfly/server/item"
+
+// Recipe is implemented by all recipe types.
+type Recipe interface {
+	// Inputs returns the items required to craft the recipe.
+	Inputs() []InputItem
+	// UpdateInputs updates the inputs of the recipe.
+	UpdateInputs(inputs []InputItem)
```
Sandertv, line 10: Is there any specific reason we allow this?

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,57 @@
...
+	// Output returns the item that is produced when the recipe is crafted.
+	Output() item.Stack
+	// UpdateOutput updates the output of the recipe.
+	UpdateOutput(output item.Stack)
```
Sandertv, line 14: Same as above?

server/item/recipe/register.go
```diff
@@ -0,0 +1,14 @@
...
+var recipes []Recipe
+
+// Register registers a new recipe.
+func Register(recipe Recipe) {
+	recipes = append(recipes, recipe)
+}
+
+// Recipes returns each recipe in a slice.
+func Recipes() []Recipe {
+	return recipes
```
Sandertv, line 13: Would maybe copy the recipe slice here so that people cannot directly modify the slice and edit the recipes with it.

server/item/recipe/vanilla.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for _, s := range vanillaRecipes.Shapeless {
+		inputs, ok := s.Input.toInputItems()
+		if !ok {
+			continue
+		}
+		output, ok := s.Output.ToStack()
+		if !ok {
+			continue
+		}
```
Sandertv, line 43: Could maybe merge this by checking `ok && outputOK`.

server/item/recipe/vanilla.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for _, s := range vanillaRecipes.Shaped {
+		inputs, ok := s.Input.toInputItems()
+		if !ok {
+			continue
+		}
+		output, ok := s.Output.ToStack()
+		if !ok {
+			continue
+		}
```
Sandertv, line 58: Same as above.

server/session/player.go
```diff
@@ -87,9 +102,11 @@ const (
 	containerHotbar         = 27
 	containerInventory      = 28
 	containerOffHand        = 33
+	containerCraftingOffset = 46
```
Sandertv, line 105: This is a list of container IDs, which the crafting offset, as far as I can tell, is not. Might want to call this something else (unless this is actually used as a container ID? In which case please do let me know).

server/session/handler_item_stack_request.go
```diff
@@ -85,7 +91,7 @@ func (h *ItemStackRequestHandler) handleRequest(req protocol.ItemStackRequest, s
 			err = h.handleBeaconPayment(a, s)
 		case *protocol.CraftCreativeStackRequestAction:
 			err = h.handleCreativeCraft(a, s)
-		case *protocol.CraftResultsDeprecatedStackRequestAction:
+		case *protocol.ConsumeStackRequestAction, *protocol.CraftResultsDeprecatedStackRequestAction:
```
Sandertv, line 94: I'm glad you're ignoring the ConsumeStackRequestAction, we probably shouldn't be using this. (Might even expose a vulnerability in vanilla?)

server/session/handler_item_stack_request.go
```diff
@@ -170,6 +176,58 @@ func (h *ItemStackRequestHandler) handleSwap(a *protocol.SwapStackRequestAction,
 	return nil
 }
 
+// handleCraft handles the Craft stack request action.
+func (h *ItemStackRequestHandler) handleCraft(recipeNetworkID uint32, auto bool, timesCrafted int, s *Session) error {
+	// Get our recipe.
+	r, ok := s.recipeMapping[recipeNetworkID]
+	if !ok {
+		return fmt.Errorf("invalid recipe network id sent")
```
Sandertv, line 184: Might want to add the recipe network ID to this error.

server/session/handler_item_stack_request.go
```diff
@@ -325,16 +383,22 @@ func (h *ItemStackRequestHandler) verifySlots(s *Session, slots ...protocol.Stac
...
-	h.tryAcknowledgeChanges(slot)
+	if err := h.tryAcknowledgeChanges(s, slot); err != nil {
+		return err
+	}
 	if len(h.responseChanges) > 256 {
 		return fmt.Errorf("too many unacknowledged request slot changes")
 	}
+	inv, ok := s.invByID(int32(slot.ContainerID))
+	if !ok {
+		return fmt.Errorf("slot pointed to inventory with id %v, but inventory could not be found", slot.ContainerID)
```
Sandertv, line 394: No need to check this, h.tryAcknowledgeChanges already verifies that the inventory exists.

server/session/handler_item_stack_request.go
```diff
@@ -485,3 +556,184 @@ func (h *ItemStackRequestHandler) reject(id int32, s *Session) {
 	}
 	h.changes = map[byte]map[byte]changeInfo{}
 }
+
+// inputMapFromInputs makes an input map from inputs.
```
Sandertv, line 560: Not clear by this documentation exactly what this is for, might want to expand this documentation a bit.

server/session/handler_item_stack_request.go
```diff
@@ -485,3 +556,184 @@ func (h *ItemStackRequestHandler) reject(id int32, s *Session) {
...
+		st := oldSt.Grow(-inputs[index].Count())
+		h.setItemInSlot(protocol.StackRequestSlotInfo{
+			ContainerID:    containerCraftingGrid,
+			Slot:           slot,
+			StackNetworkID: item_id(st),
+		}, st, s)
+		index++
+	}
+
+	return nil
```
Sandertv, line 738: Lots of semi-duplicated code here. Do you reckon there's any way we could re-use some of the code so these functions can become smaller and easier to understand?

server/session/player.go
```diff
@@ -78,6 +84,15 @@ func (s *Session) sendInv(inv *inventory.Inventory, windowID uint32) {
 	s.writePacket(pk)
 }
 
+const (
+	craftingSizeSmall       = 4
+	craftingSizeLarge       = 9
+	craftingGridSmallOffset = 28
+	craftingGridLargeOffset = 32
+	craftingResultIndex     = 50
+	craftingFlagAll         = 32767
```
Sandertv, line 93: This is simply `math.MaxInt16`.

server/session/session.go
```diff
@@ -66,8 +67,11 @@ type Session struct {
 	openedWindowID                 atomic.Uint32
 	inTransaction, containerOpened atomic.Bool
 	openedWindow, openedPos        atomic.Value
+	openedContainerID              atomic.Uint32
 	swingingArm                    atomic.Bool
 
+	recipeMapping map[uint32]recipe.Recipe
```
Sandertv, line 73: I would toss this map out. This just keeps a lot of memory used that isn't used at any other point than at the start to send the recipes.

server/session/player.go
```diff
@@ -576,15 +593,79 @@ func stackToItem(it protocol.ItemStack) item.Stack {
...
+		networkID := uint32(index) + 1
+		s.recipeMapping[networkID] = i
+
+		switch newRecipe := i.(type) {
+		case *recipe.ShapelessRecipe:
+			recipeList = append(recipeList, &protocol.ShapelessRecipe{
+				RecipeID:        uuid.New().String(),
+				Input:           itemsToRecipeIngredientItems(newRecipe.Inputs()),
+				Output:          []protocol.ItemStack{deleteDamage(stackFromItem(newRecipe.Output()))},
+				Block:           "crafting_table", // This won't always be the case in the future, but for now it's fine.
```
Sandertv, line 644: Consider adding the block used to the recipes somehow, that'll save us work in the future when we do implement other crafting stations.

server/session/session.go
JustTalDevelops: Right now, recipes in the recipe package aren't being stored by their recipe network ID. Should I adjust it to do so, because otherwise I'm unsure on how to do so?

server/item/recipe/recipe.go
JustTalDevelops: Forgot why I did this, doesn't really make much sense, removed.

server/session/handler_item_stack_request.go
JustTalDevelops: I don't have any immediate ideas, but I might (probably) be able to find some ways to merge some bits of code together. Will try this later.

server/item/recipe/dimensions.go
```diff
@@ -0,0 +1,20 @@
...
+type Dimensions [2]int
+
+// DimensionsFrom returns Dimensions from a shape.
+func DimensionsFrom(shape []string) (Dimensions, error) {
+	height := len(shape)
+	if height > 3 || height <= 0 {
+		return Dimensions{}, fmt.Errorf("shaped recipes may only have 1, 2 or 3 rows, not height")
+	}
+
+	width := len(shape[0])
```
Sandertv, line 15: Do we have to check that all rows have the same width?

server/item/recipe/item.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// inputItem converts an inputItemData to an InputItem.
+func (i inputItemData) inputItem() (InputItem, bool) {
+	it, ok := world.ItemByName(i.Name, int16(i.MetadataValue))
+	if !ok {
+		return InputItem{}, false
+	}
+
+	return InputItem{Stack: item.NewStack(it, int(i.Count)), Variants: i.MetadataValue == 32767}, true
```
Sandertv, line 33: Would use `math.MaxInt16` here.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,43 @@
+package recipe
+
+import "github.com/df-mc/dragonfly/server/item"
+
+// Recipe is implemented by all recipe types.
+type Recipe interface {
+	// Inputs returns the items required to craft the recipe.
+	Inputs() []InputItem
```
Sandertv, line 8: Would maybe name this `Input` to match the naming of `Output`. Of course Output only returns one `item.Stack`, but `Input` is a valid plural too so that might be the way to go.

server/session/session.go
Sandertv: You can ignore this, it seemed to be that this map was only used in the `protocolRecipes` method but that was because GitHub collapsed `handler_item_stack_request.go`.

Sandertv (commented):
Got a couple of last comments. When all open comments are resolved I think this is ready to go.

JustTalDevelops:
Looks like this will also have to be delayed to v0.6.0 as well. I just haven't had the time to properly work out the inventory-based crafting system.

server/session/handler_item_stack_request.go
```diff
@@ -183,6 +190,128 @@ func call(ctx *event.Context, slot int, it item.Stack, f func(ctx *event.Context
...
+	_, shaped := craft.(recipe.ShapedRecipe)
+	_, shapeless := craft.(recipe.ShapelessRecipe)
+	if !shaped && !shapeless {
+		return fmt.Errorf("recipe with network id %v is not a shaped or shapeless recipe", a.RecipeNetworkID)
+	}
+
+	input := craft.Input()
+	size := s.craftingSize()
+	offset := s.craftingOffset()
+	consumed := make([]uint32, 0, size)
```
T14Raptor, line 208: What about using a `[]bool` here?

server/session/handler_item_stack_request.go
```diff
@@ -183,6 +190,128 @@ func call(ctx *event.Context, slot int, it item.Stack, f func(ctx *event.Context
...
+	if !ok {
+		return fmt.Errorf("recipe with network id %v does not exist", a.RecipeNetworkID)
+	}
+	_, shaped := craft.(recipe.ShapedRecipe)
+	_, shapeless := craft.(recipe.ShapelessRecipe)
+	if !shaped && !shapeless {
+		return fmt.Errorf("recipe with network id %v is not a shaped or shapeless recipe", a.RecipeNetworkID)
+	}
+
+	input := craft.Input()
```
T14Raptor, line 205: Would inline this

server/session/handler_item_stack_request.go
```diff
@@ -183,6 +190,128 @@ func call(ctx *event.Context, slot int, it item.Stack, f func(ctx *event.Context
...
+		}); ind >= 0 {
+			i = i.Grow(expectancies[ind].Count())
+			expectancies = slices.Delete(expectancies, ind, ind+1)
+		}
+		expectancies = append(expectancies, i)
+	}
+
+	for _, expected := range expectancies {
+		for id, inv := range map[byte]*inventory.Inventory{containerCraftingGrid: s.ui, containerFullInventory: s.inv} {
+			for slot, has := range inv.Slots() {
```
T14Raptor, line 279: If you use `inv.Items()` here, you can get rid of the `has.Empty()` check.

T14Raptor (commented):
Minimal things, not sure how to improve other things, looks good to me otherwise

server/session/handler_item_stack_request.go
JustTalDevelops: I need the slot, and if I use Items, the slot won't be exact.

T14Raptor: Ah, right, might be better to reorder these loops though

JustTalDevelops: That'd cause mismatch issues if the items are split up between different inventories and vary

JustTalDevelops:
This should be complete now.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,93 @@
...
+	Output() []item.Stack
+	// Block returns the block that is used to craft the recipe.
+	Block() world.Block
+	// Priority returns the priority of the recipe. Recipes with lower priority are preferred compared to recipes with
+	// higher priority.
+	Priority() uint32
+}
+
+// ShapelessRecipe is a recipe that has no particular shape.
+type ShapelessRecipe struct {
```
Sandertv, line 22: Would consider calling this `Shapeless` as it already has the `recipe.` prefix.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,93 @@
...
+	return ShapelessRecipe{recipe: recipe{
+		input:    input,
+		output:   output,
+		priority: priority,
+		block:    block,
+	}}
+}
+
+// ShapedRecipe is a recipe that has a specific shape that must be used to craft the output of the recipe.
+type ShapedRecipe struct {
```
Sandertv, line 37: Would consider calling this `Shaped` for the above reason.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,93 @@
...
+	Priority() uint32
+}
+
+// ShapelessRecipe is a recipe that has no particular shape.
+type ShapelessRecipe struct {
+	recipe
+}
+
+// NewShapelessRecipe creates a new shapeless recipe and returns it.
+func NewShapelessRecipe(input []item.Stack, output []item.Stack, block world.Block, priority uint32) ShapelessRecipe {
```
Sandertv, line 27: Would remove the `Recipe` at the end here too.

server/item/recipe/recipe.go
```diff
@@ -0,0 +1,93 @@
...
+
+// ShapedRecipe is a recipe that has a specific shape that must be used to craft the output of the recipe.
+type ShapedRecipe struct {
+	recipe
+	// shape contains the width and height of the shaped recipe.
+	shape Shape
+}
+
+// NewShapedRecipe creates a new shaped recipe and returns it.
+func NewShapedRecipe(input []item.Stack, output []item.Stack, block world.Block, priority uint32, shape Shape) ShapedRecipe {
```
Sandertv, line 44: Would remove the `Recipe` suffix here as well.

Sandertv, line 44: Documentation for this could be improved. Would clarify that `block` would be the block required for being able to craft this recipe. I would also consider removing `priority` from the API. I don't think that is useful generally.

server/item/recipe/recipe.go
Sandertv, line 27: This could also do with a better documentation. See comment on `NewShapedRecipe`.

server/session/handler_item_stack_request.go
```diff
@@ -183,6 +190,135 @@ func call(ctx *event.Context, slot int, it item.Stack, f func(ctx *event.Context
...
+
+	expectancies := make([]item.Stack, 0, len(input))
+	for _, i := range input {
+		if i.Empty() {
+			// We don't actually need this item - it's empty, so avoid putting it in our expectancies.
+			continue
+		}
+
+		if ind := slices.IndexFunc(expectancies, func(stack item.Stack) bool {
+			return stack.Comparable(i)
```
Sandertv, line 268: If you change this to `i.Comparable`, you can simply do `slices.IndexFunc(expectancies, i.Comparable)`.

server/item/stack.go
```diff
@@ -326,6 +337,12 @@ func (s Stack) Comparable(s2 Stack) bool {
 	return true
 }
 
+// Variants returns true if multiple variants are allowed for the item stack. This is used mainly for crafting recipe
+// inputs.
+func (s Stack) Variants() bool {
```
Sandertv, line 342: I don't really know about adding this and `WithVariants`. This is purely for recipes and the method doesn't actually mean anything to the Stack.

Sandertv (commented):
Mostly looks good, got a couple of comments.

server/item/stack.go
JustTalDevelops: I wouldn't say this necessarily - it can be useful externally, too, for example, using `Comparable` for only the type of item. It also prevents the entire "RecipeItem" wrapper too.

Sandertv (approved):
Looks good, thanks a lot for the PR!

PR #332, opened 2021-11-25
Implement splash potions

JustTalDevelops:


server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+		owner: owner,
+	}
+	s.transform = newTransform(s, pos)
+
+	return s
+}
+
+// Name ...
+func (s *SplashPotion) Name() string {
+	return "Splash Variant"
```
DaPigGuy, line 54:
```suggestion
	return "Splash Potion"
```

Accidental refactor?

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+		colour := color.RGBA{R: 0x38, G: 0x5d, B: 0xc6, A: 0xff}
+		if hasEffects {
+			colour, _ = effect.ResultingColour(effects)
+		}
+
+		w.AddParticle(pos, particle.Splash{Colour: colour})
+		w.PlaySound(pos, sound.GlassBreak{})
+
+		if hasEffects {
+			ignoreFunc := func(entity world.Entity) bool {
```
T14Raptor, line 134: I would call this `ignore`

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+				_, canSplash := entity.(splashable)
+				if !canSplash || entity == s {
```
T14Raptor, line 136: You can get rid of this if statement and just have `return !canSplash || entity == s`

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+					return true
+				}
+				return false
+			}
+
+			for _, otherEntity := range w.EntitiesNearby(aabb.GrowVec3(mgl64.Vec3{4.125, 2.125, 4.125}), ignoreFunc) {
+				splashEntity := otherEntity.(splashable)
+
+				distance := world.Distance(splashEntity.Position().Add(mgl64.Vec3{0, splashEntity.EyeHeight(), 0}), pos)
```
T14Raptor, line 145:
```suggestion
				distance := world.Distance(EyePosition(splashEntity), pos)
```

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+	return s.New(
+		nbtconv.MapVec3(data, "Pos"),
+		nbtconv.MapVec3(data, "Motion"),
+		float64(nbtconv.MapFloat32(data, "Pitch")),
+		float64(nbtconv.MapFloat32(data, "Yaw")),
+	)
+}
+
+// EncodeNBT encodes the SplashPotion entity's properties as a map and returns it.
+func (s *SplashPotion) EncodeNBT() map[string]interface{} {
```
T14Raptor, line 227: This should also save the entity's potion type, correct?

server/item/splash_potion.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"github.com/df-mc/dragonfly/server/item/potion"
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/df-mc/dragonfly/server/world/sound"
+)
+
+// SplashPotion is an item that grants effects when thrown.
+type SplashPotion struct {
+	// Variant is the type of splash potion.
+	Variant potion.Potion
```
T14Raptor, line 12: I would call this `Type` for consistency with the drinkable potion item.

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+	yaw, pitch := s.Rotation()
+	return map[string]interface{}{
+		"Pos":    nbtconv.Vec3ToFloat32Slice(s.Position()),
+		"Yaw":    yaw,
+		"Pitch":  pitch,
+		"Motion": nbtconv.Vec3ToFloat32Slice(s.Velocity()),
+		"Damage": 0.0,
+	}
+}
```
DaPigGuy, line 236: Decode/encode potion type missing

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+					continue
+				}
+
+				distanceMultiplier := 1 - (distance / 4)
+				if entityResult, ok := result.(trace.EntityResult); ok && entityResult.Entity() == otherEntity {
+					distanceMultiplier = 1.0
+				}
+
+				for _, eff := range effects {
+					if (eff.Type() == effect.InstantHealth{} || eff.Type() == effect.InstantDamage{}) {
```
DaPigGuy, line 156: check if type implements effect.LastingType?

server/entity/splash_potion.go
JustTalDevelops: Woops, yeah

server/entity/splash_potion.go
```diff
@@ -0,0 +1,245 @@
...
+	transform
+	yaw, pitch float64
+
+	ticksLived int
+
+	closeNextTick bool
+
+	owner world.Entity
+
+	variant potion.Potion
```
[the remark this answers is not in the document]
JustTalDevelops: I can't, that conflicts with Go's built in `type`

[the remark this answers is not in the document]
T14Raptor: `t`

JustTalDevelops:
Got all the previous comments resolved. Anything else I should change?

server/entity/splash_potion.go
```diff
@@ -68,14 +67,14 @@ func (s *SplashPotion) AABB() physics.AABB {
 func (s *SplashPotion) SetVariant(variant potion.Potion) {
```
T14Raptor, line 66: Don't forget to change the method names etc.

server/entity/splash_potion.go
```diff
@@ -0,0 +1,239 @@
...
+		w.AddParticle(m.pos, particle.Splash{Colour: colour})
+		w.PlaySound(m.pos, sound.GlassBreak{})
+
+		if hasEffects {
+			ignore := func(entity world.Entity) bool {
+				_, canSplash := entity.(splashable)
+				return !canSplash || entity == s
+			}
+
+			for _, otherEntity := range w.EntitiesWithin(aabb.GrowVec3(mgl64.Vec3{4.125, 2.125, 4.125}), ignore) {
```
T14Raptor, line 139: This aabb should be grown even further (maybe by 3), then should check if the otherEntity collides with the `aabb.GrowVec3(mgl64.Vec3{4.125, 2.125, 4.125})`

server/entity/splash_potion.go
```diff
@@ -0,0 +1,239 @@
...
+					distanceMultiplier = 1.0
+				}
+
+				for _, eff := range effects {
+					if _, ok := eff.Type().(effect.LastingType); !ok {
+						splashEntity.AddEffect(eff.WithPotency(distanceMultiplier))
+						continue
+					}
+
+					splashEntity.AddEffect(eff.WithDuration(time.Duration(float64(eff.Duration().Milliseconds())*0.75*distanceMultiplier) * time.Millisecond))
```
T14Raptor, line 158: Shouldn't this continue if the duration is <= a second?

server/entity/splash_potion.go
```diff
@@ -0,0 +1,239 @@
...
+	closeNextTick bool
+
+	owner world.Entity
+
+	t potion.Potion
+	c *ProjectileComputer
+}
+
+// NewSplashPotion ...
+func NewSplashPotion(pos mgl64.Vec3, yaw, pitch float64, owner world.Entity) *SplashPotion {
```
T14Raptor, line 34: I would probably put the potion type as a argument here

server/entity/splash_potion.go
```diff
@@ -0,0 +1,239 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.owner = owner
+}
+
+// DecodeNBT decodes the properties in a map to a SplashPotion and returns a new SplashPotion entity.
+func (s *SplashPotion) DecodeNBT(data map[string]interface{}) interface{} {
+	p := s.New(
+		nbtconv.MapVec3(data, "Pos"),
+		nbtconv.MapVec3(data, "Motion"),
```
T14Raptor, line 211: After you add the potion type to the constructor, you can just use that instead of this, it'll be cleaner

T14Raptor (commented):
Few more things, everything else looks good

server/entity/splash_potion.go
JustTalDevelops: I can't change the constructor of New without causing conflicts to other implementations of New. Do you want me to use NewSplashPotion instead? If so, I'll still have to have another line after it anyways, to set the velocity, since NewSplashPotion doesn't allow for it.

JustTalDevelops: Is that okay?

JustTalDevelops: Also, by changing the constructor of NewSplashPotion, New will somehow need to set a default potion otherwise that won't work either. If I should continue this way, shall I just make it set it to water?

T14Raptor: I was referring to NewSplashPotion, not .New. NewSplashPotion should have the potion type anyways for consistency.

JustTalDevelops: New does use NewSplashPotion though, so I will have to initialize it with some potion anyways. I assume this should just be Water?

JustTalDevelops:
Anything else that should be changed?

server/entity/effect/effect.go
```diff
@@ -57,6 +57,21 @@ func NewAmbient(t LastingType, lvl int, d time.Duration) Effect {
...
 
+// potentType represents an effect type which can have its potency changed.
+type potentType interface {
+	Type
+	// WithPotency updates the potency of the type with the one given and returns it.
+	WithPotency(potency float64) Type
+}
+
+// WithPotency returns a new effect with the given potency, if the effect type allows for it.
+func (e Effect) WithPotency(potency float64) Effect {
```
Sandertv, line 68: I don't think we should expose this as a method. I think it'd be a better idea if we exposed `PotentType`. This method doesn't do anything for most effects.

server/entity/effect/effect.go
```diff
@@ -74,6 +89,12 @@ func (e Effect) Level() int {
 	return e.lvl
 }
 
+// WithDuration returns a new effect with the given duration.
+func (e Effect) WithDuration(d time.Duration) Effect {
```
Sandertv, line 93: This method also doesn't apply for instant effects, it should be possible to do this a different way.

Sandertv: To elaborate on that, say you do have an `effect.Effect` value, you could just create a new effect with the same level and type but with a different duration.

server/entity/splash_potion.go
```diff
@@ -0,0 +1,248 @@
...
+	return "minecraft:splash_potion"
+}
+
+// AABB ...
+func (s *SplashPotion) AABB() physics.AABB {
+	return physics.NewAABB(mgl64.Vec3{-0.125, 0, -0.125}, mgl64.Vec3{0.125, 0.25, 0.125})
+}
+
+// SetType ...
+func (s *SplashPotion) SetType(t potion.Potion) {
```
Sandertv, line 68: Any reason to expose this method? We should just pass it to the New method. See 5157a8cae26406d6327a21695f6b923900c1f1d9 for what I mean. I think it's better we keep the potion type immutable.

server/entity/splash_potion.go
```diff
@@ -0,0 +1,248 @@
...
+				return !canSplash || entity == s
+			}
+
+			for _, otherEntity := range w.EntitiesWithin(aabb.GrowVec3(mgl64.Vec3{8.25, 4.25, 8.25}), ignore) {
+				splashEntity := otherEntity.(splashable)
+				if !splashEntity.AABB().Translate(splashEntity.Position()).IntersectsWith(aabb.GrowVec3(mgl64.Vec3{4.125, 2.125, 4.125})) {
+					continue
+				}
+
+				distance := world.Distance(EyePosition(splashEntity), m.pos)
```
Sandertv, line 143: Eye position sounds... weird? Where did you find that?

server/entity/splash_potion.go
```diff
@@ -0,0 +1,248 @@
...
+					}
+
+					distanceAccountedDuration := time.Duration(float64(eff.Duration().Milliseconds())*0.75*distanceMultiplier) * time.Millisecond
+					if distanceAccountedDuration < time.Second {
+						continue
+					}
+					splashEntity.AddEffect(eff.WithDuration(distanceAccountedDuration))
+				}
+			}
+		} else if blockResult, ok := result.(trace.BlockResult); ok && s.t.Equals(potion.Water()) {
```
Sandertv, line 166: These water splash potions also deal damage to endermen, blazes, striders and snow golems. Could you add a TODO for that?

server/entity/splash_potion.go
```diff
@@ -0,0 +1,248 @@
...
+					distanceMultiplier = 1
+				}
+
+				for _, eff := range effects {
+					if _, ok := eff.Type().(effect.LastingType); !ok {
+						splashEntity.AddEffect(eff.WithPotency(distanceMultiplier))
+						continue
+					}
+
+					distanceAccountedDuration := time.Duration(float64(eff.Duration().Milliseconds())*0.75*distanceMultiplier) * time.Millisecond
```
Sandertv, line 159: Any reason for converting this to Milliseconds? In the end it's rounded to ticks when the effect is added anyway.

Sandertv: Also this variable name (and a couple ones above) is really long, consider using a bit of a shorter name, makes it easier to see the logic.

server/entity/splash_potion.go
```diff
@@ -0,0 +1,248 @@
...
+
+// ignores returns whether the SplashPotion should ignore collision with the entity passed.
+func (s *SplashPotion) ignores(entity world.Entity) bool {
+	_, ok := entity.(Living)
+	return !ok || entity == s || (s.ticksLived < 5 && entity == s.owner)
+}
+
+// New creates a SplashPotion with the position, velocity, yaw, and pitch provided. It doesn't spawn the SplashPotion,
+// only returns it.
+func (s *SplashPotion) New(pos, vel mgl64.Vec3, yaw, pitch float64) world.Entity {
```
Sandertv, line 195: Would accept the potion type here too. No need to implement the Projectile interface, I removed that in the latest commit. Make sure to pull that commit though.

server/item/splash_potion.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Use ...
+func (s SplashPotion) Use(w *world.World, user User, ctx *UseContext) bool {
+	splash, ok := world.EntityByName("minecraft:splash_potion")
+	if !ok {
+		return false
+	}
+
+	p, ok := splash.(projectile)
```
Sandertv, line 27: I removed the `projectile` interface in the latest commit. Make sure to replace that with an interface that has the `New` method following the signature of SplashPotion.New.

server/item/splash_potion.go
```diff
@@ -0,0 +1,61 @@
...
+		return false
+	}
+
+	yaw, pitch := user.Rotation()
+	e := p.New(eyePosition(user), directionVector(user).Mul(0.5), yaw, pitch)
+	if o, ok := e.(owned); ok {
+		o.Own(user)
+	}
+	if pot, ok := e.(splashPotion); ok {
+		pot.SetType(s.Type)
```
Sandertv, line 38: With that changed New method you should be able to remove this.

server/item/splash_potion.go
```diff
@@ -0,0 +1,61 @@
...
+func (s SplashPotion) EncodeItem() (name string, meta int16) {
+	return "minecraft:splash_potion", int16(s.Type.Uint8())
+}
+
+// splashPotion represents an entity instance of a SplashPotion.
+type splashPotion interface {
+	// SetType sets the type of the splash potion.
+	SetType(t potion.Potion)
+	// Type returns the type of the splash potion.
+	Type() potion.Potion
```
Sandertv, line 60: Would remove these 2 methods and replace it with the `New` method.

server/session/entity_metadata.go
```diff
@@ -58,6 +59,15 @@ func parseEntityMetadata(e world.Entity) entityMetadata {
...
 	}
+	if s, ok := e.(splash); ok {
+		pot := s.Type()
+		id := pot.Uint8()
+
+		m[dataKeyPotionAuxValue] = int16(id)
+		if len(pot.Effects()) > 0 {
+			m.setFlag(dataKeyFlags, dataFlagEnchanted)
+		}
+	}
```
Sandertv, line 70: Can compact this a bit without really harming readability.

Sandertv (commented):
Got a couple of comments, but they should be pretty easy to resolve. Looks pretty good so far!

server/entity/splash_potion.go
JustTalDevelops: https://github.com/df-mc/dragonfly/blob/157e9240dfc865f6512bb7b200b16ff11d5c760f/server/entity/direction.go#L52

Sandertv: Yeah but I mean, it doesn't really make sense to me that this checks the distance of the splash potion compared to the eye position of the entity.

server/entity/splash_potion.go
JustTalDevelops: How would I do it otherwise?

Sandertv: Just remove the `Milliseconds()` and `* time.Milliseconds`?

JustTalDevelops:
Anything else I can change @Sandertv?

server/entity/splash_potion.go
```diff
@@ -0,0 +1,233 @@
...
+func (s *SplashPotion) Rotation() (float64, float64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.yaw, s.pitch
+}
+
+// splashable represents an entity that can be splashed by a potion.
+type splashable interface {
+	Living
+	Eyed
```
Sandertv, line 77: Can remove this now.

JustTalDevelops:
Is there anything else I should change here?

Sandertv (approved):
Looks good to me, feel free to merge when you think it's ready.

PR #346, opened 2021-12-05
add a drops parameter to HandleBlockBreak

AndreasHGK:
This PR makes it possible to alter what items are dropped when a block is mined by a player.


Sandertv (approved):

PR #355, opened 2021-12-10
Fixed the vanilla client attempting to chunk blend a Dragonfly world

JustTalDevelops:


server/world/mcdb/provider.go
```diff
@@ -28,6 +28,13 @@ type Provider struct {
 // chunkVersion is the current version of chunks.
 const chunkVersion = 27
 
+var (
+	// lastOpenedWithVersion is the current last opened with version, required by the latest Minecraft data provider.
+	lastOpenedWithVersion = []int32{1, 18, 0, 2, 0}
+	// minimumCompatibleClientVersion is the minimum compatible client version, required by the latest Minecraft data provider.
+	minimumCompatibleClientVersion = []int32{1, 18, 0, 0, 0}
```
Sandertv, line 35: Do you reckon we could base these on the protocol version? I'm not a huge fan of having to update these everytime. I suppose the minimum might not be easy to do but with the lastOpenedWithVersion that should be possible(?).

Sandertv: Also are all 3 of NetworkVersion, LastOpenedWithVersion and MinimumCompatibleClientVersion needed for it to work?

JustTalDevelops: I haven't tried each one of them separately, but honestly better to be safe then sorry. I can look into basing these off of protocol.CurrentVersion though.

Sandertv: That would be a good improvement.

Sandertv: Any update on this? Do you reckon we could resolve this and get this merged by the end of the month?

JustTalDevelops: Updated.

JustTalDevelops:
Another review on this would be great.

Sandertv (approved):
Looks good to me!

PR #362, opened 2021-12-16
Adds a function that changes the nickname of an entity only on the side of the player

max-xoo:
I don't have anything to add, at least I don't think we can add anything else


JustTalDevelops:
Closing this because of various reasons discussed on Discord. (from a while ago)

PR #368, opened 2021-12-18
made it so you can set lines without following the order 1, 2, 3, 4

RestartFU:
imo, this is something that we need to be able to do.


Sandertv:
I'm more of a fan of adding an `Insert` method as opposed to replacing it with a map. This breaks adding new lines to the scoreboard after removing one.

RestartFU:
alright

PR #369, opened 2021-12-19
Implement item frames

JustTalDevelops:


server/block/item_frame.go
```diff
@@ -0,0 +1,167 @@
...
+	empty
+	transparent
+
+	// Facing is the direction from the frame to the block.
+	Facing cube.Face
+	// Item is the item that is displayed inside the frame.
+	Item item.Stack
+	// Rotations is the number of rotations for the item in the frame. Each rotation is 45 degrees, with the exception
+	// being maps having 90 degree rotations.
+	Rotations uint8
```
Sandertv, line 25: Would make this an `int` for the API.

server/block/item_frame.go
```diff
@@ -0,0 +1,167 @@
...
+	if !w.Block(pos.Side(i.Facing)).Model().FaceSolid(pos.Side(i.Facing), i.Facing.Opposite(), w) {
+		w.BreakBlock(pos)
+	}
+}
+
+// allItemFrames ...
+func allItemFrames() (frames []world.Block) {
+	for _, f := range cube.Faces() {
+		for _, g := range []bool{true, false} {
+			for _, l := range []bool{true, false} {
```
Sandertv, line 161: These loops don't really add anything. I'd just have 4 lines of append with large=`true`/`false` and glowing=`true`/`false`.

server/block/item_frame.go
JustTalDevelops: Changed.

server/block/item_frame.go
JustTalDevelops: Changed.

server/block/item_frame.go
```diff
@@ -0,0 +1,166 @@
...
+	Item item.Stack
+	// Rotations is the number of rotations for the item in the frame. Each rotation is 45 degrees, with the exception
+	// being maps having 90 degree rotations.
+	Rotations int
+	// Drops is the chance of the item dropping when the frame is broken. In vanilla, this is always 1.0.
+	Drops float64
+	// Glowing makes the frame the glowing variant.
+	Glowing bool
+	// Large makes the frame show up large. This typically only occurs with maps.
+	Large bool
```
Sandertv, line 31: Does this field actually work with non-map items?

server/block/item_frame.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Facing is the direction from the frame to the block.
+	Facing cube.Face
+	// Item is the item that is displayed inside the frame.
+	Item item.Stack
+	// Rotations is the number of rotations for the item in the frame. Each rotation is 45 degrees, with the exception
+	// being maps having 90 degree rotations.
+	Rotations int
+	// Drops is the chance of the item dropping when the frame is broken. In vanilla, this is always 1.0.
+	Drops float64
```
Sandertv, line 27: Would consider renaming this to something like `DropChance`. The name of this is a bit ambiguous right now.

Sandertv (commented):
Got some small nitpicks. Also please, once you've got those fixed, make sure to pull the latest commit and re-run the block hash generator to make sure these are up to date.

server/block/item_frame.go
DaPigGuy: No

server/block/item_frame.go
```diff
@@ -0,0 +1,166 @@
...
+	w.SetBlock(pos, i)
+}
+
+// UseOnBlock ...
+func (i ItemFrame) UseOnBlock(pos cube.Pos, face cube.Face, _ mgl64.Vec3, w *world.World, user item.User, ctx *item.UseContext) bool {
+	pos, face, used := firstReplaceable(w, pos, face, i)
+	if !used {
+		return false
+	}
+	if !w.Block(pos.Side(face.Opposite())).Model().FaceSolid(pos.Side(face.Opposite()), face, w) {
```
DaPigGuy, line 79: This is preventing quite a lot of valid placements

DaPigGuy (changes requested):
Block picking item frames in vanilla Minecraft will give the item within the item frame & not the item frame itself, if not empty.

server/block/item_frame.go
Sandertv: Maybe in that case it'd be better not to expose this as a field but base it on whether the item is a map or not? We obviously don't have maps right now but we could add a TODO for that.

JustTalDevelops:
> * Block picking item frames in vanilla Minecraft will give the item within the item frame & not the item frame itself, if not empty. This should also be implemented within dragonfly for both parity & QoL.
> 
>     * Because dropping items from an item frame currently relies on the Punchable interface which itself relies on the PlayerAction packet, the ItemFrameDropItem packet is unhandled resulting in debug messages. Either the correct packet should be handled, or a nil handler should be provided.

To solve the first issue, perhaps we should have a "Pickable" interface which certain blocks could inherit to give a different item stack compared to the block itself, and check for that interface whenever a block pick is attempted?

The second problem should be quite straightforward to fix, by setting a nil handler.

server/block/item_frame.go
JustTalDevelops: Could you give an example of a few placements that should be possible but don't work?

server/block/item_frame.go
JustTalDevelops: Renamed

Sandertv (approved):
This looks good. If everything is tested, please feel free to merge.

JustTalDevelops:
All tested, thank you for the review!

PR #371, opened 2021-12-24
Add stone bricks

AndreasHGK:


cmd/blockhash/main.go
```diff
@@ -80,9 +80,11 @@ func (b *hashBuilder) sortNames() {
...
 		b.names = append(b.names, name)
 	}
-	sort.Slice(b.names, func(i, j int) bool {
-		return b.names[i] < b.names[j]
-	})
+	sort.Slice(
+		b.names, func(i, j int) bool {
+			return b.names[i] < b.names[j]
+		},
+	)
```
Sandertv, line 87: If you can i'd check to see if you can disable this reformatting.

server/block/stone_bricks.go
```diff
@@ -0,0 +1,36 @@
+package block
+
+import "github.com/df-mc/dragonfly/server/world"
+
+type StoneBricks struct {
```
Sandertv, line 5: Don't forget to document this.

Sandertv (commented):
Looking pretty good! Like I mentioned in one of the comments, see if you can disable the reformatting thing in GoLand that forces everything on a new line. Just one documentation thing apart from that but the PR looks good otherwise!

server/block/stone_bricks_type.go
```diff
@@ -0,0 +1,96 @@
...
+	return StoneBricksType{stoneBricks(2)}
+}
+
+// ChiseledStoneBricks is the chiseled variant of stone bricks.
+func ChiseledStoneBricks() StoneBricksType {
+	return StoneBricksType{stoneBricks(3)}
+}
+
+// SmoothStoneBricks is the smooth variant of stone bricks.
+func SmoothStoneBricks() StoneBricksType {
```
Sandertv, line 33: I don't think this is a thing, right? Do you reckon we could remove this?

PR #376, opened 2021-12-28
Implement bows and arrows

JustTalDevelops:


JustTalDevelops:
Almost, am going to implement tipped arrows and then this should be ready. Probably will have to extend this to v0.6.0 as a result though, as I'm not exactly sure how I would implement them at this stage, unless you're OK with them not in this implementation.

Sandertv:
We can postpone it to v0.6.0, no need to rush! Let me know if there's anything specific I can do.

JustTalDevelops:
Sounds good, will just postpone it then, probably will have this completely done early January.

JustTalDevelops:
Think this is ready for review now.

server/entity/arrow.go
```diff
@@ -0,0 +1,285 @@
...
+	closeNextTick, critical bool
+
+	owner                        world.Entity
+	shotByPlayer, shotInCreative bool
+
+	c *ProjectileComputer
+}
+
+// NewArrow ...
+func NewArrow(pos mgl64.Vec3, yaw, pitch float64, owner world.Entity, critical, shotByPlayer, shotInCreative bool, baseDamage float64, tip potion.Potion) *Arrow {
```
Sandertv, line 43: This seems like a ton of fields crammed into one function. First of all I'd move the `tip` parameter to a separate `NewTippedArrow()` function.

Sandertv: I would also rename shotInCreative to something that indicates its actual functionality: Not receiving an item when picking it up.

Sandertv: Same for `shotByPlayer`, would rename that to a name that implies it cannot be picked up.

server/entity/arrow.go
```diff
@@ -0,0 +1,285 @@
...
+	a.critical = critical
+	a.mu.Unlock()
+
+	for _, v := range a.World().Viewers(a.Position()) {
+		v.ViewEntityState(a)
+	}
+}
+
+// Tip ...
+func (a *Arrow) Tip() potion.Potion {
```
Sandertv, line 99: Please document this.

server/entity/arrow.go
```diff
@@ -0,0 +1,285 @@
...
+
+// Critical ...
+func (a *Arrow) Critical() bool {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.critical
+}
+
+// SetCritical ...
+func (a *Arrow) SetCritical(critical bool) {
```
Sandertv, line 88: Please document this.

server/entity/arrow.go
```diff
@@ -0,0 +1,285 @@
...
+
+// CollisionPos ...
+func (a *Arrow) CollisionPos() cube.Pos {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.collidedBlockPos
+}
+
+// Critical ...
+func (a *Arrow) Critical() bool {
```
Sandertv, line 81: Please document this.

server/entity/arrow.go
```diff
@@ -0,0 +1,285 @@
...
+	return "Arrow"
+}
+
+// EncodeEntity ...
+func (a *Arrow) EncodeEntity() string {
+	return "minecraft:arrow"
+}
+
+// CollisionPos ...
+func (a *Arrow) CollisionPos() cube.Pos {
```
Sandertv, line 74: Please document this.

server/entity/damage/source.go
```diff
@@ -114,3 +120,8 @@ func (SourceFireTick) ReducedByArmour() bool {
 func (SourceLava) ReducedByArmour() bool {
 	return true
 }
+
+// ReducedByArmour ...
+func (SourceProjectile) ReducedByArmour() bool {
+	return false
```
Sandertv, line 126: This doesn't sound right?

server/session/player.go
```diff
@@ -227,7 +227,7 @@ func (s *Session) Transfer(ip net.IP, port int) {
 // SendGameMode sends the game mode of the Controllable entity of the session to the client. It makes sure the right
 // flags are set to create the full game mode.
 func (s *Session) SendGameMode(mode world.GameMode) {
-	flags, id, perms := uint32(0), int32(packet.GameTypeSurvivalSpectator), uint32(0)
+	flags, id, perms := uint32(0), int32(packet.GameTypeSurvival), uint32(0)
```
Sandertv, line 230: Have you tested that this does not result in any broken gamemodes? What's the reason for this change?

Sandertv (commented):
Got a couple of comments to have a look at. Might need another review though, let me know once you've got these resolved!

server/session/player.go
JustTalDevelops: This was a mistake I made in the v1.17.0 update, this was always meant to be GameTypeSurvival. It seemed to have not caused any issues until implementing bows, as it causes the bow to always release client side.

JustTalDevelops: If you back a few DF versions, you'll see that this was also GameTypeSurvival, as it should have been.

server/entity/damage/source.go
JustTalDevelops: Oops, this should be true, yes. Will change this.

server/entity/arrow.go
JustTalDevelops: The thing is, `shotInCreative` only does not allow you to receive an item when picking it up when the player's gamemode is not creative. Personally I feel like `shotByPlayer` & `shotByCreative` make the most sense for what they do: control specific pickup options based on the player and their gamemode?

JustTalDevelops: Also, when I do move over `tip` to it's own `NewTippedArrow()` function, how should that look in the means of the projectile initialization? Should the `New` function switch over to use `NewTippedArrow()` as well, or shall I create another function for `NewTipped()` which does the same thing as `New` but specifically for tipped arrows?

server/session/entity_metadata.go
```diff
@@ -50,6 +50,11 @@ func parseEntityMetadata(e world.Entity) entityMetadata {
 	if u, ok := e.(using); ok && u.UsingItem() {
 		m.setFlag(dataKeyFlags, dataFlagUsingItem)
 	}
+	if c, ok := e.(arrow); ok {
+		if c.Critical() {
+			m.setFlag(dataKeyFlags, dataFlagCritical)
+		}
+	}
```
DaPigGuy, line 57:
```suggestion
	if c, ok := e.(arrow); ok && c.Critical() {
		m.setFlag(dataKeyFlags, dataFlagCritical)
	}
```

server/entity/arrow.go
T14Raptor: > Personally I feel like shotByPlayer & shotByCreative make the most sense for what they do: control specific pickup options based on the player and their gamemode?
The name's "shotByPlayer" and "shotByCreative" does not make it obvious at all what they actually do. And although their behaviors are only used in vanilla when a player shoots it and/or is in creative, it does not mean that the behaviors would like to be used by developers when making things such as mini-games

server/item/bow.go
```diff
@@ -0,0 +1,96 @@
...
+		}); ok {
+			arr := arrow.Item().(Arrow)
+			tip = arr.Tip
+
+			ctx.DamageItem(1)
+			ctx.Consume(arrow.Grow(-arrow.Count() + 1))
+		} else {
+			return
+		}
+	}
```
DaPigGuy, line 60: Currently not setting arrow tips for creative players

JustTalDevelops: Changed.

server/item/bow.go
```diff
@@ -0,0 +1,98 @@
...
+	force := math.Min((t*t+t*2)/3, 1)
+	if force < 0.1 {
+		return
+	}
+
+	var tip potion.Potion
+	creative := releaser.GameMode().CreativeInventory()
+	if arrow, ok := ctx.FirstFunc(func(stack Stack) bool {
+		name, _ := stack.Item().EncodeItem()
+		return name == "minecraft:arrow"
```
T14Raptor, line 49: Why not use a type assertion here?

server/player/player.go
```diff
@@ -1191,10 +1197,53 @@ func (p *Player) UseItem() {
 // the item started being used.
 func (p *Player) ReleaseItem() {
```
mmm545, line 1198: shouldn't this also be applicable for the item.Releasable interface too? (talking about the documentation)

mmm545 (changes requested):

server/item/bow.go
JustTalDevelops: Changed.

server/player/player.go
JustTalDevelops: Changed.

server/entity/arrow.go
JustTalDevelops: Just refactored both of those variables, but I'm still not sure on the `NewTippedArrow()` discussion. Could I have some thoughts on what I mentioned above?

server/entity/arrow.go
```diff
@@ -0,0 +1,333 @@
...
+	defer a.mu.Unlock()
+	return a.obtainArrowOnPickup
+}
+
+// ObtainArrowOnPickup sets whether the arrow should or should not be obtained when picked up by players.
+func (a *Arrow) ObtainArrowOnPickup(obtainArrowOnPickup bool) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.obtainArrowOnPickup = obtainArrowOnPickup
+}
```
Sandertv, line 141: Lots of methods here that I don't think we need necessarily. Is there any merit of having these getters and especially setters? These are passed in the `NewArrow` method and it sounds like a better idea to have all these things, like whether an arrow is received on pickup and if pickup is allowed, be immutable.

JustTalDevelops: Changed. Will keep a way to set and get base damage though, because I think that could be useful to some people.

JustTalDevelops:
@Sandertv Could you please elaborate on this discussion a little?
https://github.com/df-mc/dragonfly/pull/376#discussion_r777156216

server/entity/arrow.go
Sandertv: I'd have a `NewTippedArrow` which accepts a `tip` parameter, then a `NewArrow` function which does not and calls NewTippedArrow with an 'empty' potion effect.

JustTalDevelops: Should `New` use `NewTippedArrow` then?

JustTalDevelops:
Think this should be ready for another review?

server/entity/damage/source.go
```diff
@@ -45,12 +45,20 @@ type SourceFireTick struct{}
...
 // SourceLightning is used for damage caused by being struck by lightning.
 type SourceLightning struct{}
 
+// SourceProjectile is used for damage caused by a projectile.
+type SourceProjectile struct {
+	// Owner is the entity that was responsible for the projectile.
+	Owner world.Entity
+}
+
+type SourceThorns struct{}
```
Sandertv, line 60: This type hasn't been documented (and shouldn't be here 😉).

Sandertv (approved):
I think this is ready to go. Thanks for the PR!

PR #377, opened 2021-12-28
Add an Effect method to Player

Prim69:


Sandertv:
What in specific would you use this for?

Prim69:
Something like /nightvision, if the player has the effect then remove it, otherwise add it

Or maybe someone wants to display the players health a certain color based on what effect they have
image
image
image


I also wanted to implement crits and saw that Pocketmine checks if the player doesn't have blindness before applying crits.

Sandertv:
I see the use for that. I propose we change this to a `Player.Effect(effectType) (effect, ok)` method though so you can actually get the effect itself too.

Prim69:
That's done

RestartFU:
> That's done

I think he meant to return the effet itself, not the Type.

Prim69:
Oops, thanks, fixed it

Prim69:
This is ready for a review

Sandertv:
Consider adding the Effect method to the EffectManager so that it can be re-used and doesn't require iterating over the effects.

server/entity/effect.go
```diff
@@ -75,6 +75,18 @@ func (m *EffectManager) Remove(e effect.Type, entity Living) {
...
+// Effect returns the effect instance and true if the entity has the effect otherwise it will return an empty effect
+// instance and false.
+func (m *EffectManager) Effect(e effect.Type) (effect.Effect, bool) {
+	m.mu.Lock()
+	existing, ok := m.effects[reflect.TypeOf(e)]
+	m.mu.Unlock()
+	if ok {
+		return existing, ok
+	}
+	return effect.Effect{}, ok
```
Sandertv, line 87: Can simplify this to `return existing, ok`

Sandertv: This PR should be ready to go once you resolve this.

server/entity/effect.go
```diff
@@ -75,6 +75,15 @@ func (m *EffectManager) Remove(e effect.Type, entity Living) {
 	}
 }
 
+// Effect returns the effect instance and true if the entity has the effect otherwise it will return an empty effect
+// instance and false.
+func (m *EffectManager) Effect(e effect.Type) (effect.Effect, bool) {
+	m.mu.Lock()
+        existing, ok := m.effects[reflect.TypeOf(e)]
+        m.mu.Unlock()
+        return existing, ok
```
T14Raptor, line 84: Fix formatting here

After this, we can merge it

Sandertv (approved):

PR #380, opened 2022-01-01
Add a method to send a jukebox popup to Player

Prim69:
The method was there in session but it was never added to player


Sandertv:
Looks good, thank you!

PR #389, opened 2022-01-03
player/chat/subscriber.go: add logrus bundle subscriber

Mohamed587100:


Sandertv:
I'm not sure if we should add this into dragonfly. Might be better put into a separate library.

Mohamed587100:
> I'm not sure if we should add this into dragonfly. Might be better put into a separate library.

but I think dragonfly assumes using logrus
just like

PR #390, opened 2022-01-03
Added a few building blocks

mmm545:
- Hay bale
- Nether wart block and its warped variant
- Purpur block and its variants

### DIFF
--- a/server/block/hay_bale.go
+++ b/server/block/hay_bale.go
@@ -1,0 +1,63 @@
+package block
+
+import (
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/go-gl/mathgl/mgl64"
+)
+
+// HayBale is a decorative, flammable block that can also be used to
+// feed horses, breed llamas, reduce fall damage, and extend campfire smokes.
+type HayBale struct {
+	solid
+
+	// Axis is the axis which the hay bale block faces.
+	Axis cube.Axis
+
+	// Deprecated it has no use, however it must be implemented for it to work.
+	Deprecated int32
+}
+
+// UseOnBlock ...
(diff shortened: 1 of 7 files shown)


DaPigGuy (changes requested):
Hay bales are currently missing their fall damage reduction & note block interaction.

server/block/hay_bale.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Instrument ...
+func (HayBale) Instrument() sound.Instrument {
+	return sound.Banjo()
+}
+
+// EntityLand ...
+func (h HayBale) EntityLand(pos cube.Pos, w *world.World, e world.Entity, distance *float64) {
+	if _, ok := e.(fallDistanceEntity); ok {
+		*distance = *distance * 0.2
```
Sandertv, line 28: Think you can do `*distance *= 0.2` here.

server/block/nether_wart_block.go
```diff
@@ -0,0 +1,30 @@
+package block
+
+// NetherWartBlock is a decorative block found in crimson forests and crafted using Nether wart.
+type NetherWartBlock struct {
```
Sandertv, line 4: Would name this `NetherWart` for consistency with other blocks and rename the file too.

server/block/purpur.go
```diff
@@ -0,0 +1,64 @@
...
+type Purpur struct {
+	solid
+	bassDrum
+
+	// Pillar specifies if the block is the pillar variant or not.
+	Pillar bool
+
+	// Axis is the axis which the purpur pillar block faces.
+	Axis cube.Axis
+}
```
Sandertv, line 20: I would consider splitting this up into `Purpur` and `PurpurPillar`, because a normal purpur block cannot have an Axis. It's a bit counter intuitive to have this available in that case.

Sandertv (commented):
Thanks for the PR and sorry it took so long to get reviewed!

I've got some comments, but your PR looks pretty good! Just some minor changes and it should be ready to go.

Also: I've introduced a conflict with a recent commit, make sure to resolve those conflicts or let us know if you need help.

Sandertv:
As discussed on Discord, I will be resolving the comments on the PR because @mmm545 doesn't currently have access to his PC.

Thanks for the PR!

Issue #391, opened 2022-01-04
Cannot break Dragon Egg in creative


RoyalMCPE:
Attempting to break the Dragon Egg in creative will cause it to teleport instead of breaking. 

Minecraft Version: 1.18.2
OS: LinuxMint 20.1
Go Version: 1.17.3 linux/amd64
DF Version: ba7d3845ae612bd0449c27e357a23657f087ec3d

T14Raptor:
Fixed by 8ee5e6f42600d9916714f254110e413c929b4dae.

PR #392, opened 2022-01-04
this should fix issue #391

RestartFU:
in this issue, you can't break the dragon egg, even in creative


T14Raptor:
Looks good, thanks!

PR #397, opened 2022-01-09
Implement Scoretags

Prim69:
This adds the following methods to Player: SetScoreTag, ScoreTag

image

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -419,7 +419,7 @@ func (p *Player) NameTag() string {
 	return p.nameTag.Load()
 }
 
-// SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the players'
+// SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the player's.
 // name tag.
 func (p *Player) SetScoreTag(a ...interface{}) {
 	p.scoreTag.Store(format(a))
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -421,8 +421,8 @@ func (p *Player) NameTag() string {
 
 // SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the players'
 // name tag.
-func (p *Player) SetScoreTag(scoreTag string) {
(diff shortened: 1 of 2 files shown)


server/player/player.go
```diff
@@ -418,6 +419,18 @@ func (p *Player) NameTag() string {
 	return p.nameTag.Load()
 }
 
+// SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the players'
```
T14Raptor, line 422:
```suggestion
// SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the player's
```

server/player/player.go
```diff
@@ -418,6 +419,18 @@ func (p *Player) NameTag() string {
...
+// name tag.
+func (p *Player) SetScoreTag(scoreTag string) {
+	p.scoreTag.Store(scoreTag)
+	p.updateState()
+}
+
+// ScoreTag returns the current score tag of the Player. It can be changed using SetScoreTag and by default has no value.
```
T14Raptor, line 429:
```suggestion
// ScoreTag returns the current score tag of the player. It can be changed using SetScoreTag and by default is empty.
```

T14Raptor (commented):
Looks good, just two things.

server/player/player.go
```diff
@@ -418,6 +419,18 @@ func (p *Player) NameTag() string {
 	return p.nameTag.Load()
 }
 
+// SetScoreTag changes the score tag displayed over the player in-game. The score tag is displayed under the players'
+// name tag.
+func (p *Player) SetScoreTag(scoreTag string) {
```
Sandertv, line 424: I'd make this accept `a ...interface{}` for easier usage.

server/player/player.go
Prim69: I think the first one is right?

T14Raptor: What do you mean?

Prim69:
Everythings resolved now

Sandertv (approved):

PR #398, opened 2022-01-10
Implement custom item support

TwistedAsylumMC:
Implements custom item support including support for armour, food, off hand and durable items. Also contains optional automatic resource pack building


server/item/category/category.go
```diff
@@ -0,0 +1,61 @@
+package category
```
Sandertv, line 1: Would move this file to `item/category.go` and get rid of the category package.

server/item/component/item.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/item/armour"
+	"github.com/df-mc/dragonfly/server/world"
+	"strings"
+)
+
+// FromItem returns all the components of the given custom item. If the item has no components, the second return
+// value is false.
+func FromItem(it world.CustomItem) (map[string]interface{}, bool) {
```
Sandertv, line 12: Consider moving this somewhere within the internal package, this doesn't look like something we'd want to expose to the user.

server/server.go
```diff
@@ -442,6 +465,19 @@ func (server *Server) finaliseConn(ctx context.Context, conn session.Conn, l Lis
...
+
+	itemComponentEntries := make([]protocol.ItemComponentEntry, len(server.itemComponents))
+	for name, component := range server.itemComponents {
+		itemComponentEntries = append(itemComponentEntries, protocol.ItemComponentEntry{
+			Name: name,
+			Data: component,
+		})
+	}
+	if err := conn.WritePacket(&packet.ItemComponent{Items: itemComponentEntries}); err != nil {
+		_ = l.Disconnect(conn, "Internal server error.")
```
T14Raptor, line 477: Just ignore what's returned instead of pulling a pocketmine and having a random iNterNal SerVer erRoR

server/item/category/category.go
JustTalDevelops: What are your thoughts on instead moving it to the `creative` package? The category is only utilized in creative, anyways.

Sandertv: That sounds fair but I think that caused import cycles too.

AndreasHGK: Both alternatives seem to cause import cycles

Sandertv (commented):
Got a couple of comments! Also, you might want to replace all `interface{}` with `any` for consistency with the rest of the codebase in Go 1.18.

server/internal/item_internal/components.go
```diff
@@ -0,0 +1,84 @@
...
+	identifier, _ := it.EncodeItem()
+	name := strings.Split(identifier, ":")[1]
+
+	builder := NewComponentBuilder(it.Name(), identifier, category)
+
+	if x, ok := it.(item.Armour); ok {
+		builder.AddComponent("minecraft:armor", map[string]interface{}{
+			"protection": int32(x.DefencePoints()),
+		})
+		builder.AddComponent("minecraft:knockback_resistance", map[string]interface{}{
```
Sandertv, line 22: Still a bunch of `interface{}` here instead of `any`, would just do a replace all in dragonfly with interface{} => any.

server/server.go
```diff
@@ -582,10 +615,15 @@ func vec64To32(vec3 mgl64.Vec3) mgl32.Vec3 {
 // itemEntries loads a list of all custom item entries of the server, ready to be sent in the StartGame
 // packet.
 func (server *Server) itemEntries() (entries []protocol.ItemEntry) {
-	for name, rid := range itemRuntimeIDs {
+	for _, it := range world.Items() {
```
Sandertv, line 618: This re-introduces the issue where not all vanilla items are sent to the client causing inventory glitches. Do you reckon we could somehow fix this? Not sure exactly how the custom items get assigned runtime IDs, but if they're just appended we could just start counting from the end of `itemRuntimeIDs`?

JustTalDevelops: The runtime IDs do seem to start right at the end of `itemRuntimeIDs`, so this should be possible.

Sandertv (approved):
I think this looks good, thanks a lot for the work you've put into this PR!

PR #401, opened 2022-01-11
Fix chunk loading after converting a pre-1.13 world to the latest format

JustTalDevelops:


Sandertv (approved):

PR #402, opened 2022-01-13
Implement a way to directly add a resource pack to a server.

JustTalDevelops:


config.toml
```diff
@@ -44,5 +44,8 @@
   Folder = "players"
 
 [Resources]
+  # LoadPacks is true if packs should be loaded by the folder specified below. If this is false, then packs won't be
+  # attempted to be loaded from the folder, however you will still be able to add your own pack through the code.
+  LoadPacks = true
```
Sandertv, line 49: Is this really useful? If you don't want to load packs, just don't add them to the folder?

server/server.go
```diff
@@ -253,6 +255,12 @@ func (server *Server) PlayerProvider(provider player.Provider) {
 	server.playerProvider = provider
 }
 
+// LoadResourcePack loads a resource pack to the server. The pack will eventually be sent to clients who join the
+// server when started.
+func (server *Server) LoadResourcePack(pack *resource.Pack) {
```
Sandertv, line 260: This doesn't actually load them. I'd call this AddResourcePack.

server/server.go
```diff
@@ -653,20 +661,20 @@ func (server *Server) biomes() map[string]interface{} {
 // loadResources loads resource packs from path of specifed directory.
 func (server *Server) loadResources(p string, log internal.Logger) {
 	if _, err := os.Stat(p); os.IsNotExist(err) {
-		_ = os.Mkdir(p, 0777)
+		_ = os.Mkdir(p, os.ModePerm)
```
Sandertv, line 664: This makes it less obvious which permissions are used, I'd just leave this as 0777.

config.toml
JustTalDevelops: I can see it be useful in some cases, such as testing or simply just the fact that an empty resources folder isn't too elegant - especially if your server doesn't plan on using packs anyways. Up to you though.

server/server.go
JustTalDevelops: Changed.

server/server.go
JustTalDevelops: Reverted.

config.toml
Sandertv: I think even with this setting in the majority of the cases the folder will be there anyway because people don't enable this from the start and honestly I think having an empty folder is fine. Feels a bit pointless to have a config setting just to prevent an empty folder.

JustTalDevelops: Removed.

Sandertv (approved):
Looks good, thanks!

PR #404, opened 2022-01-14
Fix golden apple giving regen 2 for 5 minutes instead of 5 seconds

RestartFU:


server/item/golden_apple.go
```diff
@@ -1,9 +1,10 @@
 package item
 
 import (
+	"time"
+
 	"github.com/df-mc/dragonfly/server/entity/effect"
 	"github.com/df-mc/dragonfly/server/world"
```
DaPigGuy, line 7:
```suggestion
	"github.com/df-mc/dragonfly/server/entity/effect"
	"github.com/df-mc/dragonfly/server/world"
	"time"
```

PR #406, opened 2022-01-15
Add a HandleChangeWorld method in the Player handler

Prim69:


server/player/handler.go
```diff
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
+	// HandleChangeWorld handles when the player is added to a new world.
```
Sandertv, line 25: Might be a good idea to document that these worlds may be nil when added to a world for the first time or removed at the end.

Sandertv (commented):
One small documentation improvement to be made, this looks good otherwise!

Prim69:
That's done!

server/player/handler.go
```diff
@@ -22,8 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
-	// HandleChangeWorld handles when the player is added to a new world.
-	HandleChangeWorld(old *world.World, new *world.World)
+	// HandleChangeWorld handles when the player is added to a new world. These values may be nil.
```
T14Raptor, line 25: Pretty sure only the new world could be nil here.

Prim69: What if a world gets deleted?

Prim69: Unless you mean old, that makes sense

T14Raptor: No, I mean new, do you mean deleted as in closed?

T14Raptor: Wait, I mean old, I just confused the shit out of myself

Prim69:
Anything else I should change?

server/player/handler.go
```diff
@@ -22,6 +22,8 @@ type Handler interface {
 	HandleMove(ctx *event.Context, newPos mgl64.Vec3, newYaw, newPitch float64)
 	// HandleTeleport handles the teleportation of a player. ctx.Cancel() may be called to cancel it.
 	HandleTeleport(ctx *event.Context, pos mgl64.Vec3)
+	// HandleChangeWorld handles when the player is added to a new world. oldWorld may be nil.
+	HandleChangeWorld(oldWorld *world.World, newWorld *world.World)
```
Sandertv, line 26: Would go back to calling this just `old` and `new`, the types make it clear these are worlds. Should be good to go after.

Prim69: new conflicts with Go's `new` function

Sandertv: Could name them `before` and `after` then.

Prim69:
Could I get another check on this

Sandertv (approved):

PR #413, opened 2022-01-25
Add Slime Block

max-xoo:
I needed the slime blocks so I implemented them, it's client side everything that is slowing down and jumping

Here is a video to show that it works client side: [click here](https://www.youtube.com/watch?v=DC3cgPGDb5M&ab_channel=MaXoooZ)
(it's on df of course)

### DIFF
--- a/server/block/hash.go
+++ b/server/block/hash.go
@@ -112,6 +112,7 @@ const (
 	hashSeaPickle
 	hashShroomlight
 	hashSign
+	hashSlime
 	hashSnow
 	hashSoulSand
 	hashSoulSoil
@@ -136,7 +137,6 @@ const (
 	hashWoodStairs
 	hashWoodTrapdoor
 	hashWool
-	hashSlime
 )
 
 func (Air) Hash() uint64 {
@@ -575,12 +575,12 @@ func (s Sign) Hash() uint64 {
 	return hashSign | uint64(s.Wood.Uint8())<<8 | uint64(s.Attach.Uint8())<<11
 }
 
-func (Snow) Hash() uint64 {
-	return hashSnow
-}
-
 func (Slime) Hash() uint64 {
 	return hashSlime
+}
+
+func (Snow) Hash() uint64 {
+	return hashSnow
 }
 
 func (SoulSand) Hash() uint64 {
--- a/server/block/hash.go
+++ b/server/block/hash.go
@@ -136,6 +136,7 @@ const (
 	hashWoodStairs
 	hashWoodTrapdoor
 	hashWool
+	hashSlime
 )
 
 func (Air) Hash() uint64 {
@@ -576,6 +577,10 @@ func (s Sign) Hash() uint64 {
 
 func (Snow) Hash() uint64 {
 	return hashSnow
+}
+
+func (Slime) Hash() uint64 {
+	return hashSlime
 }
 
 func (SoulSand) Hash() uint64 {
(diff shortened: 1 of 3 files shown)


server/block/slime.go
```diff
@@ -0,0 +1,21 @@
+package block
+
+// Slime is a full-sized block of snow.
```
JustTalDevelops, line 3: Looks like you forgot to change this?

server/block/hash.go
```diff
@@ -136,6 +136,7 @@ const (
 	hashWoodStairs
 	hashWoodTrapdoor
 	hashWool
+	hashSlime
```
JustTalDevelops, line 139: I would keep this list in alphabetical order - but you should probably be using the blockhash tool for this.

server/block/slime.go
max-xoo: Oh yeah, good point, I'm changing that

server/block/hash.go
max-xoo: Ok I'll change that too

server/block/slime.go
```diff
@@ -0,0 +1,21 @@
+package block
+
+// A Slime block is a translucent block that entities can bounce on
+type Slime struct {
+	solid
+}
+
+// BreakInfo ...
+func (s Slime) BreakInfo() BreakInfo {
+	return newBreakInfo(0.1, alwaysHarvestable, shovelEffective, oneOf(s))
```
mmm545, line 10: i don't think there is any specific tool that's effective for breaking slimes?

max-xoo: With the hand the block is one shot

server/block/slime.go
```diff
@@ -0,0 +1,21 @@
+package block
+
+// A Slime block is a translucent block that entities can bounce on
```
JustTalDevelops, line 3: Just to stay consistent, I'd remove the `A ` prefix and add a period at the end of the sentence.

server/block/slime.go
AndreasHGK: yea, but the it says here the tool to harvest slime blocks is a shovel

server/block/slime.go
max-xoo: I do that

server/block/slime.go
```diff
@@ -0,0 +1,21 @@
...
+type Slime struct {
```
T14Raptor, line 4: This should be called `SlimeBlock` and the file should be renamed to `slime_block.go`

server/block/slime.go
JustTalDevelops: You might as well also remove the `block` suffix after `Slime`.

server/block/slime.go
```diff
@@ -0,0 +1,21 @@
+package block
+
+// Slime is a translucent block that entities can bounce on.
+type Slime struct {
```
T14Raptor, line 4: This also should implement `block.LightDiffuser` with `LightDiffusionLevel()` returning the correct light diffusion level

server/block/slime.go
max-xoo: done

server/block/slime.go
max-xoo: done

server/block/slime.go
T14Raptor, line 4: This also also should implement `block.EntityLander`, and should reset the entity's fall distance if it implements `FallDistanceEntity`

T14Raptor (changes requested):

DaPigGuy:
Closing in favor of #445

PR #417, opened 2022-01-30
Move HandleDeath to before dragonfly handles the death

Prim69:
This allows users to get the contents of the player's inventory when they died, change the drops when a player dies, etc.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -807,7 +807,7 @@ func (p *Player) kill(src damage.Source) {
 	}
 
 	p.addHealth(-p.MaxHealth())
-	
+
 	p.handler().HandleDeath(src)
 	p.StopSneaking()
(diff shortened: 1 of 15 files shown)


server/player/player.go
```diff
@@ -806,6 +806,8 @@ func (p *Player) kill(src damage.Source) {
 		viewer.ViewEntityAction(p, entity.DeathAction{})
 	}
 
+	p.handler().HandleDeath(src)
```
T14Raptor, line 809: I think this should get called after the health is set to zero for the player.

Prim69: Done

T14Raptor:
Run go fmt over this then I can merge

Prim69:
All good now

T14Raptor:
Thanks!

PR #418, opened 2022-02-06
Check for text before sending SetTitle packet

Prim69:
Sometimes SendTitle is used to send just an action bar message or a subtitle, it would be unnecessary to send a SetTitle packet for an empty title


Sandertv (approved):

PR #431, opened 2022-02-22
Add an option to disable ClientCache in config

Prim69:


T14Raptor:
What would be the purpose in having this disabled? I don't see any valid reason.

Prim69:
We discussed on discord but I'm gonna lea e this here for others

The point of this is for proxies using dragonflys server.Listen, typically its off when you dial but when using server.Listen you can't dial

Sandertv:
Right, but that's a gophertunnel issue as opposed to a dragonfly issue. I don't think we should change dragonfly to deal with that.

Sandertv:
Doesn't feel right to add an exception for that though, could the proxy not just implement this cache system? It's not all that difficult. Alternatively, could the proxy not just send that it doesn't support the cache system, like the Nintendo switch? You'd just get the normal chunk packets that way.

RestartFU:
> Doesn't feel right to add an exception for that though, could the proxy not just implement this cache system? It's not all that difficult. Alternatively, could the proxy not just send that it doesn't support the cache system, like the Nintendo switch? You'd just get the normal chunk packets that way.

it is not a proxy, I know prim said so but we are not dialing, we are simpy making a listener.

Sandertv:
Hmmm I still don't exactly see the issue there. `session.Conn` has a `ClientCacheEnabled() bool` method. If you just always return false there, you'd achieve the same result I think?

PR #445, opened 2022-03-11
Add slime blocks

Prim69:


Sandertv:
I actually feel like it might be better to name this `block.Slime`. @T14Raptor I understand we've been adding the `Block` suffix for consistency, but maybe we should remove them from all blocks? The stuttering you get from `block.SlimeBlock` is a bit awkward.

T14Raptor:
I think we should.

server/block/slime_block.go
```diff
@@ -0,0 +1,39 @@
...
+// SlimeBlock is a block that slows down entities and bounces them up if they drop onto it.
+type SlimeBlock struct {
+	solid
+	transparent
+}
+
+// EntityLand ...
+func (SlimeBlock) EntityLand(_ cube.Pos, _ *world.World, e world.Entity) {
+	if fallEntity, ok := e.(fallDistanceEntity); ok {
+		fallEntity.ResetFallDistance()
```
Sandertv, line 17: I'm thinking we should probably do the proper velocity calculations here too, so that it also works for non-session entities.

Prim69:
I added velocity logic for non-player entities, there's just a small issue

https://user-images.githubusercontent.com/13991048/178130940-ce232956-5bd9-4d0d-bdcf-289560e65969.mov

https://user-images.githubusercontent.com/13991048/178130942-e573e10b-ecae-45d7-b60d-01204a81194f.mov

when you throw an item at a horizontal angle, it just slides across (which is not vanilla behavior)

server/block/slime_block.go
```diff
@@ -0,0 +1,45 @@
...
+	solid
+	transparent
+}
+
+// EntityLand ...
+func (SlimeBlock) EntityLand(_ cube.Pos, _ *world.World, e world.Entity) {
+	if e, ok := e.(fallDistanceEntity); ok {
+		e.ResetFallDistance()
+	}
+	if e, ok := e.(*entity.Item); ok {
```
JustTalDevelops, line 20: Wouldn't check for `Item` explicitly, but maybe just a general interface which allows access to Velocity and SetVelocity.

Prim69: oops i thought i reverted that

Prim69: done

server/block/slime_block.go
```diff
@@ -0,0 +1,55 @@
...
+
+// SlimeBlock is a block that slows down entities and bounces them up if they drop onto it.
+type SlimeBlock struct {
+	solid
+	transparent
+}
+
+// velocityEntity is an entity that has velocity.
+type velocityEntity interface {
+	// Velocity returns the current velocity of the entity. The values in the Vec3 returned represent the speed on
```
JustTalDevelops, line 17: You don't need these docs; it's unexported.

Prim69:
here is vanilla behavior btw

https://user-images.githubusercontent.com/13991048/178131056-aec4730c-a1eb-4129-af92-d632fef29dfa.mov

server/block/slime_block.go
```diff
@@ -0,0 +1,53 @@
...
+type velocityEntity interface {
+	// Velocity returns the current velocity of the entity.
+	Velocity() mgl64.Vec3
+	// SetVelocity sets the velocity of the entity.
+	SetVelocity(v mgl64.Vec3)
+}
+
+// EntityLand ...
+func (SlimeBlock) EntityLand(_ cube.Pos, _ *world.World, e world.Entity) {
+	if e, ok := e.(fallDistanceEntity); ok {
```
DaPigGuy, line 25: players that are sneaking will still take fall damage

server/block/slime_block.go
```diff
@@ -0,0 +1,61 @@
...
+// sneakingEntity is an entity that can sneak. This is typically just players.
+type sneakingEntity interface {
+	// Sneaking returns whether the entity is sneaking.
+	Sneaking() bool
+}
+
+// EntityLand ...
+func (SlimeBlock) EntityLand(_ cube.Pos, _ *world.World, e world.Entity) {
+	if e, ok := e.(fallDistanceEntity); ok {
+		if s, ok := e.(sneakingEntity); ok && s.Sneaking() {
```
DaPigGuy, line 32:
```suggestion
		if s, ok := e.(sneakingEntity); !ok || !s.Sneaking() {
```

PR #453, opened 2022-03-18
Implement player spawn

provsalt:


server/player/player.go
```diff
@@ -827,7 +827,11 @@ func (p *Player) Respawn() {
 	case world.End:
 		_, w = w.PortalDestinations()
 	}
-	pos := w.Spawn().Vec3Middle()
+
+	pos := p.spawnPoint.Load()
+	if pos.ApproxEqual(mgl64.Vec3{}) {
+		pos = w.Spawn().Vec3Middle()
+	}
```
Sandertv, line 834: I'm a bit concerned over how this will work when the player gets moved to a different world, do you have any idea on how we can implement this in a way that it doesn't teleport players to a spawn position that is invalid in different worlds?

provsalt:
PR will be stale for awhile because I'm quite busy lately

server/player/player.go
```diff
@@ -805,7 +805,11 @@ func (p *Player) Respawn() {
 	case world.End:
 		_, w = w.PortalDestinations()
 	}
+
 	pos := w.Spawn().Vec3Middle()
+	if playerSpawn := w.PlayerSpawn(p.UUID()); !playerSpawn.ApproxEqual(mgl64.Vec3{}) {
```
T14Raptor, line 810: What if the player’s spawn is actually at 0, 0, 0. This will break that

T14Raptor, line 810: What if the player’s spawn is actually at 0, 0, 0. This will break that

Sandertv (commented):
I think the general API and functionality of this PR looks good! The major current issue though, is that I think we should try to follow vanilla as much as possible. Doing that here means we'll have to find the keys that vanilla uses to store player data in leveldb (iterate over a leveldb database from BDS?) and see how the player spawn storage is implemented. That would also help us in the future.

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	}
+	err = nbt.NewEncoderWithEncoding(buf2, nbt.LittleEndian).Encode(map[string]any{
+		"SpawnX": math.Floor(pos.X()),
+		"SpawnY": math.Floor(pos.Y()),
+		"SpawnZ": math.Floor(pos.Z()),
+	})
+	if err != nil {
+		return err
+	}
+	if err := p.db.Put([]byte("player_server_"+playerData["ServerId"]), buf2.Bytes(), nil); err != nil {
```
Sandertv, line 227: ServerId already includes `player_server_`.

server/world/world.go
```diff
@@ -801,6 +802,27 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+// PlayerSpawn returns the location of the player spawn in the world.
+func (w *World) PlayerSpawn(uuid uuid.UUID) (mgl64.Vec3, bool) {
+	if w == nil {
+		return mgl64.Vec3{}, false
+	}
+	position, err := w.conf.Provider.LoadPlayerSpawnPosition(uuid)
+	if err != nil {
+		return mgl64.Vec3{}, false
+	}
+	return position, true
```
Sandertv, line 814: Can simplify this to `return position, err == nil`.

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
 }
 
+// LoadPlayerSpawnPosition loads the players spawn position stored in the level.dat from their UUID.
+func (p *Provider) LoadPlayerSpawnPosition(uuid uuid.UUID) (pos mgl64.Vec3, err error) {
+	data, err := p.db.Get(append([]byte("player_"), uuid[:]...), nil)
+	if err != nil {
+		return mgl64.Vec3{}, err
+	}
+	buf := bytes.NewBuffer(data)
+	dec := nbt.NewDecoderWithEncoding(buf, nbt.LittleEndian)
```
Sandertv, line 158: Consider using `nbt.UnmarshalWithEncoding` instead. This isn't a stream of data anyway.

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	}
+	serverDb, err := p.db.Get([]byte("player_server_"+playerData["ServerId"]), nil)
+	if err != nil {
+		return mgl64.Vec3{}, err
+	}
+
+	var serverData map[string]any
+	buf2 := bytes.NewBuffer(serverDb)
+	dec2 := nbt.NewDecoderWithEncoding(buf2, nbt.LittleEndian)
+	if err := dec2.Decode(&serverData); err != nil {
```
Sandertv, line 175: Would use `nbt.UnmarshalWithEncoding()` here too.

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	return mgl64.Vec3{x.(float64), y.(float64), z.(float64)}, nil
+}
+
+// SavePlayerSpawnPosition saves the player spawn position passed to the level.dat.
+func (p Provider) SavePlayerSpawnPosition(uuid uuid.UUID, pos mgl64.Vec3) error {
+	data, err := p.db.Get(append([]byte("player_"), uuid[:]...), nil)
+	buf := bytes.NewBuffer(data)
+	if errors.Is(err, leveldb.ErrNotFound) {
+		buf = bytes.NewBuffer(nil)
+		err := nbt.NewEncoderWithEncoding(buf, nbt.LittleEndian).Encode(map[string]string{
```
Sandertv, line 192: Would use `nbt.MarshalWithEncoding()` here.

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+		if err := p.db.Put(append([]byte("player_"), uuid[:]...), buf.Bytes(), nil); err != nil {
+			return err
+		}
+	} else {
+		return err
+	}
+	dec := nbt.NewDecoderWithEncoding(buf, nbt.LittleEndian)
+	var playerData map[string]string
+
+	if err := dec.Decode(&playerData); err != nil {
```
Sandertv, line 208: `nbt.UnmarshalWithEncoding`

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	serverDb, err := p.db.Get([]byte("player_server_"+playerData["ServerId"]), nil)
+	buf2 := bytes.NewBuffer(serverDb)
+	if err != leveldb.ErrNotFound {
+		return err
+	}
+	err = nbt.NewEncoderWithEncoding(buf2, nbt.LittleEndian).Encode(map[string]any{
+		"SpawnX": math.Floor(pos.X()),
+		"SpawnY": math.Floor(pos.Y()),
+		"SpawnZ": math.Floor(pos.Z()),
+	})
```
Sandertv, line 223: `nbt.MarshalWithEncoding`

server/world/mcdb/provider.go
```diff
@@ -145,6 +148,88 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	dec := nbt.NewDecoderWithEncoding(buf, nbt.LittleEndian)
+	var playerData map[string]string
+
+	if err := dec.Decode(&playerData); err != nil {
+		return mgl64.Vec3{}, err
+	}
+	if playerData["MsaID"] != uuid.String() && playerData["ServerId"] == "" {
+		return mgl64.Vec3{}, fmt.Errorf("LoadPlayerSpawn: Invalid data for player %s", uuid.String())
+	}
+	serverDb, err := p.db.Get([]byte("player_server_"+playerData["ServerId"]), nil)
```
Sandertv, line 167: ServerId already includes `player_server_`.

Sandertv (commented):
Looks pretty good! I got a couple of comments. For the nbt writing and reading, I'd recommend using `nbt.Marshal*` and `nbt.Unmarshal*` here, as you're encoding/decoding byte slices, not streams of data. Those two functions are optimised to reduce allocation for frequent writes/reads.

provsalt:
Seems like the implementation works in dragonfly now, however it does not work in bedrock dedicated server. I'm not sure why

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,76 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+		return mgl64.Vec3{}, err
+	}
+
+	var serverData map[string]any
+	if err := nbt.UnmarshalEncoding(serverDb, &serverData, nbt.LittleEndian); err != nil {
+		return mgl64.Vec3{}, err
+	}
+
+	x, y, z := serverData["SpawnX"], serverData["SpawnY"], serverData["SpawnZ"]
+	if x == nil && y == nil && z == nil {
```
DaPigGuy, line 172:
```suggestion
	if x == nil || y == nil || z == nil {
```

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,76 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+			"MsaID":    uuid.String(),
+			"ServerID": "player_server_" + uuid.String(),
+		}, nbt.LittleEndian)
+		if err != nil {
+			return err
+		}
+		if err := p.db.Put(append([]byte("player_"), uuid[:]...), data, nil); err != nil {
+			return err
+		}
+	} else {
```
T14Raptor, line 192:
```suggestion
	} else if err != nil {
```

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,76 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	}
+
+	var serverData map[string]any
+	if err := nbt.UnmarshalEncoding(serverDb, &serverData, nbt.LittleEndian); err != nil {
+		return mgl64.Vec3{}, err
+	}
+
+	x, y, z := serverData["SpawnX"], serverData["SpawnY"], serverData["SpawnZ"]
+	if x == nil || y == nil || z == nil {
+		return mgl64.Vec3{}, fmt.Errorf("LoadPlayerSpawn: player spawn position is non-existant")
```
T14Raptor, line 173:
```suggestion
		return mgl64.Vec3{}, fmt.Errorf("LoadPlayerSpawn: player spawn position is non-existent")
```

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,76 @@ func (p *Provider) SaveSettings(s *world.Settings) {
 	p.saveDifficulty(s.Difficulty)
 }
 
+// LoadPlayerSpawnPosition loads the players spawn position stored in the level.dat from their UUID.
+func (p *Provider) LoadPlayerSpawnPosition(uuid uuid.UUID) (pos mgl64.Vec3, err error) {
```
JustTalDevelops, line 148: Maybe it'd be better to return a bool alongside the error, to specify if the player spawn didn't exist or if there's an actual error in the data.

JustTalDevelops: `(pos mgl64.Vec3, exists bool, err error)`

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,76 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	}
+	var playerData map[string]string
+
+	if err := nbt.UnmarshalEncoding(data, &playerData, nbt.LittleEndian); err != nil {
+		return err
+	}
+	if playerData["MsaID"] != uuid.String() && playerData["ServerId"] == "" {
+		return fmt.Errorf("SavePlayerSpawn: Invalid data for player %s", uuid.String())
+	}
+	coords, err := nbt.MarshalEncoding(map[string]any{
```
Sandertv, line 203: So there's one last thing I think you might consider changing here. Right now, we overwrite all data stored with just the spawn X, spawn Y and spawn Z. Maybe it's an idea to first load the map of data (as done in `LoadPlayerSpawnPosition`), then overwrite only the `SpawnX`, `SpawnY` and `SpawnZ` fields and write the map again. That way we don't lose all the other fields that are still stored there, which means better compatibility with vanilla.

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,85 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	}
+
+	var serverData map[string]any
+	if err := nbt.UnmarshalEncoding(serverDB, &serverData, nbt.LittleEndian); err != nil {
+		return mgl64.Vec3{}, false, err
+	}
+
+	x, y, z := serverData["SpawnX"], serverData["SpawnY"], serverData["SpawnZ"]
+	if x == nil || y == nil || z == nil {
+		return mgl64.Vec3{}, false, fmt.Errorf("LoadPlayerSpawn: player spawn position is non-existent")
```
JustTalDevelops, line 173: Can just return nil for the error here.

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,85 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	if err != nil {
+		return mgl64.Vec3{}, false, err
+	}
+	var playerData map[string]string
+
+	if err := nbt.UnmarshalEncoding(data, &playerData, nbt.LittleEndian); err != nil {
+		return mgl64.Vec3{}, false, err
+	}
+	if playerData["MsaID"] != uuid.String() && playerData["ServerId"] == "" {
+		return mgl64.Vec3{}, false, fmt.Errorf("LoadPlayerSpawn: Invalid data for player %s", uuid.String())
```
JustTalDevelops, line 159: Would make this lowercase to be consistent with Go convention. (e.g. `load player spawn: invalid data for player: %v`)

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,85 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	} else if err != nil {
+		return err
+	}
+	var playerData map[string]string
+
+	if err := nbt.UnmarshalEncoding(data, &playerData, nbt.LittleEndian); err != nil {
+		return err
+	}
+	if playerData["MsaID"] != uuid.String() && playerData["ServerId"] == "" {
+		return fmt.Errorf("SavePlayerSpawn: Invalid data for player %s", uuid.String())
```
JustTalDevelops, line 201: Same here (e.g. `save player spawn: invalid data for player: %v`)

server/world/world.go
```diff
@@ -841,6 +842,24 @@ func (w *World) SetSpawn(pos cube.Pos) {
 	}
 }
 
+// PlayerSpawn returns the location of the player spawn in the world.
+func (w *World) PlayerSpawn(uuid uuid.UUID) (mgl64.Vec3, bool, error) {
+	if w == nil {
+		return mgl64.Vec3{}, false, nil
+	}
+	position, exist, err := w.prov.Load().LoadPlayerSpawnPosition(uuid)
```
JustTalDevelops, line 850: Just return this directly?

server/world/world.go
```diff
@@ -841,6 +842,24 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+	return position, exist, err
```
DaPigGuy, line 851:
```suggestion
	return w.prov.Load().LoadPlayerSpawnPosition(uuid)
```

server/world/world.go
```diff
@@ -841,6 +842,24 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+}
+
+// SetPlayerSpawn sets the spawn of the player. If The player has a player spawn in the world, the player will
```
DaPigGuy, line 854:
```suggestion
// SetPlayerSpawn sets the spawn of the player. If the player has a player spawn in the world, the player will
```

server/world/provider.go
```diff
@@ -52,4 +59,8 @@ func (NopProvider) LoadBlockNBT(ChunkPos) ([]map[string]any, error) { return nil
 func (NopProvider) SaveBlockNBT(ChunkPos, []map[string]any) error   { return nil }
 func (NopProvider) SaveChunk(ChunkPos, *chunk.Chunk) error          { return nil }
 func (NopProvider) LoadChunk(ChunkPos) (*chunk.Chunk, bool, error)  { return nil, false, nil }
-func (NopProvider) Close() error                                    { return nil }
+func (NopProvider) LoadPlayerSpawnPosition(uuid.UUID) (mgl64.Vec3, bool, error) {
```
JustTalDevelops, line 62: Would inline these if possible.

server/world/provider.go
```diff
@@ -14,6 +16,11 @@ type Provider interface {
 	// SaveSettings saves the settings of a World.
 	SaveSettings(*Settings)
 
+	// LoadPlayerSpawnPosition loads the player spawn point if found, otherwise an error will be returned.
+	LoadPlayerSpawnPosition(uuid uuid.UUID) (pos mgl64.Vec3, exists bool, err error)
+	// SavePlayerSpawnPosition saves the player spawn point, in vanilla, this can be done with beds in the overworld
+	// and respawn anchors in the nether.
```
DaPigGuy, line 22:
```suggestion
	// SavePlayerSpawnPosition saves the player spawn point. In vanilla, this can be done with beds in the overworld
	// or respawn anchors in the nether.
```

server/world/mcdb/provider.go
```diff
@@ -141,6 +144,85 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+	data, err := p.db.Get(append([]byte("player_"), uuid[:]...), nil)
+	if err != nil {
+		return mgl64.Vec3{}, false, err
+	}
+	var playerData map[string]string
+
+	if err := nbt.UnmarshalEncoding(data, &playerData, nbt.LittleEndian); err != nil {
+		return mgl64.Vec3{}, false, err
+	}
+	if playerData["MsaID"] != uuid.String() && playerData["ServerId"] == "" {
```
DaPigGuy, line 158: Is this not supposed to be `or`?

DaPigGuy (approved):

JustTalDevelops (approved):

server/world/world.go
```diff
@@ -801,6 +802,23 @@ func (w *World) SetSpawn(pos cube.Pos) {
 	}
 }
 
+// PlayerSpawn returns the location of the player spawn in the world.
+func (w *World) PlayerSpawn(uuid uuid.UUID) (mgl64.Vec3, bool, error) {
```
Sandertv, line 806: For consistency with other methods in World and ease of use, I would change this signature to `PlayerSpawn(uuid.UUID) mgl64.Vec3`. If an error is returned, log that to the World's logger. If an error occurs or if the player spawn was not found, just return the world's spawn point.

server/world/world.go
```diff
@@ -801,6 +802,23 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+	if w == nil {
+		return mgl64.Vec3{}, false, nil
+	}
+	return w.conf.Provider.LoadPlayerSpawnPosition(uuid)
+}
+
+// SetPlayerSpawn sets the spawn of the player. If the player has a player spawn in the world, the player will
+// be teleported to this location on respawn.
+func (w *World) SetPlayerSpawn(uuid uuid.UUID, pos mgl64.Vec3) error {
```
Sandertv, line 815: Would also not return any error here and just log that to the World's logger.

Sandertv (commented):
Last comments, ready to be merged after this!

Sandertv:
Looking at how vanilla stores this again, they store TAG_Ints meaning all `mgl64.Vec3` should actually probably be a `cube.Pos`. The type of these fields is an `int32`. Make sure to cast these properly when you encode to NBT and decode it.

server/world/mcdb/provider.go
```diff
@@ -145,6 +147,83 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+
+	var serverData map[string]any
+	if err := nbt.UnmarshalEncoding(serverDB, &serverData, nbt.LittleEndian); err != nil {
+		return cube.Pos{}, false, err
+	}
+	x, y, z := serverData["SpawnX"], serverData["SpawnY"], serverData["SpawnZ"]
+	if x == nil || y == nil || z == nil {
+		return cube.Pos{}, false, nil
+	}
+	return cube.Pos{x.(int), y.(int), z.(int)}, true, nil
```
Sandertv, line 177: These are of the type `int32`, so you might want to change them to `int(x.(int32))`.

server/world/mcdb/provider.go
```diff
@@ -145,6 +147,83 @@ func (p *Provider) SaveSettings(s *world.Settings) {
...
+		return err
+	}
+
+	var serverData map[string]any
+	if err := nbt.UnmarshalEncoding(serverDB, &serverData, nbt.LittleEndian); err != nil {
+		return err
+	}
+	serverData["SpawnX"] = pos.X()
+	serverData["SpawnY"] = pos.Y()
+	serverData["SpawnZ"] = pos.Z()
```
Sandertv, line 216: These should be of the type `int32`, so make sure to cast them like `int32(pos.X())` beforehand.

server/world/world.go
```diff
@@ -801,6 +802,30 @@ func (w *World) SetSpawn(pos cube.Pos) {
 	}
 }
 
+// PlayerSpawn returns the location of the player spawn in the world.
+func (w *World) PlayerSpawn(uuid uuid.UUID) (pos cube.Pos) {
```
Sandertv, line 806: You can remove the named return here, just `cube.Pos`.

server/world/world.go
```diff
@@ -801,6 +802,30 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+	return pos
+}
+
+// SetPlayerSpawn sets the spawn of the player. If the player has a player spawn in the world, the player will
+// be teleported to this location on respawn.
+func (w *World) SetPlayerSpawn(uuid uuid.UUID, pos cube.Pos) {
+	if w == nil {
+		return
+	}
+	w.conf.Log.Debugf("failed to set player spawn: %v", w.conf.Provider.SavePlayerSpawnPosition(uuid, pos))
```
Sandertv, line 826: This will always debug this message. I would first call SavePlayerSpawnPosition and only log the error if the error is non-nil. Also, I'd recommend using `Log.Errorf` for this!

server/world/world.go
```diff
@@ -801,6 +802,30 @@ func (w *World) SetSpawn(pos cube.Pos) {
...
+func (w *World) PlayerSpawn(uuid uuid.UUID) cube.Pos {
+	if w == nil {
+		return cube.Pos{}
+	}
+	pos, exist, err := w.conf.Provider.LoadPlayerSpawnPosition(uuid)
+	if !exist {
+		pos = w.Spawn()
+	}
+	if err != nil {
+		w.conf.Log.Debugf("failed to get player spawn: %v", err)
```
Sandertv, line 815: Would use `Errorf` here as well!

Sandertv:
Thanks a lot for the PR! I'll make some minor changes and merge it into master!

PR #456, opened 2022-03-23
world/world.go: SetBlock: Added an Opts struct.

### LINKED ISSUE #281: Merge SetBlock, BreakBlock and PlaceBlock into a single SetBlock
These three methods have similar behaviour and only very slightly different implications. Instead of these three methods, we could have some `*Opts` struct passed to SetBlock with an option to perform block updates and/or liquid displacement as currently carried out by BreakBlock and PlaceBlock. The replacement of `block.Air` with `nil` (as proposed in #280) makes it possible to detect when a block is 'broken'.

In addition to that, I propose we get rid of the particles produced by BreakBlock so we can get rid of the internal proxy variables as mentioned by @T14Raptor.

T14Raptor:
Would probably be nice to have a function for adding block breaking particles to the world in the block package, will do soon

Closes #281


server/world/world.go
```diff
@@ -234,33 +234,69 @@ func (w *World) highestObstructingBlock(x, z int) int {
 	return w.ra[0]
 }
 
+// Opts ...
+type Opts struct {
```
Sandertv, line 238: Would maybe rename this to `SetOpts` in case we ever add `Opts` of a different kind.

Sandertv (commented):
Looks pretty good. Let me know when this is ready for me to document.

T14Raptor:
Ready for documentation.

PR #458, opened 2022-03-24
Implemented knock back enchantment.

JustTalDevelops:


server/player/player.go
```diff
@@ -1446,6 +1447,12 @@ func (p *Player) AttackEntity(e world.Entity) {
 		}
 		if vulnerable {
 			p.Exhaust(0.1)
+
+			if s, ok := i.Enchantment(enchantment.KnockBack{}); ok {
+				diff := float64(s.Level()) / 2
+				force += diff
+				height += diff
+			}
```
Sandertv, line 1455: Would maybe move these calculations into a method the enchantment itself.

JustTalDevelops: Changed.

Sandertv (commented):
Is this tested to match vanilla?

JustTalDevelops:
Yes.

PR #466, opened 2022-03-31
Add configuration to force resource packs onto players

provsalt:


server/config.go
```diff
@@ -50,6 +50,9 @@ type Config struct {
 	Resources struct {
 		// Folder controls the location where resource packs will be loaded from.
 		Folder string
+
```
Sandertv, line 53: Would remove the space here for consistency.

Sandertv (commented):
Just two small nitpicks for consistency, looks good otherwise!

server/config.go
```diff
@@ -50,6 +50,8 @@ type Config struct {
 	Resources struct {
 		// Folder controls the location where resource packs will be loaded from.
 		Folder string
+		// Required is a boolean to force the client to load the resource pack on join. If they do not accept, they'll have to leave the server
```
Sandertv, line 53: You forgot a `.` at the end here.

Sandertv (approved):
Looks good! Thank you for the PR!

PR #468, opened 2022-04-02
player/handler.go: Added a HandleItemConsume function.

JustTalDevelops:


Sandertv (approved):
Seems like a reasonable addition! Looks good, thanks!

PR #471, opened 2022-04-03
Switch to the new sub-chunk request system

JustTalDevelops:
The sub-chunk request system is a new chunk-system introduced in v1.18.0 (improved upon in v1.18.10) that allows the client to request only the sub-chunks it sees from the server, saving a considerable amount of bandwidth.

This PR also includes a function to get/calculate the heightmap of a chunk, since the sub-chunk specific heightmap is now provided to the client.


JustTalDevelops:
Could I get a review on this? @T14Raptor @Sandertv

server/world/viewer.go
```diff
@@ -25,9 +25,10 @@ type Viewer interface {
 	// ViewEntityTeleport views the teleportation of an entity. The entity is immediately moved to a different
 	// target position.
 	ViewEntityTeleport(e Entity, position mgl64.Vec3)
-	// ViewChunk views the chunk passed at a particular position. It is called for every chunk loaded using
-	// the world.Loader.
-	ViewChunk(pos ChunkPos, c *chunk.Chunk, blockNBT map[cube.Pos]Block)
+	// ViewSubChunks views the sub-chunks requested in the offsets passed.
+	ViewSubChunks(center SubChunkPos, offsets [][3]int8)
```
Sandertv, line 29: I don't think this needs to be added to the Viewer interface. It is only called within the session package and is more of an implementation detail.

server/world/viewer.go
```diff
@@ -25,9 +25,10 @@ type Viewer interface {
...
+	// ViewSkeletonChunk views the skeleton chunk passed. It is called before the viewer requests sub-chunks.
+	ViewSkeletonChunk(pos ChunkPos, c *chunk.Chunk)
```
Sandertv, line 31: I'd name this `ViewChunk`, the fact that it is sent as a skeleton chunk is an implementation detail specific to a client Viewer.

server/world/chunk/light_area.go
```diff
@@ -132,13 +132,19 @@ func (a *lightArea) iterEdges(filter func(a, b *SubChunk) bool, f func(a, b cube
 	}
 }
 
-// iterHeightmap iterates over the heightmap of the lightArea and calls the function f with the heightmap value, the
-// heightmap value of the highest neighbour and the Y value of the highest non-empty SubChunk.
+// iterHeightmap iterates over the HeightMap of the lightArea and calls the function f with the HeightMap value, the
+// HeightMap value of the highest neighbour and the Y value of the highest non-empty SubChunk.
```
Sandertv, line 136: Would revert these heightmap -> HeightMap changes as it doesn't refer to the HeightMap method.

server/world/chunk/chunk.go
```diff
@@ -110,20 +123,33 @@ func (chunk *Chunk) HighestLightBlocker(x, z uint8) int16 {
...
-					return int16(y) | chunk.subY(index)
+					return int16(y) | chunk.SubY(index)
 				}
 			}
 		}
 	}
 	return int16(chunk.r[0])
 }
 
+// HeightMap returns the height map of the chunk.
```
Sandertv, line 140: Would also document that this method recalculates it if the chunk was edited.

server/session/world.go
```diff
@@ -22,90 +22,137 @@ import (
...
+	defer s.blobMu.Unlock()
 
-	enc := nbt.NewEncoderWithEncoding(s.chunkBuf, nbt.NetworkLittleEndian)
-	for bp, b := range blockEntities {
-		if n, ok := b.(world.NBTer); ok {
-			d := n.EncodeNBT()
-			d["x"], d["y"], d["z"] = int32(bp[0]), int32(bp[1]), int32(bp[2])
-			_ = enc.Encode(d)
-		}
+	s.openChunkTransactions = append(s.openChunkTransactions, map[uint64]struct{}{hash: {}})
```
Sandertv, line 147: I don't think this will work as intended. This system expects a map[uint64]struct{} with all hashes of every sub-chunk in a chunk. Every chunk load tick, `8 - len(s.openChunkTransactions)` chunks are loaded, which means that with this change a single chunk could prevent any other chunks from being sent because the chunk has more than 8 subchunks.

Sandertv (commented):
Got a couple of comments.

server/session/world.go
JustTalDevelops: Hmm, what do you suggest I do instead?

Sandertv: I would open a single transaction containing a map[uint64]struct{} with the hashes that came from all the `Offset` requested together.

Sandertv: Essentially this needs to be an indication of how many chunks are currently in the process of being loaded client-side. If this number goes up too high, we need to slow down chunk sending.

server/world/viewer.go
JustTalDevelops: Should I also re-add the block NBT parameter, ignoring it in the session implementation?

Sandertv: Nah you don't have to do that.

Sandertv (approved):
I think this looks good! Thanks for the PR!

PR #478, opened 2022-04-23
Filled maps (non-persistent, not rendered from terrain)

Endermanbugzjfc:
# Items
- [X] Filled map variants
# Packets
- [x] Map info request handler
- [X] Broadcast pixel changes with offset
- [x] Broadcast decoration changes
# APIs
- [X] API to ~~register~~ create map data
- [X] API to update map data


Endermanbugzjfc:
Compiler exit with code 2??

Endermanbugzjfc:
https://discord.com/channels/623638955262345216/623917901409419274/967355750089371668

Endermanbugzjfc:
I am not sure how I should indicate a map as persisted. Using map ID `0`? Map ID is an `int64`.

Endermanbugzjfc:
Change `MapDataViewer` to `MapViewer`?

Endermanbugzjfc:
> I am not sure how I should indicate a map as persisted. Using map ID `0`? Map ID is an `int64`. It is not as easy as adding another field. Since this requires saving another unnecessary and non-official NBT tag in disk.

I am going to use the `map_is_init` NBT field for this. Map data provider will only give access to player whose map has `true` for that field.

Endermanbugzjfc:
> > I am not sure how I should indicate a map as persisted. Using map ID `0`? Map ID is an `int64`. It is not as easy as adding another field. Since this requires saving another unnecessary and non-official NBT tag in disk.
> 
> I am going to use the `map_is_init` NBT field for this. Map data provider will only give access to player whose map has `true` for that field.

I decide to not mess with persistence in this PR.

server/entity/direction.go
```diff
@@ -27,6 +28,22 @@ func Facing(e world.Entity) cube.Direction {
 	return 0
 }
 
+// Bearing splits the 360 degree to sections and returns the section number by yaw.
+func Bearing(yaw float64, max int64) int64 {
```
JustTalDevelops, line 32: This isn't used anywhere.

Endermanbugzjfc: > This isn't used anywhere.

This was for getting the rotation of decoration. But Sofe suggested me to use another approach.

Endermanbugzjfc: Function removed.

Endermanbugzjfc:
Locked map item should be added. But I can not find the item ID on Fandom.

PR #482, opened 2022-05-01
player/player.go: fix bug when respawning

AndreasHGK:
falling into the void and respawning keeps a player's fall distance, thus making them die once more of fall damage


Sandertv (approved):

Sandertv:
Thanks for the PR!

PR #486, opened 2022-05-04
Make AlwaysConsumable items display their animation when a player is fully saturated

AndreasHGK:


Sandertv (approved):

PR #506, opened 2022-05-16
Implement explosions

provsalt:


server/block/ancient_debris.go
```diff
@@ -16,6 +16,16 @@ func (a AncientDebris) BreakInfo() BreakInfo {
...
 
+// Resistance ...
+func (a AncientDebris) Resistance() float64 {
+	return 1200
+}
+
+// AlwaysExplodeDrop ..
+func (a AncientDebris) AlwaysExplodeDrop() bool {
+	return false
+}
```
Sandertv, line 27: I would add this info to the existing BreakInfo!

server/world/explosion/source.go
```diff
@@ -0,0 +1,35 @@
...
+import "github.com/df-mc/dragonfly/server/world"
+
+// Source is the source of the explosion. This can range from creepers, TNT, and other sources.
+type Source interface {
+	// ExplodeInfo returns the explosion information for the source.
+	ExplodeInfo() *Explosion
+}
+
+// TNTSource is an explosion source that is caused by TNT.
+type TNTSource struct {
```
JustTalDevelops, line 12: I wouldn't implement these here but rather in the blocks or entities themselves.

server/world/explosion/source.go
```diff
@@ -0,0 +1,35 @@
+package explosion
+
+import "github.com/df-mc/dragonfly/server/world"
+
+// Source is the source of the explosion. This can range from creepers, TNT, and other sources.
+type Source interface {
+	// ExplodeInfo returns the explosion information for the source.
+	ExplodeInfo() *Explosion
```
JustTalDevelops, line 8: I don't think this should return an *Explosion, but maybe something similar to BreakInfo

PR #507, opened 2022-05-16
trace/trace.go: fix infinite loop when startPos == endPos

AndreasHGK:


server/block/cube/trace/trace.go
```diff
@@ -10,10 +10,11 @@ import (
 // A function 'f' is passed which is called for each voxel, if f returns false, the function will return.
 // TraverseBlocks panics if the start and end positions are the same.
 func TraverseBlocks(start, end mgl64.Vec3, f func(pos cube.Pos) (con bool)) {
-	dir := end.Sub(start).Normalize()
+	dir := end.Sub(start)
 	if dir.LenSqr() <= 0.0 {
```
T14Raptor, line 14: Might be worth using `mgl64.FloatEqual` here

PR #517, opened 2022-05-28
Implement 1.19 support

JustTalDevelops:


server/player/player.go
```diff
@@ -246,6 +246,12 @@ func (p *Player) SendJukeboxPopup(a ...any) {
 	p.session().SendJukeboxPopup(format(a))
 }
 
+// SendToast sends a toast to the player. This toast is shown at the top of the screen, similar to achievements or pack
+// loading.
+func (p *Player) SendToast(title, message string) {
```
Sandertv, line 251: Do you reckon we could come up with a nicer name than "toast"? It's such a shit name lmao.

provsalt: > Do you reckon we could come up with a nicer name than "toast"? It's such a shit name lmao.

Perhaps alert?

server/session/handler_player_action.go
```diff
@@ -51,6 +51,8 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
 			return nil
 		}
 		s.c.ContinueBreaking(cube.Face(face))
+	case protocol.PlayerActionStartItemUseOn, protocol.PlayerActionStopItemUseOn:
+		// TODO: Properly utilize these actions.
```
T14Raptor, line 55: Do we have to?

server/session/session.go
```diff
@@ -439,6 +439,7 @@ func (s *Session) registerHandlers() {
 		packet.IDPlayerAction:          &PlayerActionHandler{},
 		packet.IDPlayerAuthInput:       &PlayerAuthInputHandler{},
 		packet.IDPlayerSkin:            &PlayerSkinHandler{},
+		packet.IDRequestAbility:        nil, // TODO: Properly support this.
```
Sandertv, line 442: Think you can remove the TODO for this, we don't really need to handle this at any point in the near future.

Sandertv (commented):
Should be good to go once 1.19 is released!

server/session/handler_player_action.go
JustTalDevelops: It would be ideal in my opinion. @Sandertv thoughts?

server/session/session.go
JustTalDevelops: Removed.

PR #519, opened 2022-05-31
Allow for a more visually accurate progress bar.

### LINKED ISSUE #518: Allow more precise XP progress values

T14Raptor:
Resolves #518.

### DIFF
--- a/server/entity/experience.go
+++ b/server/entity/experience.go
@@ -11,6 +11,7 @@ import (
 type ExperienceManager struct {
 	mu         sync.RWMutex
 	experience int
+	d          float64
 }
 
 // NewExperienceManager returns a new ExperienceManager with no experience.
@@ -30,7 +31,7 @@ func (e *ExperienceManager) Add(amount int) (level int, progress float64) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 	e.experience += amount
-	return progressFromExperience(e.experience)
+	return progressFromExperience(e.total())
 }
 
 // Remove removes experience from the total experience and recalculates the level and progress if necessary.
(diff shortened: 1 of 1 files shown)


server/entity/experience.go
```diff
@@ -108,21 +116,21 @@ func progressFromExperience(experience int) (level int, progress float64) {
...
-		return int(float64(level*level)*2.5 - 40.5*float64(level) + 360)
+		return float64(level*level)*2.5 - 40.5*float64(level) + 360
 	}
-	return int(float64(level*level)*4.5 - 162.5*float64(level) + 2220)
+	return float64(level*level)*4.5 - 162.5*float64(level) + 2220
 }
 
 // experienceForLevel returns the amount experience needed to reach level + 1.
-func experienceForLevel(level int) int {
+func experienceForLevel(level int) float64 {
```
Sandertv, line 129: There doesn't seem any reason to make this a float64.

server/entity/experience.go
```diff
@@ -108,13 +116,13 @@ func progressFromExperience(experience int) (level int, progress float64) {
 }
 
 // experienceForLevels calculates the amount of experience needed in total to reach a certain level.
-func experienceForLevels(level int) int {
+func experienceForLevels(level int) float64 {
```
Sandertv, line 119: Would probably go back to returning an int here too. The less we return floats in these functions the better I think.

T14Raptor: Might lose some experience points by truncating here

Sandertv: Technically though experience can only be full numbers?

Sandertv (approved):
Looks good, thanks!

PR #521, opened 2022-05-31
Implemented functional anvils

### LINKED ISSUE #500: Implement anvils

T14Raptor:
Closes #500.


JustTalDevelops:
This is missing updated repair costs on result items and documentation, but aside from that, this is pretty much ready for review.

server/session/handler_item_stack_request.go
```diff
@@ -479,6 +483,190 @@ func (h *ItemStackRequestHandler) handleMineBlock(a *protocol.MineBlockStackRequ
...
+					rarityCost := t.Rarity().ApplyCost
+					if enchant {
+						rarityCost = max(1, rarityCost/2)
+					}
+					cost += rarityCost * resultLevel
+					if first.Count() > 1 {
+						cost = 40
+					}
+				} else {
+					return nil
```
DaPigGuy, line 583: This doesn't seem quite right. Anvil use should only fail if both **a)** not repairing, and **b)** no enchantments from the second item are compatible

main.go
```diff
@@ -28,7 +31,14 @@ func main() {
...
 
-	for srv.Accept(nil) {
+	for srv.Accept(func(p *player.Player) {
+		go func() {
+			for {
+				p.SendTip(entity.Facing(p))
+				time.Sleep(200 * time.Millisecond)
+			}
+		}()
+	}) {
```
DaPigGuy, line 41: Don't forget to remove debug code

DaPigGuy (changes requested):

server/session/handler_item_stack_request.go
JustTalDevelops: This condition should never be met if repairing though?

JustTalDevelops: The second point is valid though

DaPigGuy: > This condition should never be met if repairing though?
It can if repairing with another damaged item

JustTalDevelops: I think this is resolved with the latest commit, let me know if not.

main.go
JustTalDevelops: Removed.

DaPigGuy (changes requested):
Bedrock behavior differs from Java Edition and allows combining two items with incompatible enchantments for repairs

DaPigGuy (changes requested):

server/block/anvil.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/df-mc/dragonfly/server/world/sound"
+	"github.com/go-gl/mathgl/mgl64"
+)
+
+// Anvil is a block that allows players to repair items, rename items, and combine enchantments.
+type Anvil struct {
+	gravityAffected
+	solid
```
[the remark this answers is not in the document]
JustTalDevelops: Added.

server/block/cobblestone.go
```diff
@@ -15,6 +15,11 @@ func (c Cobblestone) BreakInfo() BreakInfo {
 	return newBreakInfo(2, pickaxeHarvestable, pickaxeEffective, oneOf(c))
 }
 
+// Cobblestone ...
+func (c Cobblestone) Cobblestone() bool {
+	return true
```
DaPigGuy, line 20:
```suggestion
	return !c.Mossy
```

server/entity/damage/source.go
```diff
@@ -64,6 +64,12 @@ type (
 		// respectively.
 		Projectile, Owner world.Entity
 	}
+
+	// SourceDamagingBlock is used for damage caused by a damaging block, such as an anvil.
+	SourceDamagingBlock struct {
```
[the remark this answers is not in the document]
JustTalDevelops: Renamed, and removed `SourceCactus`.

server/entity/falling_block.go
```diff
@@ -57,13 +90,37 @@ func (f *FallingBlock) Tick(w *world.World, _ int64) {
...
 
 	m.Send()
-	pos := cube.PosFromVec3(m.pos)
 
+	distThisTick := f.vel.Y()
+	if distThisTick < f.fallDistance.Load() {
+		f.fallDistance.Sub(distThisTick)
+	} else {
+		f.fallDistance.Store(0)
+	}
```
Sandertv, line 99: I have some difficulty making sense of what happens here. I assume the fall distance is supposed to be reset if the falling block has upwards velocity, but what exactly is the `distThisTick < f.fallDistance.Load()` check supposed to do in terms of behaviour?

server/item/enchantment.go
```diff
@@ -38,9 +39,14 @@ type EnchantmentType interface {
...
 	// MaxLevel returns the maximum level the enchantment should be able to have.
 	MaxLevel() int
-	// CompatibleWith is called when an enchantment is added to an item. It can be used to check if
-	// the enchantment is compatible with the item stack based on the item type, current enchantments etc.
-	CompatibleWith(s Stack) bool
+	// Rarity returns the enchantment's rarity.
+	Rarity() EnchantmentRarity
+	// CompatibleWithOther is called when an enchantment is added to an item. It can be used to check if
+	// the enchantment is compatible with other enchantments.
+	CompatibleWithOther(t EnchantmentType) bool
```
Sandertv, line 46: Would probably name this `CompatibleWithEnchantment` so it's more descriptive.

server/item/enchantment_rarity.go
```diff
@@ -0,0 +1,18 @@
...
+
+var (
+	// EnchantmentRarityCommon is the common enchantment rarity.
+	EnchantmentRarityCommon = EnchantmentRarity{Name: "Common", ApplyCost: 1}
+	// EnchantmentRarityUncommon is the uncommon enchantment rarity.
+	EnchantmentRarityUncommon = EnchantmentRarity{Name: "Uncommon", ApplyCost: 2}
+	// EnchantmentRarityRare is the rare enchantment rarity.
+	EnchantmentRarityRare = EnchantmentRarity{Name: "Rare", ApplyCost: 4}
+	// EnchantmentRarityVeryRare is the very rare enchantment rarity.
+	EnchantmentRarityVeryRare = EnchantmentRarity{Name: "Very Rare", ApplyCost: 8}
```
Sandertv, line 11: Might be worth making `EnchantmentRarity` an interface and having unexported implementations for each rarity, then declaring the variables like `var EnchantmentRarityCommon enchantmentRarityCommon`. That way users cannot edit these variables.

server/item/item.go
```diff
@@ -192,6 +194,18 @@ func (d defaultFood) ConsumeDuration() time.Duration {
 	return DefaultConsumeDuration
 }
 
+// DisplayName returns the display name of the item as shown in game in the language passed.
```
Sandertv, line 197: Might be worth documenting that `DisplayName` panics if an unknown (unregistered) item is passed.

server/item/stack.go
```diff
@@ -264,6 +266,18 @@ func (s Stack) Enchantments() []Enchantment {
...
+// renaming this item with an anvil.
+func (s Stack) RepairCost() int {
+	return s.repairCost
+}
+
+// WithRepairCost returns the current Stack with the repair cost set to the passed value.
+func (s Stack) WithRepairCost(repairCost int) Stack {
+	s.repairCost = repairCost
+	return s
+}
```
Sandertv, line 279: I would check if the underlying item is `Repairable`, similar to what `Stack.Damage()` does, to avoid adding repair costs to things that can't have one.

server/item/tool.go
```diff
@@ -84,3 +86,32 @@ func (n ToolNone) HarvestLevel() int { return 0 }
...
+// toolTierRepairable returns true if the ToolTier passed is repairable.
+func toolTierRepairable(tier ToolTier) func(Stack) bool {
+	return func(stack Stack) bool {
+		switch tier {
+		case ToolTierWood:
+			if planks, ok := stack.Item().(interface{ Planks() bool }); ok {
+				return planks.Planks()
+			}
+		case ToolTierStone:
+			if cobblestone, ok := stack.Item().(interface{ Cobblestone() bool }); ok {
```
Sandertv, line 99: I'd consider naming these methods differently, something to imply that they can repair wood tools. Maybe something like `RepairsWoodTools()` or something like that.

server/session/handler_anvil.go
```diff
@@ -0,0 +1,224 @@
...
+		Slot:        1,
+	}, s)
+	if first.Empty() {
+		return fmt.Errorf("no item in first input slot")
+	}
+	result := first
+
+	second, _ := h.itemInSlot(protocol.StackRequestSlotInfo{
+		ContainerID: containerAnvilMaterial,
+		Slot:        0x2,
```
Sandertv, line 41: You use a decimal `1` above, but a hexadecimal `0x2` here. Might want to make those two consistent just for the sake of it.

server/session/handler_anvil.go
```diff
@@ -0,0 +1,224 @@
...
+			ContainerID: containerAnvilMaterial,
+			Slot:        2,
+		}, item.Stack{}, s)
+	}
+	h.setItemInSlot(protocol.StackRequestSlotInfo{
+		ContainerID: containerOutput,
+		Slot:        50,
+	}, result, s)
+	return nil
+}
```
Sandertv, line 208: This is a massive function with a ton of logic in it. Consider splitting it up and particularly adding documentation inside of the code to explain what is going on. It's a really big wall of hard to comprehend logic for anyone else looking at it.

Sandertv (commented):
Thanks a lot for the PR and apologies for taking so long to review.

I've got a couple of comments, particularly on the non-logic related stuff. The actual anvil logic is really quite difficult to read, so (as mentioned in one of the comments) I'd really recommend adding comments throughout the code explaining what's happening.

server/entity/falling_block.go
JustTalDevelops: It's the same behaviour copied from the player package.

Sandertv: Are you able to make sense of it? Do we have any kind of reference to what this is based on?

server/session/handler_anvil.go
JustTalDevelops: Documented a lot of the code, just need to split it up into smaller functions. Will attempt this later today.

server/entity/falling_block.go
JustTalDevelops: See https://github.com/df-mc/dragonfly/blob/17292335c97dc66c42b4b28fba7ea08a5e105b28/server/player/player.go#L476. When the block is falling, the velocity will be negative which will be under the fall distance since it starts at zero. Otherwise, the block isn't falling so the fall distance should be zero.

Sandertv: I get that part, but when the fall distance is, say, 60 blocks, you would need a velocity of 60 upwards to cancel out the fall damage if I understand this correctly, which... doesn't seem sensible?

JustTalDevelops: Resolving after discussion on Discord.

server/item/enchantment_rarity.go
JustTalDevelops: Changed.

server/session/handler_anvil.go
JustTalDevelops: Split up a few of the biggest blocks of code, please let me know if I should split up more.

server/session/handler_anvil.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// First get the new name and the existing name. The existing name is either the custom name if it exists, or the
+	// item's display name in-game, which is locale dependent.
+	newName := filterStrings[int(a.FilterStringIndex)]
+	existingName := item.DisplayName(input.Item(), s.c.Locale())
+	if customName := input.CustomName(); len(customName) > 0 {
+		existingName = customName
+	}
+
+	// If our existing name isn't the same as the new name, then something changed,  and we should update the custom
```
DaPigGuy, line 95:
```suggestion
	// If our existing name isn't the same as the new name, then something changed, and we should update the custom
```

server/session/handler_anvil.go
```diff
@@ -0,0 +1,292 @@
...
+		if !material.Empty() && updatedRepairCost < material.RepairCost() {
+			updatedRepairCost = material.RepairCost()
+		}
+		if renameCost != actionCost || renameCost == 0 {
+			updatedRepairCost = updatedRepairCost*2 + 1
+		}
+		result = result.WithRepairCost(updatedRepairCost)
+	}
+
+	// If we're in creative mode, we have a 12% chance of the anvil degrading down one state. If that is the case, we
```
DaPigGuy, line 140:
```suggestion
	// If we're not in creative mode, we have a 12% chance of the anvil degrading down one state. If that is the case, we
```

server/session/handler_anvil.go
```diff
@@ -0,0 +1,292 @@
...
+		// level. If the first level is equal to the enchantment's level, increase the result level by one.
+		resultLevel := max(firstLevel, enchant.Level())
+		if firstLevel == enchant.Level() {
+			resultLevel = firstLevel + 1
+		}
+
+		// If the result level is greater than the enchantment's maximum level, set it to the maximum level
+		// instead.
+		if resultLevel > enchantType.MaxLevel() {
+			resultLevel = enchantType.MaxLevel()
```
DaPigGuy, line 255: Should probably be considered incompatible; vanilla will not allow combining if all enchantments on the material item are already maxed on the input item.

DaPigGuy: Similarly if all enchantments on the material item are lower in level than on the input item, combining will be blocked.

DaPigGuy (changes requested):

Sandertv (approved):
With the comments and a couple of functions that were extracted from the original code, this looks a ton more readable. Still pretty long functions but acceptable, I think.

You guys have a lot better of an idea about the anvil functionality itself. I wasn't able to find any obvious logical flaws, but all I can say is make sure that the behaviour matches bedrock edition. If you think it does and is ready, please feel free to merge this on your own accord.

PR #535, opened 2022-07-05
Sort calls to register block and item.

tristanmorgan:
Sorting the calls so that it is easier to find calls and merge new ones in.


Sandertv (approved):
I think this is a good idea, that'll clean up diffs a bit as well. Thanks for the PR!

PR #536, opened 2022-07-06
Sandstone slabs.

tristanmorgan:
WIP implementation for Sandstone slabs and their variants. I feel the `allSandstoneSlab()` needs a re-work before merge.

Partially supersedes #47


server/block/sandstone_slab.go
```diff
@@ -0,0 +1,178 @@
...
+// SandstoneSlab are blocks that allow entities to walk up blocks without jumping. They are crafted using sandstone.
+type SandstoneSlab struct {
+	transparent
+
+	// Smooth will specify if the sandstone block should be smooth sandstone slab or normal sandstone slab.
+	Smooth bool
+
+	// Cut specifies if the sandstone type is cut or not. When set to true, the sandstone slab type will represent its
+	// cut variant, for example cut sandstone slab.
+	Cut bool
```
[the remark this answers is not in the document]
tristanmorgan: I've done a re-write and made the SandstoneType include the red variants.

Sandertv:
Thanks for the updates! I'd like some feedback from other people on whether we should merge the red flag + sandstone type should be merged. Personally I think it might be more flexible if we keep `Red` and `Type` separate. Any thoughts on that?

tristanmorgan:
> Personally I think it might be more flexible if we keep Red and Type separate. Any thoughts on that?
I think you're right, I'l have a go at it and see how it falls out.

tristanmorgan:
Yup. much nicer.

server/block/sandstone_type.go
```diff
@@ -71,6 +71,26 @@ func (s sandstone) String() string {
...
 
+// Smooth varient
+func (s sandstone) Smooth() bool {
+	return s == 3
+}
+
+// Cut varient
+func (s sandstone) Cut() bool {
+	return s == 1
+}
```
Sandertv, line 82: You should be able to compare `SandstoneType`s directly using `==`, so I think we don't need these functions. Could just replace them with `s == SmoothSandstone()` and `s == CutSandstone()` respectively.

server/block/sandstone_type.go
```diff
@@ -71,6 +71,26 @@ func (s sandstone) String() string {
...
+
+// Can be a Slab
+func (s sandstone) SlabAble() bool {
+	return s != 2
+}
+
+// Can be a Stairs
```
Sandertv, line 89: Make sure to use a full sentence here with full stop and start it with the function name, so something like `SlabAble checks if a slab with this type can exist.` and `StairAble checks if a stair with this type can exist.`

Sandertv (changes requested):
Got some really minor last comments. Should be good to go once you resolve these!

server/block/sandstone_type.go
tristanmorgan: I got an error on the conversion to use the type:
`invalid operation: s == SmoothSandstone() (mismatched types sandstone and SandstoneType)`
please let me know you preferred option.

Sandertv: Sorry, I wasn't clear enough with what I meant. I quickly made a change and it should be all good now!

Sandertv (approved):
Thanks a lot for the PR! All looks good now.

PR #541, opened 2022-07-07
Add ghast sounds

Prim69:


server/world/sound/item.go
```diff
@@ -50,3 +50,9 @@ type UseSpyglass struct{ sound }
 
 // StopUsingSpyglass is a sound played when a player stops using a spyglass.
 type StopUsingSpyglass struct{ sound }
+
+// GhastWarning is a sound played when a ghast is ready to attack.
+type GhastWarning struct{ sound }
```
T14Raptor, line 55: Would add these to the entity.go file instead

Prim69:
done

JustTalDevelops (approved):

PR #542, opened 2022-07-07
Support for the default flame particle

Prim69:
The current flame particle is actually a colored flame, the difference can be seen here.
image
The top particle is particle.Flame, while the bottom particle is an orange particle.ColouredFlame

image
image


T14Raptor:
They do look noticeably different though even when putting the color to orange

JustTalDevelops:
Yeah, side by side comparison:
image

I think there should be an option for the actual flame particle, as the coloured flame particle looks much more dull compared to it which is not ideal for everyone.

Prim69:
is this alright?

Sandertv (approved):

PR #545, opened 2022-07-08
Add fire methods to the living interface

Prim69:


Sandertv:
Why is this change necessary?

Prim69:
Living items can be set on fire (discussed with tal on discord)

Sandertv:
Right, but what is the actual use of this? There's hardly a point in adding methods to an interface if those methods just aren't used anywhere.

Sandertv:
Besides, there are also entities that are not `Living` that can also be set on fire (arrows, snowballs), so baking this into the `Living` interface wouldn't necessarily make sense.

Sandertv:
I don't see the point of adding an interface if it's not used in dragonfly.

T14Raptor:
Actually, there already is an interface for it

https://github.com/df-mc/dragonfly/blob/master/server/entity/flammable.go

T14Raptor:
Closing due to an `entity.Flammable` interface already existing.

PR #547, opened 2022-07-09
Custom colors on leather armor

Prim69:


Prim69:
this is ready for a review

server/item/boots.go
```diff
@@ -60,5 +60,23 @@ func (b Boots) Boots() bool {
...
+			t.Colour = rgbaFromInt32(v)
+		}
+	}
+	return b
+}
+
+// EncodeNBT ...
+func (b Boots) EncodeNBT() map[string]any {
+	if t, ok := b.Tier.(ArmourTierLeather); ok {
+		return map[string]any{"customColor": int32FromRGBA(t.Colour)}
```
Sandertv, line 79: I think this `customColor` is not supposed to be written if `(t.Colour == color.RGBA{})`.

server/item/boots.go
```diff
@@ -60,5 +60,23 @@ func (b Boots) Boots() bool {
...
 func (b Boots) EncodeItem() (name string, meta int16) {
-	return "minecraft:" + b.Tier.Name + "_boots", 0
+	return "minecraft:" + b.Tier.Name() + "_boots", 0
+}
+
+// DecodeNBT ...
+func (b Boots) DecodeNBT(data map[string]any) any {
+	if t, ok := b.Tier.(ArmourTierLeather); ok {
+		if v, ok := data["customColor"].(int32); ok {
+			t.Colour = rgbaFromInt32(v)
```
Sandertv, line 70: Does this actually work? `b.Tier` is not stored as a pointer so I doubt this actually propagates properly. You'd want to run something like `b.Tier = ArmourTierLeather{Colour: rgbaFromInt32(v)}` I think.

server/item/item.go
```diff
@@ -215,3 +217,21 @@ func eyePosition(e world.Entity) mgl64.Vec3 {
...
+	}
+	return int32(binary.BigEndian.Uint32([]byte{x.A, x.R, x.G, x.B}))
+}
+
+// rgbaFromInt32 converts an int32 into a color.RGBA. These int32s are present in things such as signs and dyed leather armour.
+func rgbaFromInt32(x int32) color.RGBA {
+	b := make([]byte, 4)
+	binary.BigEndian.PutUint32(b, uint32(x))
+
+	return color.RGBA{A: b[0], R: b[1], G: b[2], B: b[3]}
```
Sandertv, line 236: Not a massive fan of the fact that this is identical code to that in `internal/nbtconv`, but I'm guessing that's not usable because of a circular dependency?

Sandertv (changes requested):
Thanks for the PR! I've left some comments, please resolve them whenever you have the time.

Sandertv (approved):

PR #549, opened 2022-07-10
Added a fall manager for entities

HashimTheArab:


Sandertv (commented):
Thanks for your PR! Please make sure to fix the issues pointed out by the CI. Also, for future reference, please do add a description for your PR so we can get an idea of the reason for your PR and the changes you've made.

server/entity/fall.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		if dmg < 0.5 {
+			return
+		}
+		p.Hurt(math.Ceil(dmg), damage.SourceFall{})
+	}
+}
+
+// CheckOnGround checks if the entity is currently considered to be on the ground.
+func (f *FallManager) CheckOnGround(w *world.World) bool {
```
JustTalDevelops, line 116: This is pretty useless since f.e already has an `OnGround` function.

server/entity/fall.go
```diff
@@ -0,0 +1,142 @@
...
+
+// ResetFallDistance resets the player's fall distance.
+func (f *FallManager) ResetFallDistance() {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.fallDistance = 0
+}
+
+// OnGround returns whether the entity is currently considered to be on the ground.
+func (f *FallManager) OnGround() bool {
```
JustTalDevelops, line 60: Don't see why this needs to be here?

server/entity/fall.go
Prim69: Well if the entity is a player then it will require the fall managers onground check (refer to player.OnGround)

Prim69:
anything else on this?

server/entity/fall.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/df-mc/dragonfly/server/world"
+	"math"
+	"sync"
+)
+
+// FallManager handles entities that can fall.
+type FallManager struct {
+	mu           sync.Mutex
+	e            fallEntity
+	fallDistance float64
```
T14Raptor, line 16: This can be made an atomic.Float64 then the mutex can be removed

Prim69: done :pog:

JustTalDevelops:
Would also remove the fall distance calculation code in falling block and make it adopt this system.

Prim69:
> Would also remove the fall distance calculation code in falling block and make it adopt this system.

is the latest commit fine?

Sandertv:
There is a bit of an awkward "cycle" between this fall manager and the movement computer. Maybe it'd be an idea to merge them?

Prim69:
> There is a bit of an awkward "cycle" between this fall manager and the movement computer. Maybe it'd be an idea to merge them?

Could you elaborate more on that

PR #559, opened 2022-07-13
Implement walls

TwistedAsylumMC:


server/block/stone_wall_type.go
```diff
@@ -0,0 +1,127 @@
+package block
+
+// StoneWallType represents a type of cobblestone wall.
+type StoneWallType struct {
```
Sandertv, line 4: I think we should have walls that match the way we structure blocks, with for example SandstoneWall{red bool}, CobblestoneWall{mossy bool} etc. It's much more convenient to convert a block to its respective wall that way. Maybe we could re-use code using generics somehow?

server/block/model/wall.go
```diff
@@ -0,0 +1,61 @@
...
+	return boxes
+}
+
+// FaceSolid ...
+func (w Wall) FaceSolid(cube.Pos, cube.Face, *world.World) bool {
+	return true
+}
+
+// heightFromConnection calculates the height of a connection based on the provided connection type.
+func (w Wall) heightFromConnection(connection string) float64 {
```
JustTalDevelops, line 56: You could make these connections be the heights themselves, and have a function in WallConnectionType to return the Height of the type. (and assume no connection if the height is zero)

server/block/wall.go
```diff
@@ -0,0 +1,244 @@
...
+		w.SouthConnection = connection
+	case cube.West:
+		w.WestConnection = connection
+	}
+	return w
+}
+
+// calculateState returns the wall with the correct state based on walls around it. If any of the connections have been
+// updated then the wall and true are returned.
+func (w Wall) calculateState(wo *world.World, pos cube.Pos) (Wall, bool) {
```
JustTalDevelops, line 137: Might be helpful to add some documentation here and space it out a bit more so it's more readable.

server/block/wall.go
```diff
@@ -0,0 +1,282 @@
...
+	return w, updated
+}
+
+// calculatePost calculates the correct post bit for the wall at a given position in a world. The updated wall is
+// returned and a bool to determine if any changes were made.
+func (w Wall) calculatePost(wo *world.World, pos cube.Pos) (Wall, bool) {
+	var updated bool
+	abovePos := pos.Add(cube.Pos{0, 1, 0})
+	above := wo.Block(abovePos)
+	var connections int
```
JustTalDevelops, line 197:
```go
	connections := len(sliceutil.Filter(cube.HorizontalFaces(), func(f cube.Face) bool {
		return w.ConnectionType(face.Direction()) != NoWallConnection()
	}))
```

JustTalDevelops (approved):
Looks good to me!

server/block/model/trapdoor.go
```diff
@@ -19,14 +19,19 @@ type Trapdoor struct {
 // top part of the block.
 func (t Trapdoor) BBox(cube.Pos, *world.World) []cube.BBox {
 	if t.Open {
-		return []cube.BBox{full.ExtendTowards(t.Facing.Face(), -0.8125)}
+		return []cube.BBox{full.ExtendTowards(t.Facing.Face().Opposite(), -0.8125)}
```
DaPigGuy, line 22: The original calculation seems to be correct. You can test by shooting a bow at opened trapdoors.

server/block/wall_connection_type.go
```diff
@@ -0,0 +1,59 @@
+package block
+
+// WallConnectionType represents the connection type of a wall.
+type WallConnectionType struct {
+	wallConnectionType
+}
+
+// NoWallConnection returns the cobblestone wall variant.
```
DaPigGuy, line 8: Fix these comments

server/block/wall.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+}
+
+// BreakInfo ...
+func (w Wall) BreakInfo() BreakInfo {
+	hardness := 2.0
+	name, _ := encodeWallBlock(w.Block)
+	if name == "cobbled_deepslate" || name == "deepslate_brick" || name == "deepslate_tile" || name == "polished_deepslate" {
+		hardness = 3.5
+	}
```
DaPigGuy, line 73: This should probably be in WallType

DaPigGuy: The hardness for several wall types are also missing

server/block/model/wall.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	if w.WestConnection > 0 {
+		boxes = append(boxes, cube.Box(0.75, 0, 0.25, 1, w.EastConnection, 0.75))
+	}
+	return boxes
+}
+
+// FaceSolid ...
+func (w Wall) FaceSolid(cube.Pos, cube.Face, *world.World) bool {
+	return true
```
DaPigGuy, line 46: This probably should not be returning true for all faces

DaPigGuy (changes requested):

server/block/wall.go
TwistedAsylumMC: There is no WallType

JustTalDevelops:
Make sure you update the solid check @TwistedAsylumMC.

server/block/model/trapdoor.go
```diff
@@ -27,6 +27,11 @@ func (t Trapdoor) BBox(cube.Pos, *world.World) []cube.BBox {
 }
 
 // FaceSolid always returns false.
```
DaPigGuy, line 29: Outdated doc

DaPigGuy (changes requested):

server/block/wall.go
```diff
@@ -0,0 +1,278 @@
...
+func (w Wall) calculatePost(wo *world.World, pos cube.Pos) (Wall, bool) {
+	var updated bool
+	abovePos := pos.Add(cube.Pos{0, 1, 0})
+	above := wo.Block(abovePos)
+	connections := len(sliceutil.Filter(cube.HorizontalFaces(), func(face cube.Face) bool {
+		return w.ConnectionType(face.Direction()) != NoWallConnection()
+	}))
+	var post bool
+	switch above := above.(type) {
+	case Air:
```
DaPigGuy, line 201:
```suggestion
```

DaPigGuy (approved):

server/block/wall_type.go
```diff
@@ -0,0 +1,138 @@
...
+			return "stone_brick", 7
+		} else if block.Type == MossyStoneBricks() {
+			return "mossy_stone_brick", 8
+		}
+	}
+	panic("invalid block used for wall")
+}
+
+// calculateWallHardness calculates the appropriate hardness for a wall based on the provided block.
+func calculateWallHardness(block world.Block) float64 {
```
Sandertv, line 63: Is it not an option to grab the hardness from the block's breakinfo? Or does that not match at all?

server/block/wall_type.go
```diff
@@ -0,0 +1,138 @@
...
+	case StoneBricks:
+		if block.Type == NormalStoneBricks() || block.Type == MossyStoneBricks() {
+			return 1.5
+		}
+	}
+	panic("invalid block used for wall")
+}
+
+// WallBlocks returns a list of all possible blocks for a wall.
+func WallBlocks() []world.Block {
```
Sandertv, line 113: Does this need to be exported?

Sandertv (commented):
Got a couple of comments. For the hash thing, I'm not entirely sure how to solve this. Maybe we could at least add a check somewhere to make sure it doesn't overflow and break silently.

server/block/wall_type.go
TwistedAsylumMC: Surprising they all seem to match, this method never crossed my mind

server/block/wall_type.go
TwistedAsylumMC: And it doesn't need to be, it just is for consistency

Sandertv: Fair enough.

JustTalDevelops:
Should probably also account for blast resistances now that explosions are merged.

cmd/blockhash/main.go
```diff
@@ -211,6 +211,8 @@ func (b *hashBuilder) ftype(structName, s string, expr ast.Expr, directives map[
 		return "uint64(boolByte(" + s + "))", 1
 	case "int":
 		return "uint64(" + s + ")", 8
+	case "Block":
+		return s + ".Hash()", 16
```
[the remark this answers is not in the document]
TwistedAsylumMC: There's now a temporary check to make sure this doesn't happen

Sandertv (approved):

PR #562, opened 2022-07-14
Implemented explosions

T14Raptor:


JustTalDevelops (changes requested):
(all requested changes resolved)

server/block/break_info.go
```diff
@@ -85,6 +85,8 @@ type BreakInfo struct {
 	Drops func(t item.Tool, enchantments []item.Enchantment) []item.Stack
 	// XPDrops is the range of XP a block can drop when broken.
 	XPDrops XPDropRange
+	// BlastResistance ...
+	BlastResistance float64
```
Sandertv, line 89: This should be documented. Also maybe we should default this to a number that is based on the hardness? I think that applies to the majority of the blocks.

server/block/block.go
```diff
@@ -230,6 +245,75 @@ func newFlammabilityInfo(encouragement, flammability int, lavaFlammable bool) Fl
...
+	// RemoveEffect removes any effect that might currently be active on the entity.
+	RemoveEffect(e effect.Type)
+	// Effects returns any effect currently applied to the entity. The returned effects are guaranteed not to have
+	// expired when returned.
+	Effects() []effect.Effect
+	// Speed returns the current speed of the living entity. The default value is different for each entity.
+	Speed() float64
+	// SetSpeed sets the speed of an entity to a new value.
+	SetSpeed(float64)
+}
```
Sandertv, line 290: I wouldn't copy this whole interface. Just remove methods that are not used in this package.

server/block/block.go
```diff
@@ -230,6 +245,75 @@ func newFlammabilityInfo(encouragement, flammability int, lavaFlammable bool) Fl
...
+type flammableEntity interface {
+	// FireProof is whether the entity is currently fireproof.
+	FireProof() bool
+	// OnFireDuration returns duration of fire in ticks.
+	OnFireDuration() time.Duration
+	// SetOnFire sets the entity on fire for the specified duration.
+	SetOnFire(duration time.Duration)
+	// Extinguish extinguishes the entity.
+	Extinguish()
+}
```
Sandertv, line 301: Same counts for this interface.

server/block/explosion.go
```diff
@@ -0,0 +1,164 @@
...
+type Explodable interface {
+	Explode(pos cube.Pos, c ExplosionConfig)
+}
+
+type ExplosionConfig struct {
+	World *world.World
+	Pos   mgl64.Vec3
+	Size  float64
+	Rand  rand.Source
+	Fire  bool
```
Sandertv, line 42: Would maybe make this a little less short, maybe something like `CreateFire` or `SpawnFire`.

Sandertv (changes requested):
Got a couple of comments. There's a lot of TODOs still so I won't do a super thorough review just yet. I do think the general idea here is good.

JustTalDevelops (changes requested):
Just a couple more changes that should probably be made:
- Don't think we should remove the `* 5` when initializing BreakInfo, since all of the added blast resistances are also `* 5`. I'd rather just `/ 5` in the actual explosion code. I believe PM also does the same, would also be more consistent with BDS and whatnot. Also might cause issues if other areas of the code expect vanilla blast resistance values.

- Better names for `double7` and `double8`, especially just `double`, it doesn't make sense here. (assuming you took this from the Java code)

- Exposure is not being utilized. (if it's not working as intended, you might want to fix it too)

- Would remove the remaining debug code.

server/entity/ender_pearl.go
```diff
@@ -131,6 +132,11 @@ func (e *EnderPearl) New(pos, vel mgl64.Vec3, yaw, pitch float64) world.Entity {
 	return pearl
 }
 
+// Explode ...
+func (e *EnderPearl) Explode(c block.ExplosionConfig, impact float64) {
+	e.SetVelocity(e.Position().Sub(c.Pos).Normalize().Mul(impact))
```
JustTalDevelops, line 137: This should add on to the velocity - it shouldn't override existing velocity.

server/entity/snowball.go
```diff
@@ -112,6 +113,11 @@ func (s *Snowball) New(pos, vel mgl64.Vec3, yaw, pitch float64) world.Entity {
 	return snow
 }
 
+// Explode ...
+func (s *Snowball) Explode(c block.ExplosionConfig, impact float64) {
+	s.SetVelocity(s.Position().Sub(c.Pos).Normalize().Mul(impact))
```
JustTalDevelops, line 118: This should also add on to the velocity - it shouldn't override existing velocity.

server/entity/splash_potion.go
```diff
@@ -176,6 +177,11 @@ func (s *SplashPotion) New(pos, vel mgl64.Vec3, yaw, pitch float64, t potion.Pot
 	return splash
 }
 
+// Explode ...
+func (s *SplashPotion) Explode(c block.ExplosionConfig, impact float64) {
+	s.SetVelocity(s.Position().Sub(c.Pos).Normalize().Mul(impact))
```
JustTalDevelops, line 182: This should also add on to the velocity - it shouldn't override existing velocity.

JustTalDevelops:
Make sure you resolve the Radius usage in main.go or remove the debugging code.

JustTalDevelops (changes requested):
- No knock-back to players (at least, in creative mode)

JustTalDevelops (approved):

server/block/explosion.go
```diff
@@ -0,0 +1,191 @@
...
+	"time"
+)
+
+// ExplosionConfig is the configuration for an explosion. The world, position, size, sound, particle, and more can all
+// be configured through this configuration.
+type ExplosionConfig struct {
+	// World is the world in which the explosion should take place.
+	World *world.World
+	// Pos is the center position of the explosion.
+	Pos mgl64.Vec3
```
Sandertv, line 22: I would move these to function parameters of the `Do` method so that an `ExplosionConfig` can be re-used more easily.

server/block/explosion.go
```diff
@@ -0,0 +1,191 @@
...
+					continue
+				}
+				rays = append(rays, mgl64.Vec3{x/15*2 - 1, y/15*2 - 1, z/15*2 - 1}.Normalize().Mul(0.3))
+			}
+		}
+	}
+}
+
+// Do performs the explosion as specified by the configuration.
+func (c ExplosionConfig) Do() {
```
Sandertv, line 70: Would probably rename this to `Explode()` because it doesn't necessarily have some sort of `explosion.` prefix.

server/player/player.go
```diff
@@ -655,6 +655,21 @@ func (p *Player) FinalDamageFrom(dmg float64, src damage.Source) float64 {
...
+
+	vel := p.Position().Sub(c.Pos).Normalize().Mul(impact)
+	res := 0.0
+	for _, i := range p.armour.Items() {
+		if a, ok := i.Item().(item.Armour); ok {
+			res += a.KnockBackResistance()
+		}
+	}
+
+	p.SetVelocity(vel.Mul(1 - res))
```
Sandertv, line 670: Any reason not to use Player.KnockBack() for this?

Sandertv (changes requested):
Looks fine mostly. Got a couple of comments that should be pretty easy to resolve.

server/player/player.go
JustTalDevelops: Player.KnockBack() doesn't work on entities that cannot take damage, whereas explosions allow them.

Sandertv: Consider in that case having an unexported `knockBack()` method that does not have that check (or a bool to decide it). That way we can reduce duplicate code.

server/block/explosion.go
JustTalDevelops: Moved.

server/block/explosion.go
JustTalDevelops: Renamed.

server/player/player.go
JustTalDevelops: Changed.

server/block/model/thin.go
```diff
@@ -21,7 +21,7 @@ func (t Thin) BBox(pos cube.Pos, w *world.World) []cube.BBox {
 		pos := pos.Side(f)
 		block := w.Block(pos)
 
-		// TODO(lhochbaum): Do the same check for walls as soon as they're implemented.
+		// TODO(lhochbaum): Explode the same check for walls as soon as they're implemented.
```
Sandertv, line 24: Wrong refactor.

Sandertv (changes requested):
One small error in the refactor. Looks good otherwise, make sure to test once more to make sure all is right. Feel free to merge once you're confident it works.

server/block/model/thin.go
JustTalDevelops: Resolved.

Sandertv (approved):
Looks good! Merge at your own discretion.

PR #565, opened 2022-07-14
Implemented ender chests

T14Raptor:


server/block/ender_chest.go
```diff
@@ -0,0 +1,138 @@
...
+func (c EnderChest) close(w *world.World, pos cube.Pos) {
+	for _, v := range w.Viewers(pos.Vec3()) {
+		v.ViewBlockAction(pos, CloseAction{})
+	}
+	w.PlaySound(pos.Vec3Centre(), sound.ChestClose{})
+}
+
+// EncodeNBT ...
+func (c EnderChest) EncodeNBT() map[string]interface{} {
+	return map[string]interface{}{}
```
Sandertv, line 114: This doesn't contain the `id`. Do ender chests even require block NBT at all?

server/player/player.go
```diff
@@ -2641,10 +2651,11 @@ func (p *Player) Data() Data {
 			OffHand:      offHand,
 			MainHandSlot: p.heldSlot.Load(),
 		},
-		Effects:      p.Effects(),
-		FireTicks:    p.fireTicks.Load(),
-		FallDistance: p.fallDistance.Load(),
-		Dimension:    p.World().Dimension().EncodeDimension(),
+		EnderChestInventory: p.enderChest.Items(),
```
Sandertv, line 2654: I think this should use `p.enderChest.Slots()`. `p.enderChest.Items()` only returns non-zero item stacks and it doesn't retain the association of these stacks with their corresponding slot.

Sandertv (changes requested):
Got two comments. Looks good otherwise.

server/block/ender_chest.go
T14Raptor: Pretty sure they don't, but I implemented world.NBTer to retain viewer data, is there another way to do this?

Sandertv: That makes sense. Might want to document that so someone doesn't remove it on accident in the future.

server/player/player.go
T14Raptor: Indeed it should

Sandertv (approved):

PR #576, opened 2022-07-19
Implemented fireworks and firework stars

JustTalDevelops:
Also ensures that crafting outputs have their NBT data applied.


DaPigGuy (changes requested):
Firework rockets get stuck when spawned against a block, but otherwise looks good

JustTalDevelops:
Not too sure how to fix that, since technically the firework is within a block.

server/entity/firework.go
```diff
@@ -0,0 +1,142 @@
...
+	f.mu.Unlock()
+
+	m.Send()
+
+	if m.pos[1] < float64(w.Range()[0]) && current%10 == 0 {
+		f.close = true
+		return
+	}
+
+	f.mu.Lock()
```
T14Raptor, line 90: Locking shouldn't be needed here

JustTalDevelops: Removed.

DaPigGuy (changes requested):
Missing entity damage on firework explode

JustTalDevelops:
Implemented firework damage.

DaPigGuy (approved):

server/entity/action.go
```diff
@@ -41,6 +41,9 @@ type PickedUpAction struct {
 	action
 }
 
+// FireworkParticleAction is a world.EntityAction that makes a Firework rocket display an explosion particle.
+type FireworkParticleAction struct{ action }
```
Sandertv, line 45: This is a bit of a poor name given fireworks are essentially all particles. Maybe something like `Firework{Explode/Explosion}Action`would be better?

server/entity/firework.go
```diff
@@ -0,0 +1,179 @@
...
+			}) {
+				pos := e.Position()
+				dist := m.pos.Sub(pos).Len()
+				if dist > 5.0 {
+					// The maximum distance allowed is 5.0 blocks.
+					continue
+				}
+				if _, ok := trace.Perform(m.pos, pos, w, e.BBox().Grow(0.3), func(world.Entity) bool {
+					return true
+				}); ok {
```
Sandertv, line 126: This `func(world.Entity) bool` can simply be `nil`.

server/entity/firework.go
```diff
@@ -0,0 +1,179 @@
...
+
+	m.Send()
+
+	if m.pos[1] < float64(w.Range()[0]) && current%10 == 0 {
+		f.close = true
+		return
+	}
+
+	f.ticks--
+	if f.ticks < 0 {
```
Sandertv, line 96: Would invert this check and return to reduce indentation.

server/item/firework.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if durationTicks, ok := fireworks["Flight"].(uint8); ok {
+			f.Duration = time.Duration(durationTicks) * 50 * time.Millisecond
+		}
+	}
+	return f
+}
+
+// RandomizedDuration returns the randomized flight duration of the firework.
+func (f Firework) RandomizedDuration() time.Duration {
```
Sandertv, line 77: Randomised* for BE.

server/item/firework_explosion.go
```diff
@@ -0,0 +1,51 @@
+package item
+
+// FireworkExplosion represents an explosion of a firework.
+type FireworkExplosion struct {
+	// Shape represents the shape of the explosion.
+	Shape FireworkShape
+	// Colour is the colour of the explosion.
+	Colour Colour
+	// Fade is the colour the explosion should fade into.
```
Sandertv, line 9: Might want to document that upon setting this, `Fades` must be set to true. Usually I'd be more of a fan to check against the default value to decide if it was changed, but obviously here that isn't possible. Documenting that would be useful though.

Sandertv (changes requested):
Looks fine mostly. I got some minor comments.

server/entity/firework.go
JustTalDevelops: Don't think so? We need to ignore all entities, not allow all entities.

Sandertv: Ah I see, my bad!

server/entity/action.go
JustTalDevelops: Renamed.

server/entity/firework.go
JustTalDevelops: Inverted.

server/item/firework_explosion.go
JustTalDevelops: Documented.

server/item/firework.go
JustTalDevelops: Renamed. (even though you're wrong)

Sandertv (approved):
Looks good to me, thanks for the PR!

PR #597, opened 2022-07-25
Implemented elytra

JustTalDevelops:


server/item/elytra.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Use handles the using of an elytra to auto-equip it in an armour slot.
+func (Elytra) Use(_ *world.World, _ User, ctx *UseContext) bool {
+	ctx.SwapHeldWithArmour(1)
+	return false
+}
+
+// DurabilityInfo ...
+func (Elytra) DurabilityInfo() DurabilityInfo {
+	return DurabilityInfo{
```
DaPigGuy, line 16: iirc, elytra items don't break in the traditional sense that they disappear but rather become unusable

server/player/player.go
```diff
@@ -2208,7 +2246,15 @@ func (p *Player) Tick(w *world.World, current int64) {
 		}
 	}
 
-	p.checkBlockCollisions(w)
+	if _, ok := p.Armour().Chestplate().Item().(item.Elytra); ok && p.Gliding() {
+		if t := p.glideTicks.Inc(); t%10 == 0 {
+			if (t/10)%2 == 0 {
```
DaPigGuy, line 2251: Isn't this just `t%20 == 0`?

server/item/durability.go
```diff
@@ -19,6 +19,8 @@ type DurabilityInfo struct {
 	// AttackDurability and BreakDurability are the losses in durability that the item sustains when they are
 	// used to do the respective actions.
 	AttackDurability, BreakDurability int
+	// Persistent is true if the item is persistent, i.e. it will not be destroyed when at it's last durability stage.
```
DaPigGuy, line 22:
```suggestion
	// Persistent is true if the item is persistent, i.e. it will not be destroyed when at its last durability stage.
```

server/item/elytra.go
JustTalDevelops: Implemented.

server/player/player.go
JustTalDevelops: Pretty sure actually, updated.

server/player/player.go
```diff
@@ -2399,8 +2443,62 @@ func (p *Player) insideOfSolid(w *world.World) bool {
...
+			deltaX = entityBBox.XOffset(blockBBox, deltaX)
+		}
+		entityBBox = entityBBox.Translate(mgl64.Vec3{deltaX})
+	}
+	if !mgl64.FloatEqualThreshold(deltaZ, 0, epsilon) {
+		// And finally on the Z axis.
+		for _, blockBBox := range blocks {
+			deltaZ = entityBBox.ZOffset(blockBBox, deltaZ)
+		}
+	}
```
DaPigGuy, line 2493: Seem to be missing translation for the Z axis?

DaPigGuy (approved):

server/entity/firework.go
```diff
@@ -138,12 +153,18 @@ func (f *Firework) Tick(w *world.World, current int64) {
...
 // only returns it.
-func (f *Firework) New(pos mgl64.Vec3, yaw, pitch float64, firework item.Firework, owner world.Entity) world.Entity {
+func (f *Firework) New(pos mgl64.Vec3, yaw, pitch float64, attached bool, firework item.Firework, owner world.Entity) world.Entity {
 	fw := NewFirework(pos, yaw, pitch, firework)
+	fw.attached = attached
 	fw.owner = owner
 	return fw
 }
 
+// Attached ...
```
Sandertv, line 163: Would document this.

server/item/elytra.go
```diff
@@ -0,0 +1,57 @@
...
+		MaxDurability: 433,
+		Persistent:    true,
+		BrokenItem:    simpleItem(Stack{}),
```
Sandertv, line 19: Have you tried setting the BrokenItem to an Elytra with 0 durability? That way we might not need this extra Persistent bool.

server/player/player.go
```diff
@@ -2208,7 +2246,13 @@ func (p *Player) Tick(w *world.World, current int64) {
 		}
 	}
 
-	p.checkBlockCollisions(w)
+	if _, ok := p.Armour().Chestplate().Item().(item.Elytra); ok && p.Gliding() {
+		if t := p.glideTicks.Inc(); t%20 == 0 {
+			p.armour.SetChestplate(p.damageItem(p.Armour().Chestplate(), 1))
```
Sandertv, line 2251: We should probably force the player to stop gliding if the elytra breaks here.

Sandertv (changes requested):
This looks fine mostly. Got a couple of minor comments.

server/item/elytra.go
JustTalDevelops: It'll still play the break sound etc no?

Sandertv: That'd be reasonable though, no? Only thing I'm worried about is that maybe it'll be removed from its slot, but you could test that I guess?

JustTalDevelops: The sound doesn't play in vanilla I don't think, it just instantly prevents you from gliding

server/entity/firework.go
JustTalDevelops: Documented.

server/item/elytra.go
JustTalDevelops: Removed.

server/player/player.go
JustTalDevelops: Done.

JustTalDevelops:
Also added validation, ensuring a player has an elytra before letting them fly.

server/item/elytra.go
```diff
@@ -0,0 +1,56 @@
...
+func (Elytra) Use(_ *world.World, _ User, ctx *UseContext) bool {
+	ctx.SwapHeldWithArmour(1)
+	return false
+}
+
+// DurabilityInfo ...
+func (e Elytra) DurabilityInfo() DurabilityInfo {
+	return DurabilityInfo{
+		MaxDurability: 433,
+		BrokenItem:    simpleItem(NewStack(e, 1).WithDurability(1)),
```
DaPigGuy, line 18: This wipes all item data, i.e enchantments, custom names

server/player/player.go
```diff
@@ -1028,6 +1031,31 @@ func (p *Player) StopSwimming() {
 	p.updateState()
 }
 
+// StartGliding makes the player start gliding if it is not currently doing so.
+func (p *Player) StartGliding() {
+	if !p.gliding.CAS(false, true) {
+		return
+	}
+	if _, ok := p.Armour().Chestplate().Item().(item.Elytra); ok {
```
DaPigGuy, line 1039: Need to also check durability

server/player/player.go
```diff
@@ -2215,7 +2256,17 @@ func (p *Player) Tick(w *world.World, current int64) {
 		}
 	}
 
-	p.checkBlockCollisions(w)
+	if _, ok := p.Armour().Chestplate().Item().(item.Elytra); ok && p.Gliding() {
+		if t := p.glideTicks.Inc(); t%20 == 0 {
+			d := p.damageItem(p.Armour().Chestplate(), 1)
+			p.armour.SetChestplate(d)
+			if d.Empty() {
```
DaPigGuy, line 2263: The chestplate slot will not be empty

server/item/elytra.go
JustTalDevelops: Not sure how to properly implement this then... Any ideas @Sandertv? Or should we just bring back the "Persistent" field.

Sandertv:
71e6fb1ae738199137d139eba422ebf44063c127 adds sounds when fast equipping. Could you add the case for elytras and assign the right sound event for them in session/world.go?

server/item/elytra.go
Sandertv: Hmmm yeah that's an issue... I guess the easiest solution would be to bring that field back.

JustTalDevelops: Updated.

server/player/player.go
JustTalDevelops: Fixed.

server/player/player.go
JustTalDevelops: Fixed.

DaPigGuy (approved):

Sandertv (approved):
Looks good to me, thanks for the PR!

PR #598, opened 2022-07-25
implement the milk bucket.

RestartFU:


JustTalDevelops (approved):

server/item/register.go
```diff
@@ -23,6 +23,7 @@ func init() {
 	world.RegisterItem(TurtleShell{})
 
 	world.RegisterItem(Bucket{})
+	world.RegisterItem(MilkBucket{})
```
JustTalDevelops, line 26: One thing I'm unsure about is the fact that `Bucket` and `MilkBucket` are split up, however I do understand that there's probably not a great way around this. Maybe you could rename `Bucket` to `LiquidBucket`?

RestartFU: sure

Sandertv:
I don't think having a separate `MilkBucket` and `LiquidBucket` type is the right approach for this given the rest of the item/block design. Maybe we should have a `Content BucketContent(Type)` field and have BucketContent be either LiquidContent(world.Liquid), MilkContent() (or, later on, FishContent(...))

RestartFU:
> I don't think having a separate `MilkBucket` and `LiquidBucket` type is the right approach for this given the rest of the item/block design. Maybe we should have a `Content BucketContent(Type)` field and have BucketContent be either LiquidContent(world.Liquid), MilkContent() (or, later on, FishContent(...))

I just did a commit, I don't think that my implementation is great, since we need to make it a consumable, even when its content isn't milk.

server/item/bucket.go
```diff
@@ -8,10 +8,34 @@ import (
...
+	milk   bool
+}
+
+// LiquidContent returns a new BucketContent with the liquid passed in.
+func LiquidContent(l world.Liquid) BucketContent {
+	return BucketContent{liquid: l}
+}
+
+// MilkContent returns a new BucketContent with the milk flag set.
+func MilkContent() BucketContent {
```
Sandertv, line 23: Would call these `LiquidBucketContent` and `MilkBucketContent`.

Sandertv (changes requested):
Your change is pretty much exactly what I had in mind. I've got one thing to rename, but this should be good to go afterwards.

Sandertv (approved):
Looks good now, thanks for the PR!

PR #605, opened 2022-07-26
Improvement/block layering

abimek:
Small changes to the block layering system to enable snow layers.

### DIFF
--- a/server/world/world.go
+++ b/server/world/world.go
@@ -408,7 +408,7 @@ func (w *World) DisplaceableBlock(pos cube.Pos) (DisplaceableBlock, bool) {
 		return nil, false
 	}
 	// we could possibly replace this with runtime reflections
-	if d, ok := w.Block(pos).(DisplaceableBlock); ok {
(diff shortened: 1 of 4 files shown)


server/world/world.go
```diff
@@ -385,6 +389,38 @@ func (w *World) BuildStructure(pos cube.Pos, s Structure) {
...
+	defer c.Unlock()
+	x, y, z := uint8(pos[0]), int16(pos[1]), uint8(pos[2])
+
+	id := c.Block(x, y, z, 0)
+	b, ok := BlockByRuntimeID(id)
+	if !ok {
+		w.conf.Log.Errorf("failed getting liquid: cannot get block by runtime ID %v", id)
+		return nil, false
+	}
+	// we could possibly replace this with runtime reflections
```
T14Raptor, line 410: Why?

abimek: yea ignore that comment, that was before I decided to go with a similar approach, gonna remove it

PR #606, opened 2022-07-27
Implemented beds

JustTalDevelops:


server/block/bed.go
```diff
@@ -0,0 +1,229 @@
...
+	headSide, headPos, ok := b.head(pos, w)
+	if !ok {
+		return false
+	}
+
+	w.SetPlayerSpawn(s.UUID(), headPos)
+	s.Message(text.Colourf("<grey>Respawn point set</grey>"))
+
+	time := w.Time() % world.TimeFull
+	if !(time >= world.TimeNight && time < world.TimeSunrise) {
```
DaPigGuy, line 109: Players can also sleep during thunderstorms despite the message weirdly not mentioning that in Bedrock Edition.

server/block/bed.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Model ...
+func (Bed) Model() world.BlockModel {
+	return model.Bed{}
+}
+
+// BreakInfo ...
+func (b Bed) BreakInfo() BreakInfo {
+	return newBreakInfo(0.2, alwaysHarvestable, axeEffective, oneOf(b))
```
DaPigGuy, line 35:
```suggestion
	return newBreakInfo(0.2, alwaysHarvestable, nothingEffective, oneOf(b))
```

server/world/sleep.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+	}
+	if !advanceTime {
+		// We can't advance the time - not everyone is sleeping.
+		return
+	}
+
+	totalTime := t.w.Time()
+	timeOfDay := totalTime % TimeFull
+	if timeOfDay >= TimeNight && timeOfDay < TimeSunrise {
```
DaPigGuy, line 42: See above comment on thunderstorms

DaPigGuy (changes requested):

server/block/bed.go
```diff
@@ -0,0 +1,229 @@
...
+		return false
+	}
+
+	userPos := s.Position()
+	if sidePos.Vec3Middle().Sub(userPos).Len() > 4 && pos.Vec3Middle().Sub(userPos).Len() > 4 {
+		s.Message(text.Colourf("<grey>Bed is too far away</grey>"))
+		return true
+	}
+
+	headSide, headPos, ok := b.head(pos, w)
```
DaPigGuy, line 100: Waterlogging the head of the bed blocks sleeping unless the user has either Water Breathing or Conduit Power.

server/player/player.go
```diff
@@ -1050,6 +1058,54 @@ func (p *Player) StopFlying() {
...
+	w.SetBlock(pos, b, nil)
+	w.SetSleepRequirement(time.Second * 5)
+
+	p.sleeping.Store(true)
+	p.sleepPos.Store(pos)
+	p.updateState()
+}
+
+// Wake forces the player out of bed if they are sleeping.
+func (p *Player) Wake() {
```
DaPigGuy, line 1080: Getting damaged in any way should also wake the player

server/block/bed.go
```diff
@@ -0,0 +1,237 @@
...
+
+// Bed is a block, allowing players to sleep to set their spawns and skip the night.
+type Bed struct {
+	transparent
+
+	// Colour is the colour of the bed.
+	Colour item.Colour
+	// Facing is the direction that the bed is facing.
+	Facing cube.Direction
+	// Head is true if the bed is the head side or the bottom side.
```
DaPigGuy, line 21:
```suggestion
	// Head is true if the bed is the head side.
```

server/block/bed.go
```diff
@@ -0,0 +1,237 @@
...
+			return
+		}
+		if s, ok := headSide.User.(world.Sleeper); ok {
+			s.Wake()
+		}
+	})
+}
+
+// CanDisplace ...
+func (Bed) CanDisplace(b world.Liquid) bool {
```
DaPigGuy, line 51: Use the new sourceWaterDisplacer struct

JustTalDevelops: Updated.

DaPigGuy (approved):

server/world/world.go
```diff
@@ -777,6 +789,41 @@ func (w *World) Entities() []Entity {
...
+}
+
+// BroadcastSleepingReminder broadcasts a sleeping reminder message to all sleepers in the world, excluding the sleeper
+// passed.
+func (w *World) BroadcastSleepingReminder(sleeper Sleeper) {
+	for _, s := range w.Sleepers() {
+		if s == sleeper {
+			continue
+		}
+		s.Messaget("chat.type.sleeping", sleeper.Name())
```
Sandertv, line 823: No `%` here, is that correct?

server/world/world.go
```diff
@@ -777,6 +789,41 @@ func (w *World) Entities() []Entity {
 	return m
 }
 
+// Sleepers returns a list of all sleeping entities currently added to the World.
+func (w *World) Sleepers() []Sleeper {
+	ent := w.Entities()
+	sleepers := make([]Sleeper, 0, len(ent))
```
Sandertv, line 795: This will usually allocate a slice that is way too big, because hardly any entities are sleepers. I would reduce this significantly, maybe `len(ent)/40` or something like that.

Sandertv (changes requested):
I've left a couple of comments, some of which we already discussed in the VC.

server/world/world.go
JustTalDevelops: Yes, it only needs to be added if there is text unrelated to the translation key.

server/player/player.go
```diff
@@ -1050,6 +1064,70 @@ func (p *Player) StopFlying() {
...
+	if b, ok := w.Block(pos).(block.Bed); ok {
+		if b.User != nil {
+			// The player cannot sleep here.
+			return
+		}
+		b.User = p
+		w.SetBlock(pos, b, nil)
+	}
+
+	w.SetRequiredSleepDuration(time.Second * 5)
```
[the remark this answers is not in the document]
JustTalDevelops: What do you mean by "something internal"? You must be sleeping in a bed for at least 5 seconds before the day progresses, and that cool down gets reset back to 5 every time someone else starts to sleep.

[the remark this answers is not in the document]
Sandertv: I mean having only a required sleep duration in world's settings and updating/tracking how much longer must be slept inside of World automatically, instead of having a `SetRequiredSleepDuration`.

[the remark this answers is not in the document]
JustTalDevelops: That won't work though. The sleep duration is reset in vanilla if the player stops and then starts sleeping again. If you're suggesting we should make the default required sleep duration configurable via world settings, maybe, but that doesn't entirely make sense since vanilla doesn't support it anyway.

server/block/bed.go
```diff
@@ -0,0 +1,232 @@
...
+// side returns the other side of the bed. If the other side is not a bed, the third return value is false.
+func (b Bed) side(pos cube.Pos, w *world.World) (Bed, cube.Pos, bool) {
+	face := b.Facing.Face()
+	if b.Head {
+		face = face.Opposite()
+	}
+
+	sidePos := pos.Side(face)
+	o, ok := w.Block(sidePos).(Bed)
+	return o, sidePos, ok
```
DaPigGuy, line 222: This should probably check that only one side is the head

```suggestion
	return o, sidePos, ok && b.Head != o.Head
```

server/player/player.go
```diff
@@ -1110,6 +1119,73 @@ func (p *Player) StopFlying() {
...
+// the player will not sleep.
+func (p *Player) Sleep(pos cube.Pos) {
+	ctx, sendReminder := event.C(), true
+	if p.Handler().HandleSleep(ctx, &sendReminder); ctx.Cancelled() {
+		return
+	}
+
+	w := p.World()
+	if b, ok := w.Block(pos).(block.Bed); ok {
+		if b.User != nil {
```
DaPigGuy, line 1132:
```suggestion
		if b.User != nil || !b.Head {
```

server/session/entity_metadata.go
```diff
@@ -95,6 +96,12 @@ func (s *Session) parseEntityMetadata(e world.Entity) protocol.EntityMetadata {
 	if sc, ok := e.(scoreTag); ok {
 		m[protocol.EntityDataKeyScore] = sc.ScoreTag()
 	}
+	if sl, ok := e.(sleeper); ok {
+		if pos, ok := sl.Sleeping(); ok {
+			m[protocol.EntityDataKeyBedPosition] = blockPosToProtocol(pos)
+			m.SetFlag(protocol.EntityDataKeyPlayerFlags, protocol.EntityDataFlagSleeping)
```
DaPigGuy, line 102: `protocol.EntityDataFlagSleeping` was wrongly used here when merging changes from master branch

server/world/world.go
```diff
@@ -733,6 +743,8 @@ func (w *World) RemoveEntity(e Entity) {
 	viewers := slices.Clone(c.v)
 	c.Unlock()
 
+	w.tryAdvanceDay()
```
DaPigGuy, line 746: This would bypass the 5 second timer if any entity were to be removed.

DaPigGuy (changes requested):

PR #608, opened 2022-07-28
Implemented Books and Quills + Written Book

abimek:
Complete


server/item/register.go
```diff
@@ -92,6 +92,8 @@ func init() {
 	world.RegisterItem(BlazeRod{})
 	world.RegisterItem(Bone{})
 	world.RegisterItem(Book{})
+	world.RegisterItem(WritableBook{})
+	world.RegisterItem(WrittenBook{})
```
TwistedAsylumMC, line 96: Would be nice if these were ordered. If you select all the world.RegisterItem(...) methods, go to `File > Edit > Sort Lines` then it will order them alphabetically

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
+package item
+
+type WritableBook struct {
+	Pages []string
+}
```
TwistedAsylumMC, line 5: Missing docs for the WiteableBook struct & Page field

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Page returns a specific page from the book. If the page exists, it will return the content and true, otherwise
+// it will return an empty string and false.
+func (w WritableBook) Page(page uint) (string, bool) {
```
TwistedAsylumMC, line 9: Is it neccessary to accept a uint? For API sake it will be better to accept a regular int and perform your own bound checks

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+	return w.Pages[p], true
+}
+
+// MaxCount always returns 1.
+func (w WritableBook) MaxCount() int {
+	return 1
+}
+
+// Exists checks to see weather a page exists or not
+func (w WritableBook) Exists(page uint) bool {
```
TwistedAsylumMC, line 23: Wrong spelling of whether, a name like PageExists() would also be more understandable. Same thing about the uint parameter

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return true
+}
+
+// Set writes a page to the book, if the page doesn't exist it will be created. It will panic if the
+// text is longer then 256 characters. It will return the newly updated pages.
+func (w WritableBook) Set(page uint, text string) []string {
+	if len(text) > 256 {
+		panic("text longer then 256 bytes")
+	}
```
TwistedAsylumMC, line 32: Logic here only allows 255 bytes, and the error message says the text is longer than 256 which would be incorrect if 256 bytes were provided

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	if len(pk.Text) > 256 {
+		return fmt.Errorf("text can not be longer than 256 bytes")
+	}
+	var pages []string
+	slot := int(pk.InventorySlot)
+	switch pk.ActionType {
+	case packet.BookActionReplacePage:
+		pages = book.Set(page, pk.Text)
+		break
```
JustTalDevelops: Go automatically breaks every case.

abimek: ight, will keep in mind

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+func (w WritableBook) Exists(page uint) bool {
+	if len(w.Pages) <= int(page) {
+		return false
+	}
+	return true
+}
+
+// Set writes a page to the book, if the page doesn't exist it will be created. It will panic if the
+// text is longer then 256 characters. It will return the newly updated pages.
+func (w WritableBook) Set(page uint, text string) []string {
```
TwistedAsylumMC, line 29: SetPage() might be a better name for the method, and same thing as before for the uint parameter. Is it also necessary to return the new set of pages?

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+	return pages
+}
+
+// Swap swaps two different pages, it will panic if the largest of the two numbers doesn't exist. it will
+// return the newly updated pages.
+func (w WritableBook) Swap(page1, page2 uint) []string {
+	pages := w.Pages
+	if w.Exists(max(page1, page2)) {
+		panic("invalid page number")
+	}
```
TwistedAsylumMC, line 49: This will cause more issues if even the smallest number does not exist

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+		copy(newPages, pages)
+		pages = newPages
+	}
+	pages[page] = text
+	return pages
+}
+
+// Swap swaps two different pages, it will panic if the largest of the two numbers doesn't exist. it will
+// return the newly updated pages.
+func (w WritableBook) Swap(page1, page2 uint) []string {
```
TwistedAsylumMC, line 45: Once again, a name like SwapPages() would be better and using int instead of uint for the pages

server/item/writable_book.go
```diff
@@ -0,0 +1,100 @@
...
+
+// max ...
+func max(a, b uint) uint {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func (w WritableBook) EncodeItem() (name string, meta int16) {
```
TwistedAsylumMC, line 95: Missing shadow doc

server/item/written_book.go
```diff
@@ -0,0 +1,62 @@
+package item
+
+type WrittenBook struct {
+	Title      string
+	Author     string
+	Generation uint
+	Glinted
+	Pages []string
+}
```
TwistedAsylumMC, line 9: The struct and all fields are missing docs

server/item/written_book.go
```diff
@@ -0,0 +1,62 @@
...
+
+// MaxCount always returns 1.
+func (w WrittenBook) MaxCount() int {
+	return 1
+}
+
+// Page returns a specific page from the book. If the page exists, it will return the content and true, otherwise
+// it will return an empty string and false.
+func (w WrittenBook) Page(page uint) (string, bool) {
```
TwistedAsylumMC, line 18: Accepting an int instead of a uint would be better for API sake

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,70 @@
+package session
+
+import (
+	"fmt"
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+	"golang.org/x/exp/slices"
+)
+
+type BookEditHandler struct{}
```
TwistedAsylumMC, line 10: Missing docs here

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,70 @@
...
+	it, err := s.inv.Item(int(pk.InventorySlot))
+	if err != nil {
+		return fmt.Errorf("invalid inventory slot index %v", pk.InventorySlot)
+	}
+	book, ok := it.Item().(item.WritableBook)
+	if !ok {
+		return fmt.Errorf("inventory slot %v does not contain a writable book", pk.InventorySlot)
+	}
+	// check page number and text beforehand to reduce repetition, shouldn't matter as the default values
+	// match these constraints
```
TwistedAsylumMC, line 24: Missing punctuation for comment

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,70 @@
...
+	page := uint(pk.PageNumber)
+	if page >= 50 {
+		return fmt.Errorf("page number %v is out of bounds", pk.PageNumber)
+	}
+	if len(pk.Text) > 256 {
+		return fmt.Errorf("text can not be longer than 256 bytes")
+	}
```
TwistedAsylumMC, line 31: This logic does not match the logic in (WriteableBook).Set()

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		pages = book.Swap(page, uint(pk.SecondaryPageNumber))
+		break
+	case packet.BookActionSign:
+		// Error does not need to be handled as it's confirmed at the begging that this slot contains a writable book.
+		s.inv.SetItem(slot, item.NewStack(item.WrittenBook{Title: pk.Title, Author: pk.Author, Pages: book.Pages, Generation: 0}, 1))
+		return nil
+	}
+	// Error does not need to be handled as it's confirmed at the begging that this slot contains a writable book.
+	s.inv.SetItem(slot, item.NewStack(item.WritableBook{Pages: pages}, 1))
```
TwistedAsylumMC, line 64: Might be better to fetch the pages here instead of having all the methods return the updates pages

TwistedAsylumMC (changes requested):

server/item/writable_book.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Exists checks to see weather a page exists or not
+func (w WritableBook) Exists(page uint) bool {
+	return len(w.Pages) > int(page)
+}
+
+// Set writes a page to the book, if the page doesn't exist it will be created. It will panic if the
+// text is longer then 256 characters. It will return the newly updated pages.
+func (w WritableBook) Set(page uint, text string) []string {
```
T14Raptor, line 29: Would just use an int here

server/item/writable_book.go
```diff
@@ -0,0 +1,97 @@
...
+		copy(newPages, pages)
+		pages = newPages
+	}
+	pages[page] = text
+	return pages
+}
+
+// Swap swaps two different pages, it will panic if the largest of the two numbers doesn't exist. it will
+// return the newly updated pages.
+func (w WritableBook) Swap(page1, page2 uint) []string {
```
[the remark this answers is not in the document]
T14Raptor: For both comments I made for the other method

server/item/written_book.go
```diff
@@ -0,0 +1,62 @@
+package item
+
+type WrittenBook struct {
+	Title      string
+	Author     string
+	Generation uint
+	Glinted
```
T14Raptor, line 7: Point for this?

server/item/writable_book.go
abimek: Logic here allows for 256, only panics once its above 256. Unless I'm somehow blanking on how the len() method works.

server/session/handler_book_edit.go
abimek: I'm pretty sure it does,
`
if len(text) > 256 {
	panic("text longer then 256 bytes")
}
`

server/session/handler_book_edit.go
abimek: Since the methods aren't pointer recievers I can't really update the type so all the methods need to return the pages. I can of course make them return new WritableBooks.

server/item/written_book.go
abimek: Removed the glinted field and documented everything else

server/item/writable_book.go
abimek: Yea, removed the uints.

server/session/handler_book_edit.go
abimek: Done

server/session/handler_book_edit.go
JustTalDevelops: This still isn’t resolved 😛

abimek: yes it is 😄

server/item/written_book.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// EncodeNBT ...
+func (w WrittenBook) EncodeNBT() map[string]any {
+	data := map[string]any{}
+	var pages []map[string]string
+	for _, page := range w.Pages {
+		pages = append(pages, map[string]string{
+			"text":      page,
+			"photoname": "",
```
JustTalDevelops, line 57: You don’t need to write this, the client will auto fill this as empty.

abimek: That's intresting, ok.

server/item/written_book.go
```diff
@@ -0,0 +1,70 @@
...
+// WrittenBook is an item created after a book and quill is signed. It appears the same as a regular book, but
+// without the quill, and has an enchanted-looking glint.
+type WrittenBook struct {
+	// Title is the title of the book
+	Title string
+	// Author is the author of the book
+	Author string
+	// Generation is the generation of the book. The copy tier of the book. 0 = original, 1 = copy of original,
+	// 2 = copy of copy.
+	Generation uint
```
JustTalDevelops, line 12: Would probably make a BookGeneration struct with specific options for each possible generation.

server/item/writable_book.go
```diff
@@ -0,0 +1,105 @@
+package item
+
+// A WritableBook is an item used to write written books.
```
JustTalDevelops, line 3: Would remove the A prefix for consistency.

server/item/writable_book.go
```diff
@@ -0,0 +1,105 @@
...
+	return w
+}
+
+// EncodeNBT ...
+func (w WritableBook) EncodeNBT() map[string]any {
+	if len(w.Pages) == 0 {
+		return nil
+	}
+	data := map[string]any{}
+	var pages []any
```
JustTalDevelops, line 85:
```suggestion
	pages := make([]any, 0, len(w.Pages))
```

server/item/writable_book.go
```diff
@@ -0,0 +1,105 @@
...
+	for _, page := range w.Pages {
+		pages = append(pages, map[string]any{
```
JustTalDevelops, line 87: Can probably inline this.

server/item/writable_book.go
```diff
@@ -0,0 +1,105 @@
...
+			"text": page,
+		})
+	}
+	data["pages"] = pages
```
JustTalDevelops, line 91: Can probably just remove `data` and return something like:
```go
return map[string]any{"pages": pages}

File: server/item/writable_book.go [hunk 26]
```diff
@@ -0,0 +1,105 @@
...
+				}
+			}
+		}
+	}
+	return w
+}
+
+// EncodeNBT ...
+func (w WritableBook) EncodeNBT() map[string]any {
+	if len(w.Pages) == 0 {
```
JustTalDevelops, line 81: This check is incorrect - NBT is written anyway.

server/item/writable_book.go
abimek: how so

server/item/writable_book.go
abimek: Don't believe so, the wiki says that pages may not exist. Should I return an empty map?

server/session/handler_book_edit.go
TwistedAsylumMC: Yeah i think that's the better choice, it's more consistent with other features

server/item/writable_book.go
TwistedAsylumMC: Yeah ignore this, I misread the code

server/item/writable_book.go
JustTalDevelops: Don't trust the wiki! 😛

JustTalDevelops: The client should autofill all missing NBT - you can try this yourself.

TwistedAsylumMC (approved):

server/item/written_book.go
```diff
@@ -0,0 +1,81 @@
...
+			panic("unknown written book generation")
+		}
+	}
+	return w
+}
+
+// EncodeNBT ...
+func (w WrittenBook) EncodeNBT() map[string]any {
+	data := map[string]any{}
+	var pages []map[string]string
```
JustTalDevelops, line 65:
```suggestion
	pages := make([]any, 0, len(w.Pages))
```

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+	"golang.org/x/exp/slices"
+)
+
+// BookEditHandler handles the BookEdit Packet
+type BookEditHandler struct{}
+
+// Handle ...
+func (b BookEditHandler) Handle(p packet.Packet, s *Session) error {
```
JustTalDevelops, line 14: Could maybe space this out a bit more.

server/item/written_book.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return w
+}
+
+// EncodeNBT ...
+func (w WrittenBook) EncodeNBT() map[string]any {
+	data := map[string]any{}
+	pages := make([]any, 0, len(w.Pages))
+	for _, page := range w.Pages {
+		pages = append(pages, map[string]string{
```
JustTalDevelops, line 67: Would inline this.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,65 @@
...
```
JustTalDevelops, line 14: Make sure you look this logic over; just tested for a few minutes in game and got disconnected for:
```
failed processing packet from 127.0.0.1:54647 (JustTalz): *packet.BookEdit: may only come before a page which already exists
```

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,65 @@
...
+	case packet.BookActionSwapPages:
+		if pk.SecondaryPageNumber >= 50 {
+			return fmt.Errorf("page number out of bounds")
+		}
+		if !book.ValidPage(page) || !book.ValidPage(int(pk.SecondaryPageNumber)) {
+			return fmt.Errorf("page number %v does not exist", pk.PageNumber)
+		}
+		book = book.SwapPages(page, int(pk.SecondaryPageNumber))
+	case packet.BookActionSign:
+		_ = s.inv.SetItem(slot, item.NewStack(item.WrittenBook{Title: pk.Title, Author: pk.Author, Pages: book.Pages, Generation: item.OriginalGeneration()}, 1))
```
JustTalDevelops, line 60: There definitely seems to be some sort of mismatch with the client & server when signing, got a lot of debug logs on different items than the server.

server/item/book_and_quill.go
```diff
@@ -0,0 +1,98 @@
...
+	if len(w.Pages) <= page {
+		return "", false
+	}
+	return w.Pages[page], true
+}
+
+// SetPage writes a page to the book, if the page doesn't exist it will be created. It will panic if the
+// text is longer then 256 characters. It will return a new book representing this data.
+func (w BookAndQuill) SetPage(page int, text string) BookAndQuill {
+	if page < 0 {
```
JustTalDevelops, line 33: Shouldn't this also ensure `!page > 50`?

server/item/book_and_quill.go
```diff
@@ -0,0 +1,98 @@
...
+	if pageOne < 0 || pageTwo < 0 {
+		panic("negative page number")
+	}
+	if !w.ValidPage(max(pageOne, pageTwo)) {
+		panic("invalid page number")
+	}
+	contentOne := w.Pages[pageOne]
+	contentTwo := w.Pages[pageTwo]
+	w.Pages[pageOne] = contentTwo
+	w.Pages[pageTwo] = contentOne
```
DaPigGuy, line 60:
```suggestion
	temp := w.Pages[pageOne]
	w.Pages[pageOne] = w.Pages[pageTwo] 
	w.Pages[pageTwo] = temp
```

and frankly variable names `a`/ `b` rather than the longer `pageOne`/`pageTwo` are preferred in this specific context with the name & docs already providing enough information

server/item/written_book.go
```diff
@@ -0,0 +1,78 @@
+package item
+
+// WrittenBook is an item created after a book and quill is signed. It appears the same as a regular book, but
+// without the quill, and has an enchanted-looking glint.
+type WrittenBook struct {
+	// Title is the title of the book
```
DaPigGuy, line 6: Include punctuation in these docs

server/item/written_book.go
```diff
@@ -0,0 +1,78 @@
...
+	Title string
+	// Author is the author of the book
+	Author string
+	// Generation is the generation of the book. The copy tier of the book. 0 = original, 1 = copy of original,
```
DaPigGuy, line 10:
```suggestion
	// Generation is the copy tier of the book. 0 = original, 1 = copy of original,
```
Sentence fragment!

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,74 @@
+package session
+
+import (
+	"fmt"
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+	"golang.org/x/exp/slices"
+)
+
+// BookEditHandler handles the BookEdit Packet
```
DaPigGuy, line 10:
```suggestion
// BookEditHandler handles the BookEdit packet.
```

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,74 @@
...
+		if !book.ValidPage(page) {
+			return fmt.Errorf("unable to insert page at %v", pk.PageNumber)
+		}
+		book = item.BookAndQuill{Pages: slices.Insert(book.Pages, page, pk.Text)}
+	case packet.BookActionDeletePage:
+		if page >= len(book.Pages) && page <= len(book.Pages)+2 {
+			if !book.ValidPage(page) {
+				break
+			}
+		}
```
DaPigGuy, line 55:
```suggestion
```

This doesn't seem necessary given the next line calls `ValidPage` which checks bounds.

DaPigGuy (changes requested):

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,74 @@
...
+		if !book.ValidPage(page) {
+			return fmt.Errorf("page number %v does not exist", pk.PageNumber)
+		}
+		book = item.BookAndQuill{Pages: slices.Delete(book.Pages, page, page+1)}
+	case packet.BookActionSwapPages:
+		if pk.SecondaryPageNumber >= 50 {
+			return fmt.Errorf("page number out of bounds")
+		}
+		if !book.ValidPage(page) || !book.ValidPage(int(pk.SecondaryPageNumber)) {
+			break
```
DaPigGuy, line 65: This should probably return an error

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,74 @@
...
+
+	slot := int(pk.InventorySlot)
+	switch pk.ActionType {
+	case packet.BookActionReplacePage:
+		book = book.SetPage(page, pk.Text)
+	case packet.BookActionAddPage:
+		if len(book.Pages) >= 50 {
+			return fmt.Errorf("unable to add page beyond 50")
+		}
+		if page >= len(book.Pages) && page <= len(book.Pages)+2 {
```
DaPigGuy, line 42: I have not gotten to testing yet but I assume this is meant for inserting a page at the very end in which case this should probably be simplified to:

```suggestion
		if page == len(book.Pages) {
```

server/item/book_and_quill.go
```diff
@@ -0,0 +1,98 @@
...
+	Pages []string
+}
+
+// MaxCount always returns 1.
+func (w BookAndQuill) MaxCount() int {
+	return 1
+}
+
+// ValidPage checks to see whether a page exists or not.
+func (w BookAndQuill) ValidPage(page int) bool {
```
T14Raptor, line 15: Is this method necessary if `(BookAndQuill).Page` already returns a bool?

server/session/handler_book_edit.go
DaPigGuy: Completely disregard this

server/item/book_and_quill.go
```diff
@@ -0,0 +1,98 @@
...
+	return page >= 0 && len(w.Pages) > page
+}
+
+// Page returns a specific page from the book. If the page does not exist
+func (w BookAndQuill) Page(page int) (string, bool) {
+	if page < 0 {
```
JustTalDevelops, line 21: Can probably inline this with the check below and always just return false.

server/session/handler_book_edit.go
abimek: I break due to the client sometimes being able too see two pages ahead of the current total pages. This means it trys to swap pages that don't exist so it's simpler to just break whenever an invalid page swap is attempted. (Hopefully this makes sense)

JustTalDevelops: Would probably document that then.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,76 @@
...
+			book = book.SetPage(page, "")
+			break
+		}
+		if _, ok := book.Page(page); !ok {
+			return fmt.Errorf("unable to insert page at %v", pk.PageNumber)
+		}
+		book = item.BookAndQuill{Pages: slices.Insert(book.Pages, page, pk.Text)}
+	case packet.BookActionDeletePage:
+		// We break here instead of returning an error because the client can be a page or two ahead in the UI then
```
JustTalDevelops, line 51: Would probably put this doc right above where you actually break, for consistency.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,76 @@
...
+		// representation server side. The client still sends the swap indexes.
+		if !ok || !ok2 {
+			break
+		}
+		book = book.SwapPages(page, int(pk.SecondaryPageNumber))
+	case packet.BookActionSign:
+		_ = s.inv.SetItem(slot, item.NewStack(item.WrittenBook{Title: pk.Title, Author: pk.Author, Pages: book.Pages, Generation: item.OriginalGeneration()}, 1))
+		return nil
+	}
+	_ = s.inv.SetItem(slot, item.NewStack(book, 1))
```
JustTalDevelops, line 74: This will cause custom data such as custom names on the item to be lost, since you're initializing a new stack.

JustTalDevelops: You can probably use the `duplicateStack` function here.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,76 @@
...
+			return fmt.Errorf("unable to add page beyond 50")
+		}
+		if page >= len(book.Pages) && page <= len(book.Pages)+2 {
+			book = book.SetPage(page, "")
+			break
+		}
+		if _, ok := book.Page(page); !ok {
+			return fmt.Errorf("unable to insert page at %v", pk.PageNumber)
+		}
+		book = item.BookAndQuill{Pages: slices.Insert(book.Pages, page, pk.Text)}
```
JustTalDevelops, line 49: I think it's best to add a function for adding a new page directly to BookAndQuill.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,76 @@
...
+	case packet.BookActionDeletePage:
+		if _, ok := book.Page(page); !ok {
+			// We break here instead of returning an error because the client can be a page or two ahead in the UI then
+			// the actual pages representation server side. The client still sends the deletion indexes.
+			break
+		}
+		book = item.BookAndQuill{Pages: slices.Delete(book.Pages, page, page+1)}
```
JustTalDevelops, line 56: Same here.

DaPigGuy (approved):

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,75 @@
...
+			break
+		}
+		if _, ok := book.Page(page); !ok {
+			return fmt.Errorf("unable to insert page at %v", pk.PageNumber)
+		}
+		book = book.InsertPage(page, pk.Text)
+	case packet.BookActionDeletePage:
+		if _, ok := book.Page(page); !ok {
+			// We break here instead of returning an error because the client can be a page or two ahead in the UI then
+			// the actual pages representation server side. The client still sends the deletion indexes.
```
Sandertv, line 52: Might want to send a debug message when this happens.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,75 @@
...
+	case packet.BookActionSwapPages:
+		if pk.SecondaryPageNumber >= 50 {
+			return fmt.Errorf("page number out of bounds")
+		}
+		_, ok := book.Page(page)
+		_, ok2 := book.Page(int(pk.SecondaryPageNumber))
+		if !ok || !ok2 {
+			// We break here instead of returning an error because the client can try to swap pages that don't exist.
+			// This happens as a result of the client being a page or two ahead in the UI then the actual pages
+			// representation server side. The client still sends the swap indexes.
```
Sandertv, line 65: Would send a debug message here too.

server/session/handler_book_edit.go
```diff
@@ -0,0 +1,75 @@
...
+			break
+		}
+		book = book.SwapPages(page, int(pk.SecondaryPageNumber))
+	case packet.BookActionSign:
+		_ = s.inv.SetItem(slot, item.NewStack(item.WrittenBook{Title: pk.Title, Author: pk.Author, Pages: book.Pages, Generation: item.OriginalGeneration()}, 1))
```
Sandertv, line 70: Are custom name and lore supposed to be lost when signing the book? If not, you'd probably want to use `duplicateStack` here too.

server/item/written_book.go
```diff
@@ -0,0 +1,75 @@
...
+func (w WrittenBook) DecodeNBT(data map[string]any) any {
+	if pages, ok := data["pages"].([]any); ok {
+		w.Pages = make([]string, len(pages))
+		for i, page := range pages {
+			w.Pages[i] = page.(map[string]any)["text"].(string)
+		}
+	}
+	if v, ok := data["title"].(string); ok {
+		w.Title = v
+	}
```
Sandertv, line 41: Can change this to `w.Title, _ = data["title"].(string)`. Same for the author below.

server/item/book_and_quill.go
```diff
@@ -0,0 +1,119 @@
...
+func (b BookAndQuill) DecodeNBT(data map[string]any) any {
+	if pages, ok := data["pages"].([]any); ok {
+		for _, page := range pages {
+			if pageData, ok := page.(map[string]any); ok {
+				if text, ok := pageData["text"].(string); ok {
+					b.Pages = append(b.Pages, text)
+				}
+			}
+		}
+	}
```
Sandertv, line 95: Can change this to `pages, _ := data["pages"].([]any)`, then `for _, page := range pages {` directly after that. The if check is unnecessary.

Sandertv (changes requested):
Looks really good! I've got a couple more comments. Once these are resolved, this is probably ready to go. Nice work.

server/session/handler_book_edit.go
JustTalDevelops: This is expected to happen though - the client does this fairly often. This would essentially just clutter console with a lot of unnecessary debug.

server/session/handler_book_edit.go
JustTalDevelops: Same here.

server/session/handler_book_edit.go
Sandertv: If it doesn't otherwise occasionally cause valid actions to be cancelled, feel free not to debug. Same counts for the other places.

server/session/handler_book_edit.go
JustTalDevelops: This is the case. The client actually prefers the custom name over the title in the packet, if the custom name exists, so I would account for that too.

server/session/handler_book_edit.go
DaPigGuy: This occurs when trying to swap the last two pages while the server is not yet aware of their existence (the last two pages aren't set until input, and won't exist upon reopening in the vanilla implementation). Basically this can and will occur quite often especially since the swap button can be spammed.

PR #621, opened 2022-08-01
item/fire_charge.go: Implement the fire charge item

AndreasHGK:


server/item/fire_charge.go
```diff
@@ -0,0 +1,34 @@
...
+type FireCharge struct{}
+
+// EncodeItem ...
+func (f FireCharge) EncodeItem() (name string, meta int16) {
+	return "minecraft:fire_charge", 0
+}
+
+// UseOnBlock ...
+func (f FireCharge) UseOnBlock(pos cube.Pos, face cube.Face, _ mgl64.Vec3, w *world.World, _ User, ctx *UseContext) bool {
+	ctx.SubtractFromCount(1)
```
Sandertv, line 23: Although this does work correctly, I would encourage you to only call `ctx.SubtractFromCount(1)` in the case where you `return true` to avoid confusion in the future.

server/world/sound/item.go
```diff
@@ -58,3 +58,7 @@ type GoatHorn struct {
 
 	sound
 }
+
+// UseFireCharge is a sound played when a player lights a block on fire with a fire charge, or when a dispenser or a
+// blaze shoots a fireball.
+type UseFireCharge struct{ sound }
```
Sandertv, line 64: Because this sound is also played when a blaze shoots a fireball, I think it might be better to just name this something like `FireCharge`.

Sandertv (changes requested):
Thanks for the PR! I've got two comments, should be easily resolved. I think this looks good otherwise!

Sandertv (approved):
This looks good to me. Thanks for the PR!

PR #627, opened 2022-08-02
Implement recovery compasses

JustTalDevelops:


server/player/player.go
```diff
@@ -89,6 +89,11 @@ type Player struct {
 	lastXPPickup atomic.Value[time.Time]
 	immunity     atomic.Value[time.Time]
 
+	deathMu        sync.Mutex
+	hasDied        bool
+	deathPos       mgl64.Vec3
+	deathDimension world.Dimension
```
Sandertv, line 95: Would just make one of these fields a pointer so you can remove the `deathPos` field.

Sandertv (commented):
Looks good. I've got one small nitpick to get rid of one of the fields, but this should be good to go afterwards.

server/player/player.go
JustTalDevelops: Did you mean the `hasDied` field?

Sandertv: I did, sorry!

JustTalDevelops: Removed.

Sandertv (approved):
LGTM! Thanks for the PR!

PR #634, opened 2022-08-09
Separate network IDs from save IDs

Xd-pro:
Allows multiple entity types to have the same network ID.


server/world/entity.go
```diff
@@ -41,6 +42,12 @@ type TickerEntity interface {
 	Tick(w *World, current int64)
 }
 
+type NetworkedEntity interface {
```
DaPigGuy, line 45: Missing a doc here

server/world/entity.go
```diff
@@ -41,6 +42,12 @@ type TickerEntity interface {
...
+	Entity
+	// Returns the network type ID of the entity. This is NOT the save ID.
```
DaPigGuy, line 47: This doc should begin with `// NetworkTypeID`

server/world/entity.go
```diff
@@ -41,6 +42,13 @@ type TickerEntity interface {
 	Tick(w *World, current int64)
 }
 
+// NetworkEncodeableEntity is an Entity where the save ID and network ID is not the same.
```
DaPigGuy, line 45:
```suggestion
// NetworkEncodeableEntity is an Entity where the save ID and network ID are not the same.
```

server/world/entity.go
```diff
@@ -41,6 +42,13 @@ type TickerEntity interface {
...
+type NetworkEncodeableEntity interface {
+	Entity
```
Sandertv, line 47: No reason to have this embed `Entity`.

server/entity/text.go
```diff
@@ -24,11 +24,16 @@ func (t *Text) Name() string {
...
 }
 
-// EncodeEntity returns the ID for falling blocks.
+// EncodeEntity returns the ID for text.
 func (t *Text) EncodeEntity() string {
 	return "dragonfly:text"
 }
 
+// NetworkTypeID returns the network type ID for falling blocks.
+func (*Text) NetworkTypeID() string {
```
Sandertv, line 33: This method should be called `NetworkEncodeEntity`.

server/world/entity.go
```diff
@@ -41,6 +42,13 @@ type TickerEntity interface {
 	Tick(w *World, current int64)
 }
 
+// NetworkEncodeableEntity is an Entity where the save ID and network ID is not the same.
+type NetworkEncodeableEntity interface {
```
Sandertv, line 46: I'd consider moving this to the session package, as that's the only place where the interface is used.

Sandertv (changes requested):
I've left a couple of comments. I think this is ready to go once those and the ones left by @DaPigGuy are resolved.

server/world/entity.go
```diff
@@ -41,6 +42,8 @@ type TickerEntity interface {
 	Tick(w *World, current int64)
 }
 
+
+
```
DaPigGuy, line 46: Please revert all changes to this file

server/entity/text.go
DaPigGuy, line 33:
```suggestion
func (*Text) NetworkEncodeEntity() string {
```

Sandertv (approved):

PR #642, opened 2022-08-15
Add a bunch of particles

Prim69:
This pr adds the following particles: Dust, EntityFlame, LavaDrip, WaterDrip, and Lava


Prim69:
image
This is the lava particle but i called it combustion, pm calls it lava and has the network constant as lava, not sure if that should be the name, and also it needs a doc but idk what to do for that

T14Raptor:
Thanks!

PR #643, opened 2022-08-16
Implemented shields

JustTalDevelops:


server/player/player.go
```diff
@@ -545,9 +547,54 @@ func (p *Player) Hurt(dmg float64, source damage.Source) (float64, bool) {
 		return 0, true
 	}
 
+	w, pos := p.World(), p.Position()
 	totalDamage := p.FinalDamageFrom(dmg, source)
-	damageLeft := totalDamage
+	if ok, _ := p.Blocking(); ok && source.ReducedByArmour() {
+		affected := true
+		if src, ok := source.(damage.SourceEntityAttack); ok {
+			diff := p.Position().Sub(src.Attacker.Position()).Normalize()
```
T14Raptor, line 555: Any reason to normalize this?

JustTalDevelops: I don't think so (was taken from Java), removed.

server/entity/arrow.go
```diff
@@ -191,6 +191,11 @@ func (a *Arrow) Rotation() (float64, float64) {
 	return a.yaw, a.pitch
 }
 
+// blocker represents an entity that can block attacks with a shield.
+type blocker interface {
+	Blocking() (bool, bool)
```
Sandertv, line 196: I would document this method and probably add named return types here, because it's unclear what this should return.

server/player/player.go
```diff
@@ -545,9 +547,46 @@ func (p *Player) Hurt(dmg float64, source damage.Source) (float64, bool) {
...
+				held, other := p.HeldItems()
+				if _, ok := held.Item().(item.Shield); ok {
+					p.SetHeldItems(p.damageItem(held, i), other)
+				} else {
+					p.SetHeldItems(held, p.damageItem(other, i))
+				}
+			}
+			return 0, false
+		}
+	}
```
Sandertv, line 587: Could you pull this out in its own method and add some documentation (if and where relevant)? This is a massive block of code and Player.Hurt() is already a pretty big function as it is.

server/player/player.go
```diff
@@ -1164,6 +1214,11 @@ func (p *Player) OnFireDuration() time.Duration {
 	return time.Duration(p.fireTicks.Load()) * time.Second / 20
 }
 
+// BlockingDelay ...
```
Sandertv, line 1217: Would document this. What does it do?

server/player/player.go
```diff
@@ -1174,6 +1229,15 @@ func (p *Player) SetOnFire(duration time.Duration) {
 	p.updateState()
 }
 
+// SetBlockingDelay ...
```
Sandertv, line 1232: Would document this too.

server/player/player.go
```diff
@@ -1079,6 +1117,18 @@ func (p *Player) StopFlying() {
 	p.session().SendGameMode(p.GameMode())
 }
 
+// Blocking returns true if the player is currently blocking with a shield. The first boolean is true if the player was
+// holding a shield. The second boolean is true if the player was performing the necessary actions in order to block.
```
Sandertv, line 1121: Documentation does not seem to match the code. Blocking's first bool is only true if holding a shield and performing the necessary actions to block.

ImLighty:
Anything about this?

server/entity/arrow.go
RestartFU: note: resolved

server/player/player.go
RestartFU: any method name suggestions?

RestartFU: this has been resolved, once the conflicts are resolved, are we ready to merge?

PR #644, opened 2022-08-16
Add some extra methods to lightning

Prim69:
This pr adds the following methods to the lightning entity: WithoutDamage, WithoutBlockFire, WithoutEntityFire


server/entity/lightning.go
```diff
@@ -70,8 +79,38 @@ func (li *Lightning) New(pos mgl64.Vec3) world.Entity {
...
+// WithoutDamage returns the lightning entity with damage disabled, it won't damage any entities.
+func (li *Lightning) WithoutDamage() *Lightning {
+	li.mu.Lock()
+	li.damage = false
+	li.mu.Unlock()
+	return li
+}
+
+// WithoutBlockFire returns the lightning entity but won't set any blocks on fire.
+func (li *Lightning) WithoutBlockFire() *Lightning {
```
Sandertv, line 91: I think it might be better to add additional `New*` methods for these, like `NewLightningWithDamage` and maybe a bool in that to disable fire.

server/entity/lightning.go
```diff
@@ -70,8 +79,38 @@ func (li *Lightning) New(pos mgl64.Vec3) world.Entity {
...
+	li.mu.Lock()
+	li.blockFire = false
+	li.mu.Unlock()
+	return li
+}
+
+// WithoutEntityFire returns the lightning entity but won't set any entities on fire.
+func (li *Lightning) WithoutEntityFire() *Lightning {
```
Sandertv, line 99: Does it ever make sense to not spawn fire blocks but still set entities on fire?

Sandertv (changes requested):

server/entity/lightning.go
Prim69: I was thinking of lighting the entity on fire but not the block, or lighting none on fire

server/entity/lightning.go
Prim69: The fire bool would have to be passed to the default NewLightning too, so it would make more sense to just pass everything to NewLightning, but that looks weird so from what i see this is the best solution

Sandertv: Just choose a default for `NewLightning` and only accept a fire bool in `NewLightningWithDamage`.

Prim69:
anything else?

Sandertv (approved):
This looks good, thank you! The main reason I suggested adding a `NewLightningWithDamage` was to avoid the need for locking all those fields with a mutex by the way.

PR #645, opened 2022-08-17
Implemented bells

JustTalDevelops:


JustTalDevelops:
This should also be ready for review now.

server/block/block.go
```diff
@@ -73,6 +73,18 @@ type EntityInsider interface {
 	EntityInside(pos cube.Pos, w *world.World, e world.Entity)
 }
 
+// ProjectileHitter represents a block that handles being hit by a projectile.
+type ProjectileHitter interface {
+	// ProjectileHit is called when a projectile hits the block.
+	ProjectileHit(w *world.World, e world.Entity, pos cube.Pos, face cube.Face)
+}
+
+// EntityEjector represents a block that ejects entities when they are inside it.
```
DaPigGuy, line 82: This doesn't seem accurate as it's the item entity that's implementing this...

PR #651, opened 2022-08-20
Implemented lecterns

JustTalDevelops:


DaPigGuy (approved):

server/block/lectern.go
```diff
@@ -0,0 +1,163 @@
...
+	held, _ := u.HeldItems()
+	if _, ok := held.Item().(readableBook); ok {
+		l.Page = 0
+		l.Book = held
+
+		w.SetBlock(pos, l, nil)
+		w.PlaySound(pos.Vec3Centre(), sound.LecternBookPlace{})
+		ctx.SubtractFromCount(1)
+	}
+	return true
```
Sandertv, line 89: You should probably return false if item is not a book or if the Lectern already held a book (first check).

server/block/lectern.go
```diff
@@ -0,0 +1,163 @@
...
+
+	l.Book = item.Stack{}
+	w.SetBlock(pos, l, nil)
+	w.PlaySound(pos.Vec3Centre(), sound.Attack{})
+}
+
+// TurnPage updates the page the lectern is currently on to the page given.
+func (l Lectern) TurnPage(pos cube.Pos, w *world.World, page int) error {
+	if page == l.Page {
+		// We're already on the correct page, so ignore this packet.
```
Sandertv, line 109: Wouldn't say "ignore this packet" here because there's no packet at this stage.

server/block/model/lectern.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// Lectern is a model used by lecterns.
+type Lectern struct{}
+
+// BBox ...
+func (Lectern) BBox(cube.Pos, *world.World) []cube.BBox {
+	return []cube.BBox{full.ExtendTowards(cube.FaceDown, 0.1)}
```
Sandertv, line 13: This isn't very clear. I would use the `cube.Box()` function instead.

server/player/player.go
```diff
@@ -2747,9 +2747,9 @@ func (p *Player) addNewItem(ctx *item.UseContext) {
 	}
 }
 
-// canReach checks if a player can reach a position with its current range. The range depends on if the player
+// CanReach checks if a player can reach a position with its current range. The range depends on if the player
 // is either survival or creative mode.
-func (p *Player) canReach(pos mgl64.Vec3) bool {
+func (p *Player) CanReach(pos mgl64.Vec3) bool {
```
Sandertv, line 2752: I'm hesitant on making this an exported function. I think it'd be better to keep this unexported and do this check in the EditSign method so we can keep this as it was.

server/session/controllable.go
```diff
@@ -42,6 +42,7 @@ type Controllable interface {
 	ReleaseItem()
 	UseItemOnBlock(pos cube.Pos, face cube.Face, clickPos mgl64.Vec3)
 	UseItemOnEntity(e world.Entity) bool
+	CanReach(pos mgl64.Vec3) bool
```
Sandertv, line 45: As mentioned above, would undo this.

server/session/handler_lectern_update.go
```diff
@@ -0,0 +1,29 @@
...
+	pos := blockPosFromProtocol(pk.Position)
+	if !s.c.CanReach(pos.Vec3Middle()) {
+		return fmt.Errorf("block at %v is not within reach", pos)
+	}
+	w := s.c.World()
+	lectern, ok := w.Block(pos).(block.Lectern)
+	if !ok {
+		return fmt.Errorf("block at %v is not a lectern", pos)
+	}
+	return lectern.TurnPage(pos, w, int(pk.Page))
```
Sandertv, line 28: Would perform this code in some `(*Player).EditLectern()` method (or a better method if you feel like it, maybe something that's more clear about the fact it's only turning pages).

server/internal/nbtconv/mapread.go
```diff
@@ -78,11 +78,12 @@ func MapBlock(x map[string]any, k string) world.Block {
 // to a world.Item.
 func MapItem(x map[string]any, k string) item.Stack {
 	if m, ok := x[k].(map[string]any); ok {
-		s := readItemStack(m)
-		readDamage(m, &s, true)
-		readEnchantments(m, &s)
-		readDisplay(m, &s)
-		readDragonflyData(m, &s)
+		t := Map[map[string]any](m, "tag")
```
Sandertv, line 81: Is this correct for every item in Bedrock Edition? I feel it's weird we didn't have this before, so we need to make sure this matches Bedrock Edition.

Sandertv (changes requested):
Looks fine mostly. I'm only not sure about the NBT "tag" changes. What are those based on?

JustTalDevelops:
The NBT changes are based on vanilla Minecraft. Everything that does not define the item type goes inside the tag field.

JustTalDevelops:
This should be ready for review now.

server/player/player.go
```diff
@@ -2649,14 +2649,34 @@ func (p *Player) EditSign(pos cube.Pos, text string) error {
...
 		return nil
 	}
+
 	sign.Text = text
 	w.SetBlock(pos, sign, nil)
 	return nil
 }
 
+// TurnLecternPage edits the lectern at the cube.Pos passed by turning the page to the page passed. If no lectern is
+// present or if the Player cannot edit it, an error is returned.
```
DaPigGuy, line 2662: > or if the Player cannot edit it, an error is returned.

doesn't seem to be the case

DaPigGuy (approved):

PR #656, opened 2022-08-25
Implement partial render_offsets (scale) component

alvin0319:
With this PR you can correct the item scale. (It's more like modifying item render position. But there is a component which can set the position/rotation of custom item)

(Originally from my [CustomItemLoader](https://github.com/alvin0319/CustomItemLoader/blob/master/src/alvin0319/CustomItemLoader/item/properties/CustomItemProperties.php#L674-#L708) for PocketMine-MP)

The following images are showing the difference between non-render_offsets item and render_offsets item.

(This item has 32x32 width/height.)

Before this PR:
Minecraft 2022-08-25 오후 10_55_08

After this PR:
Minecraft 2022-08-25 오후 10_55_37

### DIFF
--- a/server/internal/iteminternal/components.go
+++ b/server/internal/iteminternal/components.go
@@ -108,7 +108,7 @@ func Components(it world.CustomItem) map[string]any {
 	return builder.Construct()
 }
 
-// calculateHandPosition calculates the position of the item to be rendered to the player according to the given size.
+// calculateHandPosition calculates the scale of the item to be rendered to the player according to the given size.
 func calculateHandPosition(it world.CustomItem, size float32) (float32, float32, float32) {
 	var x, y, z float32
 	if _, ok := it.(item.HandEquipped); ok {
--- a/server/internal/iteminternal/components.go
+++ b/server/internal/iteminternal/components.go
@@ -81,13 +81,7 @@ func Components(it world.CustomItem) map[string]any {
 	if x, ok := it.(item.HandEquipped); ok {
 		builder.AddProperty("hand_equipped", x.HandEquipped())
 	}
-
-	bounds := it.Texture().Bounds()
-
-	x, y, z := calculateItemScale(it, float32(bounds.Dx()), float32(bounds.Dy()))
-
-	itemScale := []float32{x, y, z}
-
+	itemScale := calculateItemScale(it)
 	builder.AddComponent("minecraft:render_offsets", map[string]any{
 		"main_hand": map[string]any{
 			"first_person": map[string]any{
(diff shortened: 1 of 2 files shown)


server/internal/iteminternal/components.go
```diff
@@ -81,5 +81,43 @@ func Components(it world.CustomItem) map[string]any {
 	if x, ok := it.(item.HandEquipped); ok {
 		builder.AddProperty("hand_equipped", x.HandEquipped())
 	}
+
+	x, y, z := calculateItemScale(it, float32(it.ImageSize()))
+
+	value := []float32{x, y, z}
```
TwistedAsylumMC, line 87: You may as well just make calculateItemScale return the size as a slice already, and also give a better name to value (e.g. itemScale)

server/internal/iteminternal/components.go
```diff
@@ -81,5 +81,43 @@ func Components(it world.CustomItem) map[string]any {
...
 }
+
+// calculateItemScale calculates the scale of the item to be rendered to the player according to the given size.
+func calculateItemScale(it world.CustomItem, size float32) (float32, float32, float32) {
+	var x, y, z float32
+	if _, ok := it.(item.HandEquipped); ok {
+		x, y, z = 0.075, 0.125, 0.075
+	} else {
+		x, y, z = 0.1, 0.1, 0.1
+	}
```
TwistedAsylumMC, line 118: You can default the xyz properties to use 0.1 instead of having the else field

server/world/item.go
```diff
@@ -26,6 +26,9 @@ type CustomItem interface {
 	Texture() image.Image
 	// Category is the category the item will be listed under in the creative inventory.
 	Category() category.Category
+	// ImageSize is the width/height size value of the image which will be used to correct item render scale. We assume width and height are same.
+	// Minecraft uses 16x16 for their image. If your image uses 16x16 then you could return 16.
+	ImageSize() uint8
```
TwistedAsylumMC, line 31: This isn't the most reliable method, you can calculate the size of the texture using the Bounds() method on the image.Image interface

TwistedAsylumMC (changes requested):
Thanks for this, it's something we've been needing for a while. Just a few things I think you should change

server/internal/iteminternal/components.go
```diff
@@ -81,5 +81,39 @@ func Components(it world.CustomItem) map[string]any {
...
+	height := float32(it.Texture().Bounds().Dy())
+	var x, y, z float32 = 0.1, 0.1, 0.1
+	if _, ok := it.(item.HandEquipped); ok {
+		x, y, z = 0.075, 0.125, 0.075
+	}
+	newX := x / (width / 16)
+	newY := y / (height / 16)
+	newZ := z / (width / 16)
+	return []float32{newX, newY, newZ}
+}
```
TwistedAsylumMC, line 119: Could you run go fmt over this file? Looks good after that

Sandertv (approved):
Thanks for the PR! Looks good.

PR #664, opened 2022-08-31
Implemented suspicious stew

eminarican:


server/item/suspicious_stew.go
```diff
@@ -0,0 +1,34 @@
+package item
+
+import (
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// SuspiciousStew is a food item that can give the player a status effect that depends on the flower used to craft it.
+type SuspiciousStew struct {
+	defaultFood
```
JustTalDevelops, line 9: Would probably space this out and then comment Effect, for consistency.

server/item/suspicious_stew.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// AlwaysConsumable ...
+func (SuspiciousStew) AlwaysConsumable() bool {
+	return true
+}
+
+// EncodeItem ...
+func (s SuspiciousStew) EncodeItem() (name string, meta int16) {
+	return "suspicious_stew", int16(s.Effect.Uint8())
```
JustTalDevelops, line 25: Make sure you prefix this with `minecraft:`.

server/item/suspicious_stew_type.go
```diff
@@ -0,0 +1,93 @@
+package item
+
+import (
+	"time"
+
+	"github.com/df-mc/dragonfly/server/entity/effect"
+)
+
+type stewEffect uint8
```
JustTalDevelops, line 9: Would probably move this to the bottom of the file for consistency.

JustTalDevelops (changes requested):
Thanks for the PR! Just a couple comments.

server/item/suspicious_stew_type.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// StewEffect represents a type of suspicious stew.
+type StewEffect struct {
+	effect effect.Effect
+	stewEffect
+}
+
+// Type returns suspicious stew effect type.
+func (s StewEffect) Type() effect.Effect {
```
JustTalDevelops, line 23: There's a lot of empty lines in this function and the functions below this one, I'd probably remove them.

JustTalDevelops: Also, I'd probably make this a function for `stewEffect` and then switch on the type and then return the respective effects there. I'd also make it return a slice of effects as opposed to a single effect.

eminarican: i thought it could only have an effect at a time

JustTalDevelops: Not necessarily, but I imagine it might in the future, so there's no harm in supporting that functionality.

JustTalDevelops: I'd also rename `StewEffect` to just `SuspiciousStewType` or `StewType` if you do decide to support multiple effects.

eminarican: gotcha 👍

server/item/suspicious_stew_type.go
```diff
@@ -0,0 +1,107 @@
...
+	case 6:
+		effects = append(effects, effect.New(effect.Saturation{}, 1, time.Second*3/10))
+	case 7:
+		effects = append(effects, effect.New(effect.FireResistance{}, 1, time.Second*2))
+	case 8:
+		effects = append(effects, effect.New(effect.Regeneration{}, 1, time.Second*6))
+	case 9:
+		effects = append(effects, effect.New(effect.Wither{}, 1, time.Second*6))
+	default:
+		panic(fmt.Errorf("invalid stewEffect passed: %v", s.Uint8()))
```
JustTalDevelops, line 103: Can probably just panic with `should never happen` here.

JustTalDevelops (approved):

PR #675, opened 2022-10-02
Make the item.Stack argument in HandleItemPickup a pointer

RestartFU:
This can be useful for when you use your own items, with custom functionalities and want to have items drop on death and be picked up as they were before death. In my case, if the stack isn't replaced by my own custom soup implementation on pickup, then the soup is "broken", it would be a vanilla soup.

### DIFF
--- a/server/player/handler.go
+++ b/server/player/handler.go
@@ -1,6 +1,9 @@
 package player
 
 import (
+	"net"
+	"time"
+
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/cmd"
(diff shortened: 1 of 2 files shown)


Sandertv:
For the case you're describing, can that not be solved simply by using custom items?

NiklasZeroZero:
This would also allow you to update player specific item lore. (In an RPG like context for example)

Sandertv:
Would you'd need to do that on pick-up specifically?

NiklasZeroZero:
I mean... You could handle something like that somewhere else but I believe it would be the most logical and performant place to handle item updates on player Pickup inside HandleItemPickup

PR #680, opened 2022-10-16
Fix block entities being overridden/deleted when building structures

RoyalMCPE:


Sandertv:
This looks good, thank you!

PR #681, opened 2022-10-20
Update configuration interface so the Nop world provider can optionally be used

tadhunt:
When creating a `server.Config` from a `server.UserConfig` via `server.UserConfig.Config()`,  a persistent world provider (via `mcdb.New()`) is always created — there is no way to make use of `world.NopProvider` .

This PR addresses this lack by adding a new attribute `UserConfig.World.SaveData` , and updates `server.UserConfig.Config()` to honor it.

For consistency, this change matches the way `PlayerProvider` configuration is already implemented.

### DIFF
--- a/server/conf.go
+++ b/server/conf.go
@@ -187,6 +187,12 @@ type UserConfig struct {
 		QuitMessage string
 	}
 	World struct {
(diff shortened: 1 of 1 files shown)


Sandertv (approved):

PR #683, opened 2022-10-21
Change kill() to Kill()

UramnOIL:


Sandertv:
This is intentionally unexported. Users should use the Hurt() method to kill the player.

PR #685, opened 2022-10-22
Keep Inventory

### LINKED ISSUE #684: Keep Inventory
I would like to control `dropContent` in Player#kill() when a player dies. I think it is appropriate to add the flag to HandleDeath() like:
```go
func (h *MyHandler) HandleDeath(drop *bool, src world.DamageSource) {}
```

UramnOIL:
Close #684


Sandertv (approved):
This looks good to me, thank you for the PR!

PR #686, opened 2022-10-23
Initial v1.19.40 support

JustTalDevelops:
TODO:
- [ ] Support the item tag recipe format. This isn't a massive deal right now since v1.19.40 doesn't have any new recipes, but we should still address this sooner than later.


JustTalDevelops (commented):
Going to merge this in without support for item tagging recipes. This is useful - but because of the structuring of the system, it is also quite irritating to properly integrate at this time. This will eventually be added though, most likely before the next major version update to Minecraft.

PR #692, opened 2022-11-06
Remove last ioutil usage.

tristanmorgan:
With "ioutil" being deprecated in Golang 1.16+ this replaces it with calls to "os".


Sandertv (approved):
Looks good, thank you!

PR #693, opened 2022-11-09
Golang 1.19.x formatting.

tristanmorgan:
Non functional changes to the code caused by running `go fmt ./...` on it.
@Sandertv


Sandertv (approved):
Looks good, thanks!

PR #696, opened 2022-11-16
Add ability to change xp with HandleBlockBreak event

Flonja:


Sandertv (approved):
I think this looks fine. Thank you!

PR #705, opened 2022-11-30
Avoid saving unmodified chunks

JustTalDevelops:


server/world/chunk/chunk.go
```diff
@@ -14,6 +14,8 @@ type Chunk struct {
 	r cube.Range
 	// air is the runtime ID of air.
 	air uint32
+	// modified is true if the chunk has been modified since it was last saved.
+	modified bool
```
Sandertv, line 18: I think it'd make more sense to put this in the `chunkData` of the world package.

Sandertv (approved):

PR #707, opened 2022-12-15
Fix incorrect drops

ItsZodiaX:
In case you break the chest, all items in the chest that assigned to the drops variable got reassigned by drops of the current block.


JustTalDevelops (approved):
Thanks for the PR!

PR #716, opened 2022-12-21
Implement curse of binding

curzodo:
This PR implements the curse of binding enchantment.

- Items that implement the item.Armour interface can be enchanted with the curse of binding
- Items with the curse of binding equipped in the armour inventory cannot be removed by inventory interaction
- Items with the curse of binding cannot have the enchantment removed through the use of a grindstone
- The cost has been set at 0 because this enchantment is not obtainable through an enchanting table

### DIFF
--- a/server/item/enchantment/curse_of_binding.go
+++ b/server/item/enchantment/curse_of_binding.go
@@ -25,7 +25,7 @@ func (CurseOfBinding) Cost(level int) (int, int) {
 
 // Rarity ...
 func (CurseOfBinding) Rarity() item.EnchantmentRarity {
-	return item.EnchantmentRarityCommon
+	return item.EnchantmentRarityVeryRare
 }
 
 // CompatibleWithEnchantment ...
--- a/server/item/enchantment/curse_of_binding.go
+++ b/server/item/enchantment/curse_of_binding.go
@@ -1,0 +1,41 @@
+package enchantment
+
+import (
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// Curse of Binding is an enchantment that prevents removal of a cursed item from its armour slot.
+type CurseOfBinding struct{}
+
+// Name ...
+func (CurseOfBinding) Name() string {
+	return "Curse of Binding"
+}
+
+// MaxLevel ...
+func (CurseOfBinding) MaxLevel() int {
+	return 1
+}
+
+// Cost ...
+func (CurseOfBinding) Cost(level int) (int, int) {
+	return 0, 0
+}
+
+// Rarity ...
+func (CurseOfBinding) Rarity() item.EnchantmentRarity {
+	return item.EnchantmentRarityCommon
+}
+
+// CompatibleWithEnchantment ...
+func (CurseOfBinding) CompatibleWithEnchantment(item.EnchantmentType) bool {
+	return true
+}
+
+// CompatibleWithItem ...
+func (CurseOfBinding) CompatibleWithItem(i world.Item) bool {
+	_, isArmour := i.(item.Armour)
+
+	return isArmour
+}
--- a/server/item/enchantment/curse_of_binding.go
+++ b/server/item/enchantment/curse_of_binding.go
@@ -39,3 +39,13 @@ func (CurseOfBinding) CompatibleWithItem(i world.Item) bool {
 
 	return isArmour
 }
+
+// Treasure ...
+func (CurseOfBinding) Treasure() bool {
+    return true
+}
+
+// Curse ...
+func (CurseOfBinding) Curse() bool {
+    return true
+}
--- a/server/session/handler_item_stack_request.go
+++ b/server/session/handler_item_stack_request.go
@@ -6,6 +6,7 @@ import (
 	"github.com/df-mc/dragonfly/server/entity"
 	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/item/enchantment"
 	"github.com/df-mc/dragonfly/server/item/inventory"
 	"github.com/go-gl/mathgl/mgl64"
 	"github.com/sandertv/gophertunnel/minecraft/protocol"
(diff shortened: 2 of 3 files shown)


server/session/handler_item_stack_request.go
```diff
@@ -254,6 +274,15 @@ func (h *ItemStackRequestHandler) handleDrop(a *protocol.DropStackRequestAction,
 		return err
 	}
 
+	// Do not allow an equipped item cursed with binding to be dropped.
+	creativeMode := s.c.GameMode().CreativeInventory()
+	if a.Source.ContainerID == protocol.ContainerArmor && !creativeMode {
+		i, _ := h.itemInSlot(a.Source, s)
+		if _, isCursed := i.Enchantment(enchantment.CurseOfBinding{}); isCursed {
+			return nil
```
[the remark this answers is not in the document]
DaPigGuy: Reflect this change across all 3 locations.

server/session/handler_item_stack_request.go
```diff
@@ -157,6 +158,16 @@ func (h *ItemStackRequestHandler) handleTransfer(from, to protocol.StackRequestS
 	if (dest.Count()+int(count) > dest.MaxCount()) && !dest.Empty() {
 		return fmt.Errorf("client tried adding %v to item count %v, but max is %v", count, dest.Count(), dest.MaxCount())
 	}
+
+	// Do not allow an equipped item cursed with binding to be transferred.
+	creativeMode := s.c.GameMode().CreativeInventory()
+	if from.ContainerID == protocol.ContainerArmor && !creativeMode {
+		i, _ := h.itemInSlot(from, s)
```
DaPigGuy, line 165:
```suggestion
```
Already defined w/ the exact same value earlier

DaPigGuy (changes requested):

server/item/enchantment/curse_of_binding.go
```diff
@@ -0,0 +1,51 @@
+package enchantment
+
+import (
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// CurseOfBinding is an enchantment that prevents the removal of an equipped item enchanted with the curse of binding from its armour slot.
+type CurseOfBinding struct{}
```
DaPigGuy, line 9: The naming of the file (`curse_of_binding.go`) is not consistent with the naming of Curse of Vanishing (`vanishing.go`)

curzodo: I would argue that the naming of vanishing.go is not consistent with the naming of all the other enchantment files, which use the full enchantment names, with spaces represented as underscores.

server/session/handler_item_stack_request.go
```diff
@@ -197,6 +208,15 @@ func (h *ItemStackRequestHandler) handleSwap(a *protocol.SwapStackRequestAction,
 		return err
 	}
 
+	// Do not allow an equipped item cursed with binding to be swapped out.
+	creativeMode := s.c.GameMode().CreativeInventory()
+	if a.Destination.ContainerID == protocol.ContainerArmor && !creativeMode {
```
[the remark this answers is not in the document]
curzodo: As I understand it, it is perfectly fine for a cursed item to be swapped into an armour slot in which there is no cursed item.

server/item/enchantment/curse_of_binding.go
DaPigGuy: Fair enough. Feel free to change `vanishing.go`.

server/session/handler_item_stack_request.go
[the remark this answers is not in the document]
curzodo: Earlier I was testing the changes and unless the server explicitly prevents the client from dropping/swapping/taking, then the action is possible.

DaPigGuy (commented):
It might be a good idea to move the Curse of Binding logic to its own function, since it's duplicated several times in this handler, something like:

```go
func (h *ItemStackRequestHandler) validateCurse(slot protocol.StackRequestSlotInfo, s *Session) error {
}
```

```go
if err := validateCurse(a.Source, s); err != nil {
    return err
}
```

You could also change that function to return a bool rather than an error if you want more descriptive errors (swap vs. transfer vs. drop).

server/session/handler_item_stack_request.go
```diff
@@ -157,6 +158,12 @@ func (h *ItemStackRequestHandler) handleTransfer(from, to protocol.StackRequestS
 	if (dest.Count()+int(count) > dest.MaxCount()) && !dest.Empty() {
 		return fmt.Errorf("client tried adding %v to item count %v, but max is %v", count, dest.Count(), dest.MaxCount())
 	}
+
+	// Do not allow an equipped item cursed with binding to be transferred.
+	if curseOfBinding(from.ContainerID, i, s) {
```
DaPigGuy, line 163:
```suggestion
	if curseOfBinding(from.ContainerID, i, s) || curseOfBinding(to.ContainerID, dest, s)  {
```

DaPigGuy, line 163:
```suggestion
	if curseOfBinding(from.ContainerID, i, s) || curseOfBinding(to.ContainerID, dest, s)  {
```

DaPigGuy:
@Sandertv Can we try and get this merged soon?

server/player/player.go
```diff
@@ -2930,6 +2930,11 @@ func (p *Player) useContext() *item.UseContext {
 			srcIt, _ := srcInv.Item(src)
 			dstIt, _ := dstInv.Item(dst)
 
+			// If dstIt is enchanted with curse of binding, do not swap.
+			if _, isCursed := dstIt.Enchantment(enchantment.CurseOfBinding{}); isCursed && !p.GameMode().CreativeInventory() {
```
JustTalDevelops, line 2934: This feels out of place. How could this ever happen? I assume any attempt to move the item out will be cancelled at the session level?

curzodo: It does not get cancelled at the session level.

Alternatively, I could go to modify relevant items' [Use() function](https://github.com/df-mc/dragonfly/blob/b27aea2185a0513e13e0f63b6dcadc9dfcdb9a0f/server/item/helmet.go#L16) and check if the carrier (player) is wearing armour enchanted with curse of binding.

PR #721, opened 2022-12-24
Implement frosted ice and frost walker

curzodo:
This PR implements [frosted ice](https://minecraft.fandom.com/wiki/Frosted_Ice) and [frost walker](https://minecraft.fandom.com/wiki/Frost_Walker) as defined by the linked pages.

Some features of frost walker are currently impossible to implement:
- Foxes can pick up boots with frost walker, but foxes do not exist in Dragonfly
- Frost walker prevents damage from campfires and magma blocks when stepped on, but neither exist in Dragonfly

Currently, the main function contains code that spawns players with boots enchanted with frost walker one and two, and a bucket of water, for testing purposes.


curzodo:
Some of those commits from other people are showing up because things got a little messy when I was fighting with Git, but I believe everything is as it should be.

PR #722, opened 2022-12-24
implement frosted ice and frost walker

curzodo:
This PR implements [frosted ice](https://minecraft.fandom.com/wiki/Frosted_Ice) and [frost walker](https://minecraft.fandom.com/wiki/Frost_Walker) as defined by the linked pages.

Some features of frost walker are currently impossible to implement:

    Foxes can pick up boots with frost walker, but foxes do not exist in Dragonfly
    Frost walker prevents damage from campfires and magma blocks when stepped on, but neither exist in Dragonfly

Currently, the main function contains code that spawns players with boots enchanted with frost walker one and two, and a bucket of water, for testing purposes.


server/block/water.go
```diff
@@ -58,7 +58,13 @@ func (Water) SpreadDecay() int {
...
 	w.Depth = depth
 	w.Falling = falling
-	w.Still = false
+
+	if depth == 8 {
+		w.Still = true
+	} else {
+		w.Still = false
+	}
+
```
DaPigGuy, line 67:
```suggestion
	w.Still = depth == 8
```

server/item/enchantment/frost_walker.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// MaxLevel ...
+func (FrostWalker) MaxLevel() int {
+	return 2
+}
+
+// Cost ...
+func (FrostWalker) Cost(level int) (int, int) {
+	return 25, 50
```
DaPigGuy, line 25:
```suggestion
	min := level * 10
	return min, min + 15	
```

main.go
```diff
@@ -25,7 +30,12 @@ func main() {
 	srv.CloseOnProgramEnd()
 
 	srv.Listen()
-	for srv.Accept(nil) {
+	for srv.Accept(func(p *player.Player) {
```
DaPigGuy, line 33: Remove testing code

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/go-gl/mathgl/mgl64"
+	"math/rand"
+	"time"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+// FrostedIce ...
```
DaPigGuy, line 16: Complete the doc comment here

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+type FrostedIce struct {
+	solid
+
+	// Age ...
```
DaPigGuy, line 20: Same here

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	Age int
+}
+
+// Instrument ...
+func (FrostedIce) Instrument() sound.Instrument {
```
DaPigGuy, line 25: Don't believe Frosted Ice affects note blocks, please test

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	return sound.Chimes()
+}
+
+// BreakInfo ...
+func (p FrostedIce) BreakInfo() BreakInfo {
+	return newBreakInfo(0.5, alwaysHarvestable, pickaxeEffective, silkTouchOnlyDrop(p))
```
DaPigGuy, line 31: Should not drop any item in any scenario,

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Friction ...
+func (p FrostedIce) Friction() float64 {
+	return 0.98
```
DaPigGuy, line 36: @JustTalDevelops Can you confirm this value?

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+
+// DecodeNBT ...
+func (fi FrostedIce) DecodeNBT(data map[string]any) any {
+	fi.Age, _ = data["age"].(int)
+	return fi
+}
+
+// EncodeNBT ...
+func (fi FrostedIce) EncodeNBT() map[string]any {
+	return map[string]any{"age": int32(fi.Age)}
```
DaPigGuy, line 109: Vanilla Minecraft does not treat Frosted Ice as a block entity

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+			false,
+		}, nil)
+
+		return
+	}
+
+	w.SetBlock(pos, fi, nil)
+}
+
+func frostedIce(pos cube.Pos, w *world.World) int {
```
DaPigGuy, line 71: Code styling—move this function to the bottom of the file, above `allFrostedIce`

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	Age int
+}
+
+// Instrument ...
+func (FrostedIce) Instrument() sound.Instrument {
+	return sound.Chimes()
+}
+
+// BreakInfo ...
+func (p FrostedIce) BreakInfo() BreakInfo {
```
DaPigGuy, line 30: Code styling—receiver names should be consistent, preferably `f` rather than `p` or `fi`

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	"math/rand"
+	"time"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+// FrostedIce ...
+type FrostedIce struct {
```
DaPigGuy, line 17: FrostedIce should implement LightDiffuser, the block is partially transparent.

server/block/frosted_ice.go
DaPigGuy, line 31: According to the wiki, it "can be easily destroyed without [tools](https://minecraft.fandom.com/wiki/Tools)." Please check whether `pickaxeEffective` is applicable here.

DaPigGuy (changes requested):
This is a preliminary review, I haven't checked the actual Frosted Ice/Frost Walker logic yet.

server/block/frosted_ice.go
```diff
@@ -0,0 +1,118 @@
...
+	fi.Age++
+
+	if fi.Age > 3 {
+		w.SetBlock(pos, Water{
+			empty{},
+			replaceable{},
+			true,
+			8,
+			false,
+		}, nil)
```
DaPigGuy, line 63:
```suggestion
		w.SetBlock(pos, Water{Still: true, Depth: 8}, nil)
```

You don't need to provide every value, instead provide values w/ the field name

server/item/enchantment/frost_walker.go
curzodo: Why did you choose these values?

I'm not sure if it matters, but frost walker is not an enchantment available in enchanting tables.

DaPigGuy: https://minecraft.fandom.com/wiki/Enchanting/Levels

It affects the amount of XP the grindstone gives back

curzodo: Gotcha

server/block/frosted_ice.go
curzodo: Apologies, I copy pasted the entire file from an already existing block as a template and forgot about the receiver names.

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+	return bi
+}
+
+// Friction ...
+func (fi FrostedIce) Friction() float64 {
+	return 0.98
+}
+
+// LightDiffusionLevel ...
+func LightDiffusionLevel() uint8 {
```
DaPigGuy, line 55: Missing receiver

```suggestion
func (FrostedIce) LightDiffusionLevel() uint8 {
```

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+	solid
+	
+	// Age is the melting state of frosted ice. Values range from 0 to 4.
+	// When the age is 4 however, the block melts completely into water.
+	Age int
+}
+
+// BreakInfo ...
+func (fi FrostedIce) BreakInfo() BreakInfo {
+	effective := func(_ item.Tool) bool { return true }
```
DaPigGuy, line 29: There's already an `alwaysEffective` variable in this package.

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+
+	drops := func(_ item.Tool, _ []item.Enchantment) []item.Stack { return []item.Stack{} }
+
+	bi := newBreakInfo(0.5, alwaysHarvestable, effective, drops)
```
DaPigGuy, line 33:
```suggestion
	bi := newBreakInfo(0.5, alwaysHarvestable, alwaysEffective, simpleDrops())
```

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+
+	breakHandler := func(pos cube.Pos, w *world.World, u item.User) {
+		mainHandItem, _ := u.HeldItems()
+
+		if _, isSilkTouch := mainHandItem.Enchantment(enchantment.SilkTouch{}); isSilkTouch {
+			return
+		}
+
```
DaPigGuy, line 41: Silk touch appears to have no affects on Frosted Ice

```suggestion
```

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+	// Tick block once every one and a half seconds.
+	if rand.Intn(19) != 0 {
+		return
+	}
+
+	// One of two conditions must be met in order to increase the age of frosted ice.
+	// 1) A random number generator with possible values [0, 1, 2] chooses 0 OR the frosted
+	//    ice block has fewer than 4 frosted ice blocks immediately surrounding it.
+	// 2) The light level at the block is greater than 11 minus its age.
+	if (rand.Intn(2) != 0 && (frostedIce(pos, w) >= 4)) || w.Light(pos) <= uint8((11-fi.Age)) {
```
DaPigGuy, line 70: `rand.Intn` generates a number between [0, n), where n is excluded. The chance is currently `1/2` rather than `1/3` as a result

```suggestion
	if (rand.Intn(3) != 0 && frostedIce(pos, w) >= 4) || w.Light(pos) <= uint8((11-fi.Age)) {
```

server/block/frosted_ice.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Tick ...
+func (fi FrostedIce) Tick(_ int64, pos cube.Pos, w *world.World) {
+	// Tick block once every one and a half seconds.
+	if rand.Intn(19) != 0 {
+		return
+	}
+
+	// One of two conditions must be met in order to increase the age of frosted ice.
```
DaPigGuy, line 66: Both your code & the wiki suggest that both conditions have to be met.

DaPigGuy (changes requested):

server/block/frosted_ice.go
curzodo: According to the fandom wiki, if a frosted ice block is mined with silk touch, a water block is not left in its place and nothing is dropped.

server/block/frosted_ice.go
curzodo: "If a 1⁄3 chance succeeds or the block has fewer than 4 frosted ice blocks in the 6 blocks surrounding it, and the light level (including light from the sun) at the block would be greater than 11 minus the age, the age is incremented."

I have interpreted this as:  (1⁄3 chance || less than 4) && light levels

The comma after 'surrounding it' would suggest this, no? Furthermore, I have tested frost walker in vanilla bedrock, and at night time the frosted ice does not age at all, suggesting that my interpretation is correct as aging will not occur unless the light levels are above a certain number. My code inverts the statement I wrote above.

I did however notice something quite unusual when testing on the dragonfly server. If I place a frosted ice block above the ground (so that it is not surrounded by any other blocks) it begins melting as it usually would. The same happens if I place a bucket of water above ground and then jump to freeze the source, it begins melting, even at night, when it shouldn't. I suspect there is something going on with the way dragonfly calculates light.

curzodo: My comment describing the conditions is however incorrect, I mended it.

server/block/frosted_ice.go
DaPigGuy: Tested in vanilla Minecraft—silk touch has no effect.

server/block/frosted_ice.go
curzodo: So, no EncodeNBT or DecodeNBT?

server/block/frosted_ice.go
curzodo: alwaysEffective does not exist, I can create it though?

DaPigGuy: nothingEffective*

DaPigGuy (changes requested):
FrostedIce must implement both the `ScheduledTicker` and `NeighbourUpdateTicker`, and have scheduled updates queued on neighbor updates.

curzodo:
> FrostedIce must implement both the `ScheduledTicker` and `NeighbourUpdateTicker`, and have scheduled updates queued on neighbor updates.

Why?

PR #724, opened 2022-12-26
Implemented ice

JavierLeon9966:


server/block/ice.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// LightDiffusionLevel ...
+func (Ice) LightDiffusionLevel() uint8 {
+	return 3
+}
+
+// BreakInfo ...
+func (i Ice) BreakInfo() BreakInfo {
+	return BreakInfo{
```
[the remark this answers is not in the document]
DaPigGuy: https://github.com/df-mc/dragonfly/blob/master/server/block/break_info.go#L122

DaPigGuy (changes requested):

server/block/ice.go
```diff
@@ -0,0 +1,66 @@
...
+			}); ok && p.GameMode().CreativeInventory() {
+				return
+			}
+			if mainHand, _ := u.HeldItems(); hasSilkTouch(mainHand.Enchantments()) {
+				return
+			}
+			if _, ok := w.Block(pos.Side(cube.FaceDown)).Model().(model.Solid); !ok {
+				return
+			}
+			w.SetBlock(pos, Water{}, nil)
```
DaPigGuy, line 40: Should not create water on break/melt in the Nether

server/block/ice.go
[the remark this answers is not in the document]
JavierLeon9966: I did that instead to set the blast resistance to 0.5, but I guess I can set it after `newBreakInfo()`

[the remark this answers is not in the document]
DaPigGuy: If the wiki hardness & blast resistance values are the same, the blast resistance value doesn't need to be provided.

server/block/water.go
```diff
@@ -112,6 +113,37 @@ func (w Water) ScheduledTick(pos cube.Pos, wo *world.World, _ *rand.Rand) {
...
 
+// RandomTick ...
+func (w Water) RandomTick(pos cube.Pos, wo *world.World, _ *rand.Rand) {
+	if w.Depth != 8 {
+		return
+	}
+	if wo.Temperature(pos) > 0.0 {
+		return
+	}
+	if wo.Light(pos) > 11 {
```
DaPigGuy, line 124:
```suggestion
	if wo.Light(pos) > 12 {
```

From the Minecraft wiki:
> **[Water](https://minecraft.fandom.com/wiki/Water) source blocks in a snowy [biome](https://minecraft.fandom.com/wiki/Biome) eventually freeze into ice if exposed to the sky from directly above, the [light level](https://minecraft.fandom.com/wiki/Light_Level) immediately adjacent to the water block on all sides is less than 13**, and there is at least one horizontally adjacent non-water (and non-[waterlogged](https://minecraft.fandom.com/wiki/Waterlogging)) block. This can happen at any [time of day](https://minecraft.fandom.com/wiki/Daylight_cycle), and in any [weather](https://minecraft.fandom.com/wiki/Weather) condition. **If the highest adjacent light level is 12, an ice block alternately melts and re-freezes when it receives a [block tick](https://minecraft.fandom.com/wiki/Tick#Random_tick).**

server/block/register.go
```diff
@@ -55,6 +55,7 @@ func init() {
 	world.RegisterBlock(Grass{})
 	world.RegisterBlock(Gravel{})
 	world.RegisterBlock(Honeycomb{})
+	world.RegisterBlock(Ice{})
```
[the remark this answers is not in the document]
JavierLeon9966: But I don't see any blocks being registered as items, unless you mean the `EncodeItem()` method in which I already added

server/block/ice.go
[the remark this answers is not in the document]
JavierLeon9966: If I don't provide the value then the blast resistance would just be `0`

server/block/register.go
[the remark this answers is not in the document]
DaPigGuy: Scroll further down 😅

server/block/ice.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		if mainHand, _ := u.HeldItems(); hasSilkTouch(mainHand.Enchantments()) {
+			return
+		}
+		if _, ok := w.Block(pos.Side(cube.FaceDown)).Model().(model.Solid); !ok {
+			return
+		}
+		w.SetBlock(pos, Water{}, nil)
+	})
+	b.withBlastResistance(0.5)
```
DaPigGuy, line 43:
```suggestion
```
https://github.com/df-mc/dragonfly/blob/master/server/block/break_info.go#L102

server/block/ice.go
```diff
@@ -0,0 +1,73 @@
...
+
+// LightDiffusionLevel ...
+func (Ice) LightDiffusionLevel() uint8 {
+	return 3
+}
+
+// BreakInfo ...
+func (i Ice) BreakInfo() BreakInfo {
+	b := newBreakInfo(0.5, alwaysHarvestable, pickaxeEffective, silkTouchOnlyDrop(i))
+	b.withBreakHandler(func(pos cube.Pos, w *world.World, u item.User) {
```
DaPigGuy, line 25: The returned BreakInfo does not have the break handler.

server/block/ice.go
JavierLeon9966: 0.5 * 5 = 2.5, which is not 0.5 as stated by the wiki

server/block/ice.go
JavierLeon9966: My bad

server/block/water.go
```diff
@@ -112,6 +113,37 @@ func (w Water) ScheduledTick(pos cube.Pos, wo *world.World, _ *rand.Rand) {
...
+	pos.Neighbours(func(neighbour cube.Pos) {
+		if solidAdjacent {
+			return
+		}
+		if neighbour[1] != pos[1] {
+			return
+		}
+		if _, ok := wo.Block(neighbour).Model().(model.Solid); ok {
+			solidAdjacent = true
+		}
```
DaPigGuy, line 140:
```suggestion
		_, solidAdjacent = wo.Block(neighbour).Model().(model.Solid)
```

server/block/ice.go
```diff
@@ -0,0 +1,73 @@
...
+func (i Ice) RandomTick(pos cube.Pos, w *world.World, _ *rand.Rand) {
+	if w.Light(pos) < 12 {
+		return
+	}
+	if w.Dimension().WaterEvaporates() {
+		w.SetBlock(pos, nil, nil)
+		w.AddParticle(pos.Vec3Centre(), particle.Evaporate{})
+	} else {
+		w.SetBlock(pos, Water{}, nil)
+	}
```
DaPigGuy, line 57:
```suggestion
		return
	}
	w.SetBlock(pos, Water{}, nil)
```

server/block/ice.go
DaPigGuy: The wiki values are 1/5 of the values we used within dragonfly.

server/block/ice.go
DaPigGuy: Should probably just inline this with the `return`

server/block/ice.go
JavierLeon9966: Alright then

PR #729, opened 2023-01-08
Limit transactions size

szerookii:
# Summary

This pull request fixes a bug where inventory transactions were being processed incorrectly when the quantity of inventory actions was too large.

# Details

When the quantity of actions in an inventory transaction exceeded a certain value, the server was freezing.

This pull request fixes the issue by adding a check to ensure that the quantity of actions does not exceed the maximum allowed value before processing the transaction.

# Impacted Areas in Application

- Inventory transactions

### DIFF
--- a/server/session/handler_inventory_transaction.go
+++ b/server/session/handler_inventory_transaction.go
@@ -1,6 +1,7 @@
 package session
 
 import (
+	"errors"
 	"fmt"
 	"github.com/df-mc/dragonfly/server/block/cube"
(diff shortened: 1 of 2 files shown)


Sandertv:
I don't think this is the right solution to this problem. I think we should approach the problem from much lower level in gophertunnel. I also don't see exactly how the server would be freezing, given it would at max block just one goroutine.

szerookii:
Yes, I did not explain myself well, it would indeed block that a goroutine but it would nevertheless use resources for nothing.

PR #733, opened 2023-01-25
world/biome: Added the Windswept Hills and Gravelly Hills biomes.

tristanmorgan:
I found some more biomes not previously covered (extreme_hills and extreme_hills_mutated).


JustTalDevelops (approved):
Thanks for the PR!

PR #734, opened 2023-01-27
fix spectors able to pick up items from the floor

riccskn:


riccskn:
*spectators

server/player/player.go
```diff
@@ -2087,6 +2087,9 @@ func (p *Player) Collect(s item.Stack) int {
 	if p.Dead() {
 		return 0
 	}
+	if p.GameMode() == world.GameModeSpectator {
```
TwistedAsylumMC, line 2090: Could you update this to use the `AllowsInteraction()` method from the GameMode interface? It's just a more generic way to achieve the same result and allows custom gamemodes to inherit the same functionality

riccskn: Ok

PR #741, opened 2023-02-02
Implement event structs

curzodo:
This PR implements event structs. Event structs represent specific events, for example, the player chat event or the world liquid harden event. Event handler functions such as HandleChat, or HandleLiquidHarden consequently take an event struct as opposed to zero, single or multiple arguments. Using structs to represent events in favour of individual arguments improves code readability and is more backwards compatible, as new event variables, for example adding a pointer to a string that represents player chat prefix as an argument to HandleChat, will no longer cause breaking changes.

Each event struct contains a pointer to the 'main actor' of the event. For example, in each player event struct there is a pointer to the player that event revolves around. In my opinion this is preferable to the current format that does not provide the 'main actor' to the handler function directly, where the player must be either stored in the handler struct, or the function must be defined in a scope which already defines the player it needs access to. It seems to me that it is only logical that the 'main actor' around which the entire event is based should be given directly to the handler function along with other attributes about that event. The following is an example of what the event struct system looks like:

```
type MyHandler struct {
    player.NopHandler
}

func (MyHandler) HandleChat(e player.EventChat) {
    filterChat(e)
    styleChat(e)
}

func filterChat(e player.EventChat) {
    if *e.Message = "crap" {
        e.Cancel()
    }
}

func styleChat(e player.EventChat) {}
```
For events that can be cancelled, the ctx has been embedded into the event struct, so that one may call Cancel() or Cancelled() directly on 'e', as shown above.

I am aware that comments have not been changed/added, but will perhaps need to be, I wanted to discuss that here before I made any changes. In my opinion, the comments above each handler interface function definition should be removed, and new comments made above each event struct. I am willing to do this, if desired. The new comments would be in most cases identical to the old ones, but improved grammatically, where applicable.

### DIFF
--- a/server/block/fire.go
+++ b/server/block/fire.go
@@ -184,12 +184,27 @@ func (f Fire) tick(pos cube.Pos, w *world.World, r *rand.Rand) {
 func (f Fire) spread(from, to cube.Pos, w *world.World, r *rand.Rand) {
 	if _, air := w.Block(to).(Air); !air {
 		ctx := event.C()
-		if w.Handler().HandleBlockBurn(ctx, to); ctx.Cancelled() {
+
+		evt := world.EventBlockBurn{
+			w,
+			to,
+			ctx,
+		}
+
+		if w.Handler().HandleBlockBurn(evt); evt.Cancelled() {
 			return
 		}
 	}
 	ctx := event.C()
-	if w.Handler().HandleFireSpread(ctx, from, to); ctx.Cancelled() {
+
+	evt := world.EventFireSpread{
+		w,
+		from,
+		to,
+		ctx,
+	}
+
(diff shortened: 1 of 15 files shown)


Sandertv:
Pros of this PR (as mentioned):
- Much easier to retain backward compatibility
- Somewhat more consistent API

Cons of this PR:
- Considerably more lines when calling an event (events could be on one line but they will be _very_ long lines)
- Slightly more effort to create new events as a struct for each is necessary
- Amount of types in packages blows up. Not necessarily a massive issue, but it does "pollute" the package

As for the "Each event struct contains a pointer to the 'main actor' of the event.", I don't necessarily think that is an essential change. That means you have to pass more data to every event while users already have access to that data.

I also have some doubts about embedding an `*event.Context`, it's pretty easy to forget to initialise it like this.

PR #747, opened 2023-02-05
SetServerSettings function

TacoError:


server/session/session.go
```diff
@@ -538,3 +544,14 @@ func (s *Session) sendAvailableEntities(w *world.World) {
 	}
 	s.writePacket(&packet.AvailableActorIdentifiers{SerialisedEntityIdentifiers: serializedEntityData})
 }
+
+// SetServerSettings Sets the server settings data.
+// This will be used when a ServerSettingsRequest packet is
+// sent by the client.
+func (s *Session) SetServerSettings(form form.Form) {
+	h := s.handlers[packet.IDModalFormResponse].(*ModalFormResponseHandler)
+	h.forms[h.currentID.Add(1)] = form
```
Prim69, line 553: Inc()

server/session/handler_server_settings_request.go
```diff
@@ -0,0 +1,21 @@
+package session
+
+import (
+	"github.com/sandertv/gophertunnel/minecraft/protocol/packet"
+)
+
+// ServerSettingsRequest Is send by the client
```
JustTalDevelops, line 7: Would fix the grammar here, and also inline the comment below into the one.

server/session/handler_server_settings_request.go
```diff
@@ -0,0 +1,21 @@
...
+// in order to request a ServerSettingsResponse packet
+type ServerSettingsRequest struct{}
+
+// Handle Sends the ServerSettingsResponse packet.
```
JustTalDevelops, line 11: Can just replace this with `// Handle ...`

server/session/session.go
```diff
@@ -79,6 +79,10 @@ type Session struct {
 	openChunkTransactions []map[uint64]struct{}
 	invOpened             bool
 
+	shouldSendSettingsResponse bool
+	serverSettingsResponse     []byte
+	serverSettingsFormID       uint32
```
JustTalDevelops, line 84: You don't ever update this field.

server/session/session.go
```diff
@@ -79,6 +79,10 @@ type Session struct {
 	openChunkTransactions []map[uint64]struct{}
 	invOpened             bool
 
+	shouldSendSettingsResponse bool
```
JustTalDevelops, line 82: I would rename this and the field below it to something slightly better, maybe `hasServerSettings` and `serverSettings`

JustTalDevelops: These should also probably be atomic

server/session/session.go
JustTalDevelops: This should also be atomic, if you intend on using it.

server/session/session.go
```diff
@@ -538,3 +544,14 @@ func (s *Session) sendAvailableEntities(w *world.World) {
 	}
 	s.writePacket(&packet.AvailableActorIdentifiers{SerialisedEntityIdentifiers: serializedEntityData})
 }
+
+// SetServerSettings Sets the server settings data.
+// This will be used when a ServerSettingsRequest packet is
+// sent by the client.
+func (s *Session) SetServerSettings(form form.Form) {
```
JustTalDevelops, line 551: The documentation here is poor. You should look at the Go documentation style guide: https://go.dev/blog/godoc

TacoError (commented):
sasad

Sandertv:
This method isn't actually accessible to users as access to sessions is restricted. How do you reckon we should expose this to users?

TacoError:
Didn't mean to close this, purged my github repositories. My bad.

PR #752, opened 2023-02-19
Implement MoveActorDelta

AkmalFairuz:
Implement MoveActorDelta to reduce bandwidth usage


Sandertv:
Bandwidth difference is only minimal when using the MoveActorDelta packet and there is considerably more logic to it. I'm not sure if this is worth it.

PR #754, opened 2023-03-14
add uniqueID to all entities and save using actorstorage

olebeck:
implementing https://learn.microsoft.com/en-us/minecraft/creator/documents/actorstorage
this is how the game stores entities so now entities saved from dragonfly show up when the world is loaded in single player
and entities in worlds from singleplayer show up in dragonfly.
also checks for the old keyEntities based entity data for loading older worlds, i couldnt test that since i dont have an older world but it should work.

### DIFF
--- a/server/entity/area_effect_cloud.go
+++ b/server/entity/area_effect_cloud.go
@@ -1,18 +1,21 @@
 package entity
 
 import (
+	"math/rand"
+	"time"
+
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/entity/effect"
 	"github.com/df-mc/dragonfly/server/internal/nbtconv"
 	"github.com/df-mc/dragonfly/server/item/potion"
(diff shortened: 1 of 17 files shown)


Sandertv:
Hi! First of all, thank you for the PR! It is very much appreciated.

Unfortunately, however, we are trying to make some changes to entities to reduce the amount of boilerplate code. As you were probably able to tell with this PR, you had to add the unique ID to every one of the entities individually, which is far from ideal. I'm hoping to come up with a system for entities where we can share more code so that this isn't necessary, but doing so isn't very easy hence the lack of progress. 

For now, I'm going to keep this open. I would like to merge it at some point, but ideally once we can get entities themselves to a bit of a better point.

olebeck:
ok that sounds good

Sandertv:
I have incorporated your changes into https://github.com/df-mc/dragonfly/pull/942 so this is not part of the codebase. Thank you for the PR!

PR #759, opened 2023-04-24
server/player.go: Only do thorns damage when above 0.

DidntPot:


Sandertv:
Thanks for fixing this!

PR #766, opened 2023-05-10
implement entity specific name tags

RestartFU:


Sandertv:
I am considering a more general sort of ViewLayer type (interface) that each player has. That way we don't have to manage the behaviour ourselves and people can decide what its exact behaviour should be themselves.

RestartFU:
> I am considering a more general sort of ViewLayer type (interface) that each player has. That way we don't have to manage the behaviour ourselves and people can decide what its exact behaviour should be themselves. 

what would that interface look like?

PR #767, opened 2023-05-12
added Milk() method to BuckedContent

RestartFU:


Sandertv:
Any specific reason for this as opposed to bucket.Content() == MilkBucket()?

PR #773, opened 2023-05-25
viewer: ViewEntityAnimation

unickorn:
This allows custom entities to send custom animations as well.


server/world/viewer.go
```diff
@@ -47,6 +47,8 @@ type Viewer interface {
 	// ViewEntityState views the current state of an entity. It is called whenever an entity changes its
 	// physical appearance, for example when sprinting.
 	ViewEntityState(e Entity)
+	// ViewEntityAnimation views an animation performed by an entity.
```
TwistedAsylumMC, line 50: I would probably change this to mention that it triggers an animation to start playing instead since it currently sounds like it views an already-playing animation

AndreasHGK (approved):

PR #781, opened 2023-06-07
1.20.0 Support

TwistedAsylumMC:


server/block/carpet.go
```diff
@@ -34,12 +35,12 @@ func (c Carpet) BreakInfo() BreakInfo {
 
 // EncodeItem ...
 func (c Carpet) EncodeItem() (name string, meta int16) {
-	return "minecraft:carpet", int16(c.Colour.Uint8())
+	return fmt.Sprintf("minecraft:%s_carpet", c.Colour.String()), 0
```
Sandertv, line 38: I wouldn't use fmt.Sprintf in these methods. It is quite slow and they're all strings, so + works too.

server/block/carpet.go
```diff
@@ -34,12 +35,12 @@ func (c Carpet) BreakInfo() BreakInfo {
...
 }
 
 // EncodeBlock ...
 func (c Carpet) EncodeBlock() (name string, properties map[string]any) {
-	return "minecraft:carpet", map[string]any{"color": c.Colour.String()}
+	return fmt.Sprintf("minecraft:%s_carpet", c.Colour.String()), nil
```
Sandertv, line 43: Same as above.

server/block/coral.go
```diff
@@ -94,15 +95,18 @@ func (c Coral) BreakInfo() BreakInfo {
 
 // EncodeBlock ...
 func (c Coral) EncodeBlock() (name string, properties map[string]any) {
-	return "minecraft:coral", map[string]any{"coral_color": c.Type.Colour().String(), "dead_bit": c.Dead}
+	if c.Dead {
+		return fmt.Sprintf("minecraft:dead_%s_coral", c.Type), nil
```
Sandertv, line 99: Same here too, won't repeat again

Sandertv (commented):
Looks good. I would only remove the fmt.Sprintfs in block/item encoding, because that has quite the performance penalty.

PR #783, opened 2023-06-07
made the item.Stack a pointer in HandleItemPickup

RestartFU:


AndreasHGK (approved):

Sandertv:
Although I don't quite agree with the original motivation for this PR - accounting for custom items with vanilla IDs - I imagine this is probably useful for other purposes(?)

Issue #192, opened 2021-06-26
Implement chest pairing


T14Raptor:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2450%20Funded-%2300A156.svg" />](https://oss.issuehunt.io/r/df-mc/dragonfly/issues/192)

Double chests

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [<img src='https://avatars.githubusercontent.com/u/39247615?v=4' alt='justtaldevelops' width=24 height=24> justtaldevelops](https://oss.issuehunt.io/u/justtaldevelops) ($50.00)

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/df-mc/dragonfly/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

xNatsuri:
this issue can be closed resolved in #872

zhanglin2603:
Generated at: 2026-02-12 14:07 UTC
Target: Implement chest pairing
URL: https://github.com/df-mc/dragonfly/issues/192
Estimated bounty: $50
Internal score: 23.0

Draft proposal:
**I'd like to apply for the $50 bounty on #192 (Implement chest pairing). I understand this involves adding support for pairing adjacent chests into double chests in Dragonfly, matching vanilla Minecraft behavior.**

**Delivery plan:**
- Analyze existing chest implementation and identify pairing logic gaps.
- Implement pairing/unpairing with proper block updates, inventory merging, and NBT handling.

**Please confirm the acceptance criteria, especially regarding corner cases (e.g., trapped chests, orientation, inventory persistence).**

PR #792, opened 2023-06-13
Implemented double chests

T14Raptor:
#192


TwistedAsylumMC:
Implemented by https://github.com/df-mc/dragonfly/pull/872

PR #807, opened 2023-08-04
Remove resending blocks for obstructedPos

Flonja:
It's kinda a hack I know, but go no other clue.

This is meant to remove block lag for high ping players.


Flonja:
bump

server/player/player.go
```diff
@@ -1718,25 +1720,30 @@ func (p *Player) placeBlock(pos cube.Pos, b world.Block, ignoreBBox bool) bool {
...
 // The function returns true if there is an entity in the way that could prevent the block from being placed.
-func (p *Player) obstructedPos(pos cube.Pos, b world.Block) bool {
+func (p *Player) obstructedPos(pos cube.Pos, b world.Block) (bool, bool) {
 	w := p.World()
 	blockBoxes := b.Model().BBox(pos, w)
 	for i, box := range blockBoxes {
 		blockBoxes[i] = box.Translate(pos.Vec3())
 	}
 
+	resend := true
```
RestartFU, line 1730: instead of having this here, I would have the returned type be name `(obstructed bool, resend bool)`

Flonja:
bump

didntpot:
merge plz

server/player/player.go
```diff
@@ -1751,26 +1753,31 @@ func (p *Player) placeBlock(pos cube.Pos, b world.Block, ignoreBBox bool) bool {
 }
 
 // obstructedPos checks if the position passed is obstructed if the block passed is attempted to be placed.
-// The function returns true if there is an entity in the way that could prevent the block from being placed.
-func (p *Player) obstructedPos(pos cube.Pos, b world.Block) bool {
+// The first bool indicates if there is an entity in the way that could prevent the block from being placed.
+// The second bool is used to check if the block should be updated.
```
Sandertv, line 1757: Could you expand this documentation to explain why it should not be updated?

server/player/player.go
```diff
@@ -1751,26 +1753,31 @@ func (p *Player) placeBlock(pos cube.Pos, b world.Block, ignoreBBox bool) bool {
...
+	resend = true
 	around := w.EntitiesWithin(cube.Box(-3, -3, -3, 3, 3, 3).Translate(pos.Vec3()), nil)
 	for _, e := range around {
 		switch e.Type().(type) {
 		case entity.ItemType, entity.ArrowType:
 			continue
 		default:
+			if e == p {
+				resend = false
+			}
```
Sandertv, line 1774: This section isn't entirely clear to me. If any of the entities in this 7x7x7 bounding box is the player itself, it will not resend. This doesn't quite sound right?

Flonja: If there is enough latency the client position might be in the block that the client wants to be placed.
You can easily try this out by having >=150ms and stacking blocks under you

Sandertv: If anything I would first move this block inside of the `if cube.AnyIntersections {}` block, because the player might not even be intersecting in the first place. But then I'm still not sure if it would work as expected. Imagine there was another player that this intersected with, in addition to the player itself, should it be resent or not?

PR #826, opened 2023-11-21
Fix firework explosion

sqmatheus:
Fixes:
- #821: Boosting an elytra with a firework that have explosions effects crashes the server, because the distance beetween the owner and the firework is 0. (not sure about the solution that I came with)
- `Projectile` in `ProjectileDamageSource` was the target itself.


Sandertv:
This has been fixed since. Thanks for the PR!

PR #827, opened 2023-12-05
Fix Elytras

eminarican:


eminarican:
when testing with the latest commit and the latest version of minecraft everything seems fine with landing
i suspect it was something clientside as there have been no recent changes or updates related to elytra

in addition i apologize for not knowing damage is taken whilst colliding with walls while wearing an elytra
and missing the part that vertical collisions are actually being handled as fall damage
therefore i was assuming that someone was confusing the direction needed for damage to be dealt...

so i'll close this pull request here

PR #830, opened 2023-12-12
Export functions which are important when implementing custom blocks

Erkam246:
So you don't have to copy the function and implement it on your own.


RestartFU:
these are all just small utilities to save some lines.

you can just copy paste them in your server core, and they'll work fine.

I honestly don't think they should be exported.

PR #836, opened 2023-12-21
Update note.go

JAILKOOKA:


TwistedAsylumMC:
You have not given any description for this change, and this just breaks noteblock functionality

PR #839, opened 2023-12-23
play sounds when waxing sign & when interacting with a waxed sign

Erkam246:


RestartFU (approved):
looks good, I think we can merge this

server/block/sign.go
```diff
@@ -119,10 +120,13 @@ func (s Sign) Wax(cube.Pos, mgl64.Vec3) (world.Block, bool) {
 }
 
 // Activate ...
-func (s Sign) Activate(pos cube.Pos, _ cube.Face, _ *world.World, user item.User, _ *item.UseContext) bool {
+func (s Sign) Activate(pos cube.Pos, _ cube.Face, w *world.World, user item.User, _ *item.UseContext) bool {
 	if editor, ok := user.(SignEditor); ok && !s.Waxed {
 		editor.OpenSign(pos, s.EditingFrontSide(pos, user.Position()))
 	}
+	if s.Waxed {
```
Sandertv, line 127: Could change this to `else if s.Waxed`.

server/world/sound/block.go
```diff
@@ -176,6 +176,12 @@ type ComposterReady struct{ sound }
 // LecternBookPlace is a sound played when a book is placed in a lectern.
 type LecternBookPlace struct{ sound }
 
+// WaxOn is a sound played when a sign is waxed.
+type WaxOn struct{ sound }
```
Sandertv, line 180: Not a huge fan of the naming here. Maybe `SignWaxed` would be nicer?

server/world/sound/block.go
```diff
@@ -176,6 +176,12 @@ type ComposterReady struct{ sound }
...
+
+// WaxedInteractFail is a sound played when a player tries to interact with a waxed sign.
+type WaxedInteractFail struct{ sound }
```
Sandertv, line 183: Same for the naming here. Maybe something like `WaxedSignFailedInteraction`?

Sandertv (commented):
Thanks for the PR! I have a couple of comments.

Erkam246:
Done

PR #842, opened 2024-01-10
Fixed note block, plays the previous note when breaking.

mokhtarfa:


Sandertv:
Next time please write a description of your PR so we can understand what the issue is that it aims to solve.

Thanks for the PR!

PR #843, opened 2024-01-12
Undo previous commit and improve note block punch. 

mokhtarfa:
I didn't test it in survival mode, left click on note block makes a sound, hope you merge this pr.


DaPigGuy (changes requested):
I've looked into this a bit. This seems to highlight generally wrong behavior with blocks that handle punches, e.g. dragon eggs.

However, this fix is not quite right either. For touchscreen input, there is a `Delayed Block Breaking` option that allows you to trigger a left click on blocks in creative mode without breaking. 

The behavior for calling `block.Punchable.Punch` should be changed under `player.Player.StartBreaking` to only call if the player isn't already breaking at the current position.

Additionally, please run `go fmt` locally to avoid CI failures.

PR #844, opened 2024-01-23
Implementation of a view layer

Flonja:
Co-authored-by: RestartFU

### DIFF
--- a/server/session/session.go
+++ b/server/session/session.go
@@ -81,6 +81,8 @@ type Session struct {
 
 	joinMessage, quitMessage string
 
+	viewLayer *world.ViewLayer
+
 	closeBackground chan struct{}
 }
 
@@ -164,6 +166,7 @@ func New(conn Conn, maxChunkRadius int, log Logger, joinMessage, quitMessage str
 		heldSlot:               atomic.NewUint32(0),
 		joinMessage:            joinMessage,
 		quitMessage:            quitMessage,
+		viewLayer:              world.NewViewLayer(),
 		openedWindow:           *atomic.NewValue(inventory.New(1, nil)),
 	}
 
@@ -232,6 +235,7 @@ func (s *Session) Close() error {
 // close closes the session, which in turn closes the controllable and the connection that the session
 // manages.
 func (s *Session) close() {
+	_ = s.viewLayer.Close()
 	_ = s.c.Close()
 
 	// Move UI inventory items to the main inventory.
--- a/server/session/view_layer.go
+++ b/server/session/view_layer.go
@@ -1,0 +1,11 @@
+package session
+
+import "github.com/df-mc/dragonfly/server/world"
+
+type LayerViewer interface {
+	ViewLayer() *world.ViewLayer
+}
+
+func (s *Session) ViewLayer() *world.ViewLayer {
+	return s.viewLayer
+}
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -913,9 +913,12 @@ func (s *Session) ViewEntityState(e world.Entity) {
 		if nt := s.viewLayer.NameTag(v); len(nt) > 0 {
 			metadata[protocol.EntityDataKeyName] = nt
 		}
-		invisFlag := metadata.Flag(protocol.EntityDataKeyFlags, protocol.EntityDataFlagInvisible)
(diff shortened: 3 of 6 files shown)


RestartFU (dismissed):

server/session/session.go
```diff
@@ -232,6 +235,7 @@ func (s *Session) Close() error {
 // close closes the session, which in turn closes the controllable and the connection that the session
 // manages.
 func (s *Session) close() {
+	_ = s.viewLayer.Close()
```
T14Raptor, line 238: Why should one viewer leaving close the entire view layer?

server/world/view_layer.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Invisible returns the invisibility of a viewer.
+func (v *ViewLayer) Invisible(viewer LayerViewer) bool {
+	v.viewerMu.Lock()
+	defer v.viewerMu.Unlock()
+	return v.viewers[viewer].invisible
+}
+
+// Close closes the view layer.
+func (v *ViewLayer) Close() error {
```
T14Raptor, line 85: This should also remove the view layer instance from the viewers.

Flonja: Like `v = nil`?
Can you be clearer?

server/session/session.go
Flonja: Because they left?

server/world/view_layer.go
RestartFU: `delete(viewer.viewers, v.e) `or something I think, e being the owner of the view layer

server/session/session.go
UnknownOre: what if multiple sessions has the same view layer

Flonja: 1 session = 1 view layer

T14Raptor: Then why do view layers store multiple sessions at the moment?

RestartFU: every player has it's own layer, so that each player can be different, only to a certain player

Flonja:
I feel like I should revise the naming for this... (`ViewLayer`, `Layer` and `LayerViewer`)
It gets confusing quickly and I doubt the user using this api would understand too

server/session/world.go
```diff
@@ -908,9 +908,22 @@ func (s *Session) ViewEntityAction(e world.Entity, a world.EntityAction) {
 
 // ViewEntityState ...
 func (s *Session) ViewEntityState(e world.Entity) {
+	metadata := s.parseEntityMetadata(e)
+	if v, ok := e.(LayerViewer); ok {
+		if nt := s.viewLayer.NameTag(v); len(nt) > 0 {
+			metadata[protocol.EntityDataKeyName] = nt
+		}
```
AndreasHGK, line 915: What about removing the nametag entirely? (for non-player entities)

RestartFU: wdym?

server/session/view_layer.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/df-mc/dragonfly/server/world"
+
+type LayerViewer interface {
+	ViewLayer() *world.ViewLayer
+}
+
+func (s *Session) ViewLayer() *world.ViewLayer {
+	return s.viewLayer
+}
```
RestartFU, line 11: can we add comments here

server/world/view_layer.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type LayerViewer interface {
+	ViewLayer() *ViewLayer
+}
+
+type Layer struct {
+	nameTag    string
+	visibility VisibilityLevel
+}
```
RestartFU, line 14: same here, comments please

server/world/visibility_level.go
```diff
@@ -0,0 +1,28 @@
+package world
+
+type VisibilityLevel struct {
+	visibility
+}
```
RestartFU, line 5: comment

server/session/world.go
andreashgk: I think what I meant is if the length of the nametag is 0, remove the nametag from the entity

this behavior would be better if the string was optional however where nil doesnt change the nametag and empty string removes it, imo

edit: i remember now, i was just asking if we could add a way to unset a nametag

PR #847, opened 2024-03-04
Update decorated_pot.go

daft0175:


daft0175:
And Added Ender chest close/open sound packet*

PR #855, opened 2024-03-14
Added function to compare chunks (TAKE 2) (Ethan seems to be dead)

RestartFU:
This allows for two chunks with two different pointers to still be compared for equality. (TAKE 2)

### DIFF
--- a/server/world/chunk/chunk.go
+++ b/server/world/chunk/chunk.go
@@ -2,6 +2,7 @@ package chunk
 
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
+	"slices"
 )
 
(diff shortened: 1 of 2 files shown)


server/world/chunk/chunk.go
```diff
@@ -42,6 +43,25 @@ func New(air uint32, r cube.Range) *Chunk {
 	}
 }
 
+// Equals returns if the chunk passed is equal to the current one
+func (chunk *Chunk) Equals(c *Chunk) bool {
+	if !chunk.recalculateHeightMap && !c.recalculateHeightMap && !slices.Equal(c.heightMap, chunk.heightMap) {
```
RestartFU, line 48: not sure if this should be ```chunk.recalculateHeightMap != c.recalculateHeightMap``` instead

PR #858, opened 2024-03-23
Implement totems

xNatsuri:
This PR implements the totem item

as well as implementing the TotemUseAction and the Totem sound


DaPigGuy (changes requested):

server/player/player.go
```diff
@@ -587,6 +587,20 @@ func (p *Player) Hurt(dmg float64, src world.DamageSource) (float64, bool) {
 			damageLeft = 0
 		}
 	}
+
+	if p.Health()-damageLeft < 0 {
```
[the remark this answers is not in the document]
bliszkitten: Shouldn't it be <= mgl64.Epsilon?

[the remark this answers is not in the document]
xNatsuri: fixed

server/player/player.go
```diff
@@ -587,6 +587,22 @@ func (p *Player) Hurt(dmg float64, src world.DamageSource) (float64, bool) {
...
+		hand, offHand := p.HeldItems()
+		if _, ok := hand.Item().(item.Totem); ok {
+			p.applyTotemEffects()
+			p.SetHeldItems(hand.Grow(-1), offHand)
+			return damageLeft, true
+		} else if _, ok := offHand.Item().(item.Totem); ok {
+			p.applyTotemEffects()
+			p.SetHeldItems(hand, offHand.Grow(-1))
+			return damageLeft, true
```
bliszkitten, line 600: These returns are not necessary because the player will not receive invulnerability ticks, will not play damage action, as well as possible damage to armour and retaliatory damage from the "thorns" enchantment.

xNatsuri:
bump

server/item/totem.go
```diff
@@ -0,0 +1,13 @@
+package item
+
+type Totem struct{}
```
TwistedAsylumMC, line 3: I'd just document this struct, explain what the item is. Doesn't need to be much

server/player/player.go
```diff
@@ -623,6 +637,25 @@ func (p *Player) Hurt(dmg float64, src world.DamageSource) (float64, bool) {
...
+func (p *Player) applyTotemEffects() {
+	p.addHealth(1 - p.Health())
+
+	for _, e := range p.Effects() {
+		p.RemoveEffect(e.Type())
+	}
+
+	p.AddEffect(effect.New(effect.Regeneration{}, 2, time.Second*40))
+	p.AddEffect(effect.New(effect.FireResistance{}, 1, time.Second*40))
+	p.AddEffect(effect.New(effect.Absorption{}, 1, time.Second*5))
```
TwistedAsylumMC, line 650: This is meant to give absorption 2 as per https://minecraft.wiki/w/Totem_of_Undying#Effect
> then grants 40‌[[BE](https://minecraft.wiki/w/Bedrock_Edition) only] or 45‌[[JE](https://minecraft.wiki/w/Java_Edition) only] seconds[[2]](https://minecraft.wiki/w/Totem_of_Undying#cite_note-2) of [Regeneration](https://minecraft.wiki/w/Regeneration) II, 40 seconds of [Fire Resistance](https://minecraft.wiki/w/Fire_Resistance) I and 5 seconds of [Absorption](https://minecraft.wiki/w/Absorption) II

server/player/player.go
```diff
@@ -623,6 +637,25 @@ func (p *Player) Hurt(dmg float64, src world.DamageSource) (float64, bool) {
 	return totalDamage, true
 }
 
+// applyTotemEffects is an unexported function that is used to handle totem effects.
+func (p *Player) applyTotemEffects() {
+	p.addHealth(1 - p.Health())
```
TwistedAsylumMC, line 642: What is the logic here? Is it not supposed to just be `p.addHealth(1)`?

TwistedAsylumMC (changes requested):

server/player/player.go
xNatsuri: technically yes but since I'm calling it before the player gets hurt / killed then I have to do it like this if not they would keep the health they had before they got hurt

server/player/player.go
```diff
@@ -584,9 +584,24 @@ func (p *Player) Hurt(dmg float64, src world.DamageSource) (float64, bool) {
...
+		hand, offHand := p.HeldItems()
+		if _, ok := hand.Item().(item.Totem); ok {
+			p.applyTotemEffects()
+			p.SetHeldItems(hand.Grow(-1), offHand)
+			return 0, false
+		} else if _, ok := offHand.Item().(item.Totem); ok {
+			p.applyTotemEffects()
+			p.SetHeldItems(hand, offHand.Grow(-1))
+			return 0, false
+		}
```
TwistedAsylumMC, line 602: Vanilla actually checks the offhand first for a totem, so we should probably match that behaviour too

xNatsuri:
everything should be good now.

PR #861, opened 2024-04-13
Smithing templates + Table and Armour Trims

xNatsuri:
Co-authored-by: Flonja

Everything works properly but the smithing recipes are still missing
would like some help generating the smithing_data.nbt file.


server/item/armour_trim.go
```diff
@@ -0,0 +1,48 @@
...
+
+type ArmourTrim struct {
+	Template ArmourTrimTemplate
+	Material TrimMaterial
+}
+
+type TrimMaterial interface {
+	// TrimMaterial returns the material name used for reading and writing trim data.
+	TrimMaterial() string
+	// MaterialColor returns the color code used for internal text formatting. Use text.Colourf for proper formatting.
```
TwistedAsylumMC, line 13: You recommend to use text.Colourf although you only return the formatting character rather than the section symbol also, I wonder if we should expand gophertunnel's text package to create exported constants for each colour

server/item/armour_trim.go
```diff
@@ -0,0 +1,48 @@
...
+	MaterialColor() string
```
TwistedAsylumMC, line 14: Also please use British English when possible, so MaterialColour() rather than MaterialColor()

server/item/armour_trim.go
```diff
@@ -0,0 +1,48 @@
...
+		return NetherQuartz{}
+	}
+
+	//TODO: add redstone material once pr is merged
+
+	panic("should not happen")
+}
+
+// TrimMaterials returns all the items that can be trim materials.
+func TrimMaterials() []world.Item {
```
TwistedAsylumMC, line 46: I'd call this method ArmourTrimMaterials() to give context for what the trims are for

server/item/register.go
```diff
@@ -145,6 +145,10 @@ func init() {
 		world.RegisterItem(SplashPotion{Type: p})
 		world.RegisterItem(Potion{Type: p})
 	}
+	world.RegisterItem(Template{TemplateNetheriteUpgrade()})
+	for _, t := range Templates() {
+		world.RegisterItem(Template{Template: t})
+	}
```
TwistedAsylumMC, line 151: Why is TemplateNetheriteUpgrade() separate?

server/item/stack.go
```diff
@@ -28,6 +28,8 @@ type Stack struct {
 	data map[string]any
 
 	enchantments map[EnchantmentType]Enchantment
+
+	armorTrim *ArmourTrim
```
TwistedAsylumMC, line 32: Please use armourTrim for the variable name to match the type

server/item/stack.go
```diff
@@ -42,6 +44,23 @@ func NewStack(t world.Item, count int) Stack {
...
+
+	s.armorTrim = &trim
+	return s
+}
+
+// ArmourTrim returns the ArmourTrim.
+// if this returns nil it does not have an armor trim, or it is not an armour
+func (s Stack) ArmourTrim() *ArmourTrim {
+	return s.armorTrim
+}
```
TwistedAsylumMC, line 62: Try and avoid exposing pointers in cases like this, you already do it in the WithArmourTrim() method, but I'd just adapt ArmourTrim() to return (ArmourTrim, bool) to stop the user having to handle nil pointers

server/item/template.go
```diff
@@ -0,0 +1,16 @@
+package item
+
+// Template is an item used in smithing tables to alter tools and armor.
+// They are consumed when used, but can be duplicated using an existing template, its material and diamonds.
+type Template struct {
```
TwistedAsylumMC, line 5: I'd call this a SmithingTemplate, since Template alone is rather vague when inside of the item package

server/item/template_type.go
```diff
@@ -0,0 +1,133 @@
+package item
+
+type ArmourTrimTemplate struct {
```
TwistedAsylumMC, line 3: Is ArmourTrimTemplate an appropriate name if the netherite upgrade template can be used for tools as well? Also try and stick to dragonfly's enum convention [as seen here](https://github.com/df-mc/dragonfly/blob/master/server/item/banner_pattern_type.go)

server/session/player.go
```diff
@@ -135,6 +135,34 @@ func (s *Session) sendRecipes() {
...
+		})
+	}
+
+	for _, i := range item.TrimMaterials() {
+		if material, ok := i.(item.TrimMaterial); ok {
+			name, _ := i.EncodeItem()
+
+			trimMaterials = append(trimMaterials, protocol.TrimMaterial{
+				MaterialID: material.TrimMaterial(),
+				Colour:     "§" + material.MaterialColor(),
```
TwistedAsylumMC, line 157: Not a fan of hard coding the section symbol here, but if we do something like I mentioned in the comment on the interface then we can return this directly from the MaterialColour() method

server/item/armour_trim.go
```diff
@@ -0,0 +1,48 @@
...
+
+type TrimMaterial interface {
+	// TrimMaterial returns the material name used for reading and writing trim data.
+	TrimMaterial() string
+	// MaterialColor returns the color code used for internal text formatting. Use text.Colourf for proper formatting.
+	MaterialColor() string
+}
+
+// MaterialFromString returns a TrimMaterial from a string.
+func MaterialFromString(name string) TrimMaterial {
```
TwistedAsylumMC, line 18: I'd rename this function to start with the same name as the type it returns, so TrimMaterial from string or a different name if the interface is renamed

server/item/armour_trim.go
```diff
@@ -0,0 +1,48 @@
+package item
+
+import "github.com/df-mc/dragonfly/server/world"
+
+type ArmourTrim struct {
+	Template ArmourTrimTemplate
+	Material TrimMaterial
+}
+
+type TrimMaterial interface {
```
TwistedAsylumMC, line 10: I'd probably suggest renaming this to ArmourTrimMaterial, it's already done for templates and trims can only be applied to armour anyways

server/item/template_type.go
```diff
@@ -0,0 +1,133 @@
...
+	return ArmourTrimTemplate{"eye"}
+}
+
+// TemplateSpire returns the Spire Template.
+func TemplateSpire() ArmourTrimTemplate {
+	return ArmourTrimTemplate{"spire"}
+}
+
+// Templates returns all the Templates
+func Templates() []ArmourTrimTemplate {
```
TwistedAsylumMC, line 93: I'd also rename this function to be ArmourTrimTemplates(), once again it's a rather vague name in the item package

server/item/template_type.go
```diff
@@ -0,0 +1,133 @@
...
+	return []ArmourTrimTemplate{TemplateSentry(), TemplateVex(), TemplateWild(), TemplateCoast(), TemplateDune(), TemplateWayFinder(), TemplateRaiser(), TemplateShaper(), TemplateHost(), TemplateWard(), TemplateSilence(), TemplateTide(), TemplateSnout(), TemplateRib(), TemplateEye(), TemplateSpire()}
+}
+
+// TemplateFromString returns a template based on a string.
+func TemplateFromString(name string) ArmourTrimTemplate {
```
TwistedAsylumMC, line 98: Same as above, just rename to keep consistent

TwistedAsylumMC (changes requested):

server/item/armour_trim.go
xNatsuri: This is probably a good idea I could make a pr for this on gophertunnel

server/item/armour_trim.go
xNatsuri: Oops I did mean to use British English will fix

server/item/register.go
xNatsuri: Because there’s no reason users should be able to access it since the template just converts it to nether rite and if they do add the template this way it just feels redundant since they are able to just give them netherrite armor in the first place.

server/item/template_type.go
xNatsuri: still not sure if I like the new name I provided in the newest commit lmk if you have any other idea for the name for this.

server/item/recipe/recipe.go
```diff
@@ -31,22 +31,37 @@ func NewShapeless(input []Item, output item.Stack, block string) Shapeless {
...
 
-// NewSmithing creates a new smithing recipe and returns it. The recipe can only be crafted on the block passed in the
-// parameters. If the block given a crafting table, the recipe can also be crafted in the 2x2 crafting grid in the
-// player's inventory.
-func NewSmithing(base, addition, template Item, output item.Stack, block string) Smithing {
-	return Smithing{recipe: recipe{
+// NewSmithingTransform creates a new smithing recipe and returns it. The recipe can only be crafted on the block
+// passed in the parameters. If the block given a crafting table, the recipe can also be crafted in the 2x2 crafting
+// grid in the player's inventory.
+func NewSmithingTransform(base, addition, template Item, output item.Stack, block string) SmithingTransform {
```
DaPigGuy, line 42: The block parameter/the associated comment looks rather out of place, maybe this should be removed?

server/item/smithing_template_type.go
```diff
@@ -0,0 +1,202 @@
...
+func TemplateWayFinder() ArmourSmithingTemplate {
+	return ArmourSmithingTemplate{6}
+}
+
+// TemplateRaiser returns the Raiser Template.
+func TemplateRaiser() ArmourSmithingTemplate {
+	return ArmourSmithingTemplate{7}
+}
+
+// TemplateShaper returns the Raiser Template.
```
DaPigGuy, line 47:
```suggestion
// TemplateShaper returns the Shaper Template.
```

DaPigGuy (changes requested):

server/item/armour_trim.go
```diff
@@ -0,0 +1,58 @@
...
+
+type ArmourTrim struct {
+	Template ArmourSmithingTemplate
+	Material ArmourTrimMaterial
+}
+
+type ArmourTrimMaterial interface {
+	// TrimMaterial returns the material name used for reading and writing trim data.
+	TrimMaterial() string
+	// MaterialColour returns the color code used for internal text formatting. Use text.Colourf for proper formatting.
```
TwistedAsylumMC, line 13: Change "color" to "colour", can probably remove the `text.Colourf` reference now we return the constant

server/item/recipe/recipe.go
```diff
@@ -31,22 +31,37 @@ func NewShapeless(input []Item, output item.Stack, block string) Shapeless {
...
 }
 
+// SmithingTrim represents a recipe only craftable on a smithing table using an armour trim.
+type SmithingTrim struct {
+	recipe
+}
+
+// NewSmithingTrim creates a new smithing trim recipe and returns it. The recipe can only be crafted on the block passed
+// in the parameters. If the block given a crafting table, the recipe can also be crafted in the 2x2 crafting grid in
+// the player's inventory. This is almost identical to SmithingTransform except there is no output item.
```
TwistedAsylumMC, line 57: Same thing about the crafting grid here

server/item/smithing_template.go
```diff
@@ -0,0 +1,16 @@
+package item
+
+// SmithingTemplate is an item used in smithing tables to alter tools and armor.
```
TwistedAsylumMC, line 3: Use "armour" instead of "armor"

server/item/stack.go
```diff
@@ -42,6 +44,26 @@ func NewStack(t world.Item, count int) Stack {
 	return Stack{item: t, count: count, id: newID()}
 }
 
+// WithArmourTrim returns a new stack with the ArmorTrim passed.
```
TwistedAsylumMC, line 47: "ArmourTrim" instead of "ArmorTrim"

server/item/stack.go
```diff
@@ -42,6 +44,26 @@ func NewStack(t world.Item, count int) Stack {
...
+	if _, ok := s.item.(Armour); !ok {
+		return s
+	}
+
+	s.armourTrim = &trim
+	return s
+}
+
+// ArmourTrim returns the ArmourTrim.
+// if this returns false it does not have an armor trim, or it is not an armour
```
TwistedAsylumMC, line 59: Grammar could be better here, maybe "ArmourTrim returns the ArmourTrim and true if present, otherwise false is returned."

server/session/handler_smithing.go
```diff
@@ -11,6 +11,8 @@ const (
 	smithingInputSlot = 0x33
 	// smithingMaterialSlot is the slot index of the material in the smithing table.
 	smithingMaterialSlot = 0x34
+	// smithingTemplateSlot is the slot index of the template item in the smithing table
```
TwistedAsylumMC, line 14: End with a "." for proper punctuation

TwistedAsylumMC (changes requested):

TwistedAsylumMC:
Thank you for the contribution!

PR #866, opened 2024-04-28
fix staticcheck

RestartFU:


TwistedAsylumMC:
Fixed in https://github.com/df-mc/dragonfly/commit/d21051cfe4f7a8831f0f13eec9279cf91d1929f8, didn't see this first

PR #868, opened 2024-05-02
TrampleCrop Handler

xNatsuri:
This PR implements TrampleCrop handler to prevent players from trampling crops.


Sandertv (changes requested):
I think it might be more suitable to have a HandleCropTrample method in `world.Handler`. That way we don't have to duplicate this method for every entity (because ultimately, all entities can trample crops). That would mean calling this event in the EntityLand method of `block.Farmland`.

xNatsuri:
I was looking through issues when I saw this and I saw the discussion that you and Tal both thought it would be better to have it in a player/entity handler

Sandertv:
Good point, let me think about it a bit more.

server/player/handler.go
```diff
@@ -68,6 +68,8 @@ type Handler interface {
 	// HandleBlockPick handles the player picking a specific block at a position in its world. ctx.Cancel()
 	// may be called to cancel the block being picked.
 	HandleBlockPick(ctx *event.Context, pos cube.Pos, b world.Block)
+	// HandleTrampleCrop handles the player trampling a crop.
+	HandleTrampleCrop(ctx *event.Context)
```
Sandertv, line 72: Would maybe rename this to HandleCropTrample to be more in line with the other method names.

Sandertv (commented):
I think it's fine to have this as a player handler. Got one more comment naming wise, but otherwise all is good once conflicts are resolved.

xNatsuri:
should be good to merge

xNatsuri:
added position of crop users might find it useful if they just want to cancel it in a certain area

Sandertv (approved):
Looks good! Thanks for the PR!

PR #872, opened 2024-05-22
double chest

xNatsuri:
Working Double Chest (built off Raptors PR)


RestartFU:
double chests do not drop when broken in survival

RestartFU:
> double chests do not drop when broken in survival

fixed

server/player/player.go
```diff
@@ -1835,22 +1835,42 @@ func (p *Player) BreakBlock(pos cube.Pos) {
...
+				} else if slot > 26 && pos == right {
+					drops = append(drops, i)
+				}
+			}
+		} else {
+			// If the block is a container, it should drop its inventory contents regardless whether the
+			// player is in creative mode or not.
+			drops = container.Inventory().Items()
+			container.Inventory().Clear()
+		}
```
Sandertv, line 1868: Does this code for the drops need to be here? Can we not put this in the BreakInfo of Chest?

xNatsuri: it does not i just had it here because this is where the other container drops were

server/block/chest.go
```diff
@@ -155,6 +193,24 @@ func (c Chest) DecodeNBT(data map[string]any) any {
...
 	c.Facing = facing
 	c.CustomName = nbtconv.String(data, "CustomName")
+
+	pairX, ok := data["pairx"]
+	pairZ, ok2 := data["pairz"]
+	if ok && ok2 {
+		c.paired = true
+		// TODO: type assertion checks
+		c.pairX, c.pairZ = int(pairX.(int32)), int(pairZ.(int32))
+		c.pairInv = inventory.New(54, func(slot int, _, item item.Stack) {
```
T14Raptor, line 203: Why is this logic different than what's done in the pair method?

RestartFU:
oh my fucking god
I fucked it up

PR #875, opened 2024-05-25
implement a way to remove armour trims + bug fix with trims

xNatsuri:
This PR adds the WithoutArmourTrim func and fixes an issue with ArmourSmithingTemplateFromString func not having the rib template

### DIFF
--- a/server/item/stack.go
+++ b/server/item/stack.go
@@ -2,12 +2,13 @@ package item
 
 import (
 	"fmt"
-	"github.com/df-mc/dragonfly/server/world"
 	"reflect"
 	"slices"
 	"sort"
 	"strings"
 	"sync/atomic"
+
+	"github.com/df-mc/dragonfly/server/world"
 )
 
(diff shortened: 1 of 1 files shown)


server/item/stack.go
```diff
@@ -50,11 +50,18 @@ func (s Stack) WithArmourTrim(trim ArmourTrim) Stack {
 	if _, ok := s.item.(Armour); !ok {
 		return s
 	}
-
 	s.armourTrim = &trim
 	return s
 }
 
+// WithoutArmourTrim returns the current stack but with the passed ArmourTrim removed.
+func (s Stack) WithoutArmourTrim(trim ArmourTrim) Stack {
```
Sandertv, line 58: Not sure it makes sense to require a trim being passed here, given that an item is only ever able to have one trim.

xNatsuri: that's true ill remove the trim.

DaPigGuy (approved):

PR #879, opened 2024-05-26
strictly remove arrow if it has been picked up

### LINKED ISSUE #824: Spectators can pick up arrows
That's all I have to say

UnknownOre:
fixes #824

### DIFF
--- a/server/entity/projectile.go
+++ b/server/entity/projectile.go
@@ -222,8 +222,7 @@ func (lt *ProjectileBehaviour) tryPickup(e *Ent) {
 			continue
 		}
 
-		c := collector.Collect(lt.conf.PickupItem)
-		if c == 0 {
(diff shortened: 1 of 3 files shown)


server/entity/projectile.go
```diff
@@ -221,6 +221,11 @@ func (lt *ProjectileBehaviour) tryPickup(e *Ent) {
 		if !ok {
 			continue
 		}
+
+		if collector.Collect(lt.conf.PickupItem) == 0 {
```
DaPigGuy, line 225: This breaks collecting the projectile when the item stack is empty, which will be intentional behavior for projectiles shot by a skeleton, drowned, etc.

DaPigGuy (changes requested):

UnknownOre:
should be good now

PR #880, opened 2024-05-28
Fix player skin rendering

Dasciam:
This PR fixes the behaviour of the player skin

Prior to this fix, other players who are not viewers of the player who had their skin changed did not receive a skin update

https://github.com/df-mc/dragonfly/blob/99f2abbe63701d902832443dd951f0af6d90334c/server/player/player.go#L247

### DIFF
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -76,6 +76,25 @@ func (s *Session) ViewEntity(e world.Entity) {
 	id := e.Type().EncodeEntity()
 	switch v := e.(type) {
 	case Controllable:
+		actualPlayer := false
+
+		sessionMu.Lock()
+		for _, s := range sessions {
+			if s.c.UUID() == v.UUID() {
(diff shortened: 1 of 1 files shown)


server/session/world.go
```diff
@@ -117,6 +117,8 @@ func (s *Session) ViewEntity(e world.Entity) {
 			s.writePacket(&packet.PlayerList{ActionType: packet.PlayerListActionRemove, Entries: []protocol.PlayerListEntry{{
 				UUID: v.UUID(),
 			}}})
+		} else {
+			s.ViewSkin(e)
```
DaPigGuy, line 121: Is there a reason not to send the skin regardless of whether the entity is an actual player?

Dasciam: The skin has already been set via PlayerList

DaPigGuy (approved):

PR #887, opened 2024-05-31
removed redundant chest type assertion

RestartFU:


RestartFU:
bumping this rq because I kinda need this atm, I want to implement hoppers externally.
I could still do it with very hacky stuff, but I'd rather just get this merged instead

RestartFU:
it would also allow me to remove [this](https://github.com/bedrock-gophers/inv/blob/master/inv/container.go#L77-L81) hacky feature from my inv menu library

Flonja:
instead of removing the whole check, maybe replace it with this?
```diff
		return s.armour.Inventory(), true
	case protocol.ContainerLevelEntity:
		if s.containerOpened.Load() {
			b := s.c.World().Block(s.openedPos.Load())
-			if _, chest := b.(block.Chest); chest {
+			if _, container := b.(block.Container); container {
				return s.openedWindow.Load(), true
			} else if _, enderChest := b.(block.EnderChest); enderChest {
				return s.openedWindow.Load(), true
			}
		}
	case protocol.ContainerBarrel:
		if s.containerOpened.Load() {
```

Sandertv:
Do hoppers not have a different ID entirely like barrels?

RestartFU:
> Do hoppers not have a different ID entirely like barrels?

yeah but I can change that quite easily with just a simple block check with a packet handler

RestartFU:
> Do hoppers not have a different ID entirely like barrels?



nope, as you can see here, this is what I did for my hopper PR, and it works just fine

RestartFU:
> > Do hoppers not have a different ID entirely like barrels?
> 
> 
> 
> nope, as you can see here, this is what I did for my hopper PR, and it works just fine

it is true that the containertype is different, but I could fix this myself on my side, anyway, this is about the redundant assert checks, I don't think it matters if hoppers have a different ID or not

PR #890, opened 2024-06-10
fix Abilities not working after changing gamemode

UnknownOre:


Sandertv (commented):
This looks like a really awful hack. Is there no other way to fix this that doesn't involve time.AfterFunc? How do other projects approach this?

UnknownOre:
> This looks like a really awful hack. Is there no other way to fix this that doesn't involve time.AfterFunc? How do other projects approach this?

looks like other projects send the abilities next tick

UnknownOre:
we could send the abilities in the background task every x ticks, this what pm does [here](https://github.com/pmmp/PocketMine-MP/blob/stable/src/network/mcpe/NetworkSession.php#L1304) though PM has a system to mark attributes that need updating

PR #895, opened 2024-07-09
feat: implement manual/runtime world saving

cqdetdev:
This PR implements the following:
- `(*world.World).Save()` to manually save worlds on invocation (essentially has the same logic to `(*world.World).close())`
- A change in the signature for `(*world.World).saveChunk`, this function now takes in a boolean parameter - `closeEntities` - which dictates whether or not entities in the given chunk should be closed after they're written.

Note: I realize that the `Save()` method is essentially the same as the `close`, I couldn't think of a way to merge the two functions that would semantically make any sense but I'm open to any suggestions.


Notsoraalex:
can you  summon the go thread of the repost towards the raspatory of the langauage

PR #901, opened 2024-07-17
player/player.go: added handler for fire extinguish

aabstractt:


didntpot (changes requested):
this seems incorrect

server/player/handler.go
```diff
@@ -55,6 +55,9 @@ type Handler interface {
 	// HandleSkinChange handles the player changing their skin. ctx.Cancel() may be called to cancel the skin
 	// change.
 	HandleSkinChange(ctx *event.Context, skin *skin.Skin)
+	// HandleFireExtinguish handles the player extinguishing a fire at a specific position. ctx.Cancel() may
+	// be called to cancel the fire being extinguished.
+	HandleFireExtinguish(ctx *event.Context, pos cube.Pos)
 	// HandleStartBreak handles the player starting to break a block at the position passed. ctx.Cancel() may
 	// be called to stop the player from breaking the block completely.
 	HandleStartBreak(ctx *event.Context, pos cube.Pos)
```
CJMustard1452, line 63: is there a reason for pushing cube.Pos? this seems un-needed

server/player/handler.go
```diff
@@ -152,6 +155,7 @@ func (NopHandler) HandleCommandExecution(*event.Context, cmd.Command, []string)
 func (NopHandler) HandleTransfer(*event.Context, *net.UDPAddr)                                {}
 func (NopHandler) HandleChat(*event.Context, *string)                                         {}
 func (NopHandler) HandleSkinChange(*event.Context, *skin.Skin)                                {}
+func (NopHandler) HandleFireExtinguish(*event.Context, cube.Pos)                              {}
```
CJMustard1452, line 158: same here

CJMustard1452 (commented):
EDITED: 
it didn't work due to my own error

aabstractt:
It seems to actually work as expected.

https://github.com/user-attachments/assets/b1f40282-9225-4613-b930-cc144e5db7b1

server/player/handler.go
aabstractt: I see it as necessary when wanting to cancel the fire extinguishment only in some positions, you cannot use the player's position, since both positions are different than expected.

PR #903, opened 2024-07-22
Feature/reset food saturation

aabstractt:
Add reset the exhaustion and saturation when the Food Loss handler is cancelled. This help to prevent loss very fast the food after no longer cancelling that handler.

Example

https://github.com/user-attachments/assets/ab8d9917-2c1e-4568-9b98-7da8e6bbb1d2

### DIFF
--- a/server/player/hunger.go
+++ b/server/player/hunger.go
@@ -66,7 +66,7 @@ func (m *hungerManager) Reset() {
 }
 
 // ResetExhaustion resets the exhaustion level of the player to 0.
-// Prevents the player food level decrease after non cancelling the food loss
+// Prevents the player's food level from decreasing after preventing food loss.
 func (m *hungerManager) resetExhaustion() {
 	m.mu.Lock()
 	defer m.mu.Unlock()
--- a/server/player/hunger.go
+++ b/server/player/hunger.go
@@ -65,6 +65,14 @@ func (m *hungerManager) Reset() {
 	m.mu.Unlock()
 }
 
+// ResetExhaustion resets the exhaustion level of the player to 0.
+// Prevents the player food level decrease after non cancelling the food loss
+func (m *hungerManager) ResetExhaustion() {
+	m.exhaustionLevel = 0
+	m.saturationLevel = 0
+	m.foodTick = 0
+}
+
 // exhaust exhausts the player by the amount of points passed. If the total exhaustion level exceeds 4, a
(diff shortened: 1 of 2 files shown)


didntpot (changes requested):
the logic here is wrong

didntpot:
race condition plz add mutex

CJMustard1452 (changes requested):
fix race conditions

server/player/hunger.go
```diff
@@ -65,6 +65,14 @@ func (m *hungerManager) Reset() {
 	m.mu.Unlock()
 }
 
+// ResetExhaustion resets the exhaustion level of the player to 0.
+// Prevents the player food level decrease after non cancelling the food loss
+func (m *hungerManager) resetExhaustion() {
```
didntpot, line 70: this function cause race condition plz fix

didntpot (changes requested):

server/player/hunger.go
CJMustard1452, line 70:
```suggestion
func (m *hungerManager) resetExhaustion() {
	m.mu.Lock()
	defer m.mu.Unlock()
```

CJMustard1452 (changes requested):

server/player/hunger.go
```diff
@@ -65,6 +65,16 @@ func (m *hungerManager) Reset() {
 	m.mu.Unlock()
 }
 
+// ResetExhaustion resets the exhaustion level of the player to 0.
```
didntpot, line 68:
```suggestion
// ResetExhaustion resets the player's exhaustion level to 0.```

File: server/player/hunger.go [hunk 3]
```diff
@@ -65,6 +65,16 @@ func (m *hungerManager) Reset() {
...
+// Prevents the player food level decrease after non cancelling the food loss
```
didntpot, line 69:
```suggestion
// Prevents the player's food level from decreasing after preventing food loss.
```

didntpot (changes requested):
plz

server/player/hunger.go
CJMustard1452: agreed.

CJMustard1452 (approved):
Looks good.

TwistedAsylumMC:
Should it reset it to 0 or what it was before the last event was cancelled?

Sandertv (approved):
I think this is a sensible change. Thank you for the PR!

PR #905, opened 2024-07-24
world/handler.go: Added handler for leaves decay

aabstractt:


server/block/leaves.go
```diff
@@ -66,6 +67,14 @@ func (l Leaves) RandomTick(pos cube.Pos, w *world.World, _ *rand.Rand) {
...
 		} else {
+			ctx := event.C()
+			if w.Handler().HandleLeavesDecay(ctx, pos); !ctx.Cancelled() {
+				drops := l.BreakInfo().Drops(nil, nil)
+				for _, drop := range drops {
+					dropItem(w, drop, pos.Vec3Centre())
+				}
+			}
+
 			w.SetBlock(pos, nil, nil)
```
TwistedAsylumMC, line 78: Should it not only set the block to nil if the event was not cancelled? I think the logic of only dropping items would be slightly unexpected?

PR #906, opened 2024-07-29
Implemented a new way of displaying the mob effects.

Dasciam:
With the new version 1.21, mob effect bubbles are no longer int32-packed RGBA. Now it is an array of bytes up to 8 elements packed in int64.

Before: https://youtu.be/k1qiewknQwQ
After: https://youtu.be/YI8ETqjhow0

https://github.com/sandertv/gophertunnel must be updated with the new entity metadata key.


TwistedAsylumMC:
Thanks!

PR #907, opened 2024-08-03
block/ladder.go: Change Direction to Face to allow states 0 and 1

Gewinum:
Fixed a bug where ladder wasn't being added to creative inventory. It was happening due to ladder in dragonfly not allowing facing_direction to be 0 and 1. We already discussed this in Discord:
image

### DIFF
--- a/server/block/ladder.go
+++ b/server/block/ladder.go
@@ -17,12 +17,12 @@ type Ladder struct {
 	sourceWaterDisplacer
 
 	// Facing is the side of the block the ladder is currently attached to.
-	Facing cube.Direction
+	Facing cube.Face
 }
 
 // NeighbourUpdateTick ...
 func (l Ladder) NeighbourUpdateTick(pos, _ cube.Pos, w *world.World) {
-	if _, ok := w.Block(pos.Side(l.Facing.Opposite().Face())).(LightDiffuser); ok {
+	if _, ok := w.Block(pos.Side(l.Facing.Opposite())).(LightDiffuser); ok {
 		w.SetBlock(pos, nil, nil)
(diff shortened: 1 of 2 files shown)


server/block/ladder.go
```diff
@@ -86,17 +86,17 @@ func (l Ladder) EncodeItem() (name string, meta int16) {
...
 // Model ...
 func (l Ladder) Model() world.BlockModel {
-	return model.Ladder{Facing: l.Facing}
+	return model.Ladder{Facing: cube.Direction(l.Facing)}
 }
 
 // allLadders ...
 func allLadders() (b []world.Block) {
-	for i := cube.Direction(0); i <= 3; i++ {
+	for i := cube.Face(0); i <= 5; i++ {
```
Gewinum, line 99: @TwistedAsylumMC shall I use cube.Faces() instead of this?

TwistedAsylumMC: Yep

server/block/ladder.go
```diff
@@ -17,12 +17,12 @@ type Ladder struct {
 	sourceWaterDisplacer
 
 	// Facing is the side of the block the ladder is currently attached to.
```
TwistedAsylumMC, line 19: I'd probably leave a comment saying cube.FaceUp and cube.FaceDown do not do anything. Do just check in game first that it is the case

Gewinum: in game the block appears but it is invisible and somehow you can still climb it

Gewinum:
wait ive just noticed a typo in my commit name... why did i write commend instead of commend.....

Gewinum:
womp womp

Gewinum:
I went to the aquarium this weekend, but I didn’t stay long. There’s something fishy about that place.

PR #913, opened 2024-08-18
Fixed hopper insert behaviour in composter

ipad54:
In vanilla, hopper can insert item in composter only if it's level is less than 7


server/block/composter.go
```diff
@@ -24,7 +24,7 @@ type Composter struct {
 
 // InsertItem ...
 func (c Composter) InsertItem(h Hopper, pos cube.Pos, w *world.World) bool {
-	if c.Level == 8 {
+	if c.Level >= 7 {
```
DaPigGuy, line 27: Looks like we're also missing a check for the hopper's facing direction

PR #915, opened 2024-09-01
Migrate from df-mc/atomic to sync/atomic

TwistedAsylumMC:
[df-mc/atomic](https://github.com/df-mc/atomic) was forked with the release of generics to add atomic.Value[T], and since Go 1.19 everything we used [uber-go/atomic](https://github.com/uber-go/atomic) for has since been implemented in some form in `sync/atomic`. This PR removes the `df-mc/atomic` dependency and that repo will be archived when this is merged.


Sandertv (approved):
Looks good to me! Feel free to merge if you've just ran dragonfly once and all seems good.

PR #918, opened 2024-09-02
Make block hashes more dynamic to avoid registry collisions

TwistedAsylumMC:
Currently the number of bits given to each block is fixed from `cmd/blockhash`, which limits the number of custom blocks that can be registered before there are hash collisions. This pull request makes it so the number of bits is dynamically allocated at runtime to account for every block that has been registered before the server starts to avoid collisions completely.


server/world/block.go
```diff
@@ -91,12 +102,49 @@ func RegisterBlock(b Block) {
...
+	}
+	bitSize = bits.Len64(uint64(len(blocks)))
+	sort.SliceStable(blocks, func(i, j int) bool {
+		nameOne, _ := blocks[i].EncodeBlock()
+		nameTwo, _ := blocks[j].EncodeBlock()
+		return nameOne != nameTwo && fnv1.HashString64(nameOne) < fnv1.HashString64(nameTwo)
+	})
+	for rid, b := range blocks {
+		finaliseBlock(uint32(rid), b)
+		if b.Hash() != math.MaxUint64 {
```
DaPigGuy, line 130: unknownBlock

PR #920, opened 2024-09-08
session: fix packet handling race condition

### LINKED ISSUE #919: Attacks can bypass the last immunity tick
I saw on my server that some players could do a lot more damage than their opponents by using some tricks.

In the video my opponent was supposed to get damage once every 9 ticks at the beginning, but that doesn't happen when I timed the hits and the damage was once every 8 ticks if I'm not mistaken

Video: https://youtu.be/2ig0boHn9wc

Dasciam:
Closes #919

All sessions when processing a packet wait for the tick in the world the Controllable is in, this is when packet processing occurs in the session

### DIFF
--- a/server/session/session.go
+++ b/server/session/session.go
@@ -313,15 +313,23 @@ func (s *Session) handlePackets() {
 		_ = s.Close()
 	}()
 	for {
-		pk, err := s.conn.ReadPacket()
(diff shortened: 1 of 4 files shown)


Dasciam:
Closed due to inefficient use of resources

PR #923, opened 2024-10-02
Add world.BlockSource interface to allow external libraries to use DF block models

ethaniccc:


Sandertv (commented):
Hi, thanks for the PR! I think this is a good idea, but I would consider moving `world.BlockSource` to `model.BlockSource`, as that is where the interface is used. Otherwise looks good.

server/world/block_source.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "github.com/df-mc/dragonfly/server/block/cube"
+
+// BlockSource represents a source for blocks. Blocks can be retrieved
+// and set in the block source.
+type BlockSource interface {
+	// Block returns the block at the given position in the block source.
+	Block(cube.Pos) Block
+	// SetBlock sets the block at the given position in the block source.
+	SetBlock(cube.Pos, Block, *SetOpts)
```
Sandertv, line 11: I believe SetBlock isn't used anywhere in this package, so I think you could remove this method from the interface.

ethaniccc:
> Hi, thanks for the PR! I think this is a good idea, but I would consider moving `world.BlockSource` to `model.BlockSource`, as that is where the interface is used. Otherwise looks good.

That won't be possible without moving `world.BlockModel`, as it would cause an import cycle with `model` <-> `world`

Sandertv (approved):
Looks good, sensible change! Thank you for the PR!

PR #924, opened 2024-10-03
Documentation correction: Toggles dont contain text as far as I know

theaddonn:
> // Default is the default value filled out in the input. The user may remove this value and fill out its
	// own text. The text may contain Minecraft formatting codes.

This was likely copied over from the Input's default field.

I discovered this while reading df's form implementation to help implement and especially document my own in Rust. (Thanks for all the comments in both df and gophertunnel, they are really helpful and I did not copy the comments 1:1 for the record)

This is a giant pr, so be warned!


TwistedAsylumMC:
Thanks!

PR #925, opened 2024-10-15
enchantment/flame.go: Arrow flame duration is always five seconds

Dasciam:
The flame duration has changed from some version to five seconds


Sandertv (approved):
Good find, 100 seconds on fire is pretty unlucky.

PR #926, opened 2024-10-15
block/lava.go: set on fire for 8 seconds instead of 15.

Dasciam:
On fire time after being inside the lava changed from some version to 8 seconds


DaPigGuy:
When was this changed?

Dasciam:
> When was this changed?

Like I said I don't know what version it was changed in, but you can test it on the newest version and see that on fire time is clearly not 15 seconds

Dasciam:
Also sorry for the confusion, changed the PR description

Dasciam:
Closed as this is a bug

PR #931, opened 2024-11-13
player/type.go: fix network offset

Dasciam:
Fixes players falling under blocks

Reference: https://github.com/pmmp/PocketMine-MP/blob/09bf203267aca9b83a09640e17bfac0fc7b58ff8/src/entity/Human.php#L528


Sandertv (approved):
Thank you!

PR #933, opened 2024-11-15
player/player.go: fix PickBlock nil pointer

Dasciam:


server/player/player.go
```diff
@@ -1888,7 +1888,10 @@ func (p *Player) PickBlock(pos cube.Pos) {
 	if pi, ok := b.(block.Pickable); ok {
 		pickedItem = pi.Pick()
 	} else if i, ok := b.(world.Item); ok {
-		it, _ := world.ItemByName(i.EncodeItem())
+		it, ok := world.ItemByName(i.EncodeItem())
```
DaPigGuy, line 1891: This is a workaround. `EncodeItem` should always be returning a valid registered item id & meta.

DaPigGuy (changes requested):

server/player/player.go
Dasciam: I swear to you, it crashed my server. For some reason it doesn't

Sandertv: That sounds like an issue to debug. Could you find out when that happens?

PR #936, opened 2024-11-16
implement crawling

FDUTCH:


server/block/cube/bbox.go
```diff
@@ -282,3 +282,26 @@ func (box BBox) ZOffset(nearby BBox, deltaZ float64) float64 {
 	}
 	return deltaZ
 }
+
+// Corners returns slice of all hitbox corners
```
TwistedAsylumMC, line 286: Please add a full stop at the end of this doc and the other 2 in this file.

server/block/cube/bbox.go
```diff
@@ -282,3 +282,26 @@ func (box BBox) ZOffset(nearby BBox, deltaZ float64) float64 {
...
+func (box BBox) Corners() []mgl64.Vec3 {
+	min := box.min
+	max := box.max
+	xyZ := mgl64.Vec3{min[0], min[1], max[2]}
+	xYz := mgl64.Vec3{min[0], max[1], min[2]}
+	xYZ := mgl64.Vec3{min[0], max[1], max[2]}
+	XYz := mgl64.Vec3{max[0], max[1], min[2]}
+	XyZ := mgl64.Vec3{max[0], min[1], max[2]}
+	Xyz := mgl64.Vec3{max[0], min[1], min[2]}
+	return []mgl64.Vec3{min, max, xyZ, xYz, xYZ, XYz, XyZ, Xyz}
```
TwistedAsylumMC, line 296: I'd remove these variables and do something like this:
```go
return []mgl64.Vec3{
  {min[0], min[1], max[2]},
  {min[0], max[1], min[2]},
  // etc.
}```

File: server/player/player.go [hunk 3]
```diff
@@ -1044,6 +1044,37 @@ func (p *Player) StopSwimming() {
...
+	for _, corner := range p.Type().BBox(p).Translate(p.Position()).Corners() {
+		_, isAir = p.World().Block(cube.PosFromVec3(corner).Add(cube.Pos{0, 1, 0})).(block.Air)
+		if !isAir {
+			break
+		}
+	}
+
+	if !isAir && !p.crawling.CompareAndSwap(false, true) {
+		return
+	}
```
TwistedAsylumMC, line 1060: This can be slightly simplified to
```go
for _, corner := range p.Type().BBox(p).Translate(p.Position()).Corners() {
	_, isAir = p.World().Block(cube.PosFromVec3(corner).Add(cube.Pos{0, 1, 0})).(block.Air)
	if !isAir {
		if !p.crawling.CompareAndSwap(false, true) {
			return
		}
		break
	}
}
```

server/player/player.go
```diff
@@ -2615,12 +2646,16 @@ func (p *Player) OnGround() bool {
...
-	if p.swimming.Load() {
+	switch {
+	case p.swimming.Load() || p.crawling.Load():
 		return 0.52
+	case p.sneaking.Load():
+		return 1.26
+	default:
+		return 1.62
 	}
-	return 1.62
```
TwistedAsylumMC, line 2623: No point changing this to a default cause, keep how it was before

server/player/player.go
```diff
@@ -2615,12 +2646,16 @@ func (p *Player) OnGround() bool {
 	return p.onGround.Load()
 }
 
-// EyeHeight returns the eye height of the player: 1.62, or 0.52 if the player is swimming.
+// EyeHeight returns the eye height of the player.
```
TwistedAsylumMC, line 2649: Could keep the last part of the doc and say "0.52 if the player is crawling or swimming, 1.26 when sneaking or 1.62 when standing."

TwistedAsylumMC (changes requested):
Definitely looks good! Just a few little comments

server/block/cube/bbox.go
```diff
@@ -282,3 +282,29 @@ func (box BBox) ZOffset(nearby BBox, deltaZ float64) float64 {
...
+		mgl64.Vec3{max[0], min[1], min[2]},
+	}
+}
+
+// Mul performs a scalar multiplication for min, max points of BBox.
+func (box BBox) Mul(val float64) BBox {
+	return BBox{min: box.min.Mul(val), max: box.max.Mul(val)}
+}
+
+// Volume calculates Volume of the BBox.
```
TwistedAsylumMC, line 307: Little nitpick, second "Volume" makes more sense to be "the volume"

server/player/player.go
```diff
@@ -2617,8 +2646,11 @@ func (p *Player) OnGround() bool {
 
 // EyeHeight returns the eye height of the player: 1.62, or 0.52 if the player is swimming.
```
TwistedAsylumMC, line 2647: Would be nice if you could update this doc as previously mentioned to describe all states

TwistedAsylumMC (changes requested):

TwistedAsylumMC:
Thanks!

PR #666, opened 2022-09-08
Implementation of NPC Dialogues

abimek:
Decided to make a new api instead of using the forms api, more flexible in the future.

PR #937, opened 2024-11-17
player/dialogue: Basic implementation of NPC dialogues

TwistedAsylumMC:
Kind of a continuation from https://github.com/df-mc/dragonfly/pull/666 however a lot of it is different.

Whilst this code works, we need a better method of implementing the `dialogueProvider` interface to allow entities such as players (often used for NPCs) to be a provider. We could just make it always implement the interface but I'm open to other ideas. There are also a few names I am not sure about, like the `CanShowDialogue()` method to implement this interface.

I have also decided to completely hide the idea of "opening" and "closing" buttons, they only exist in Minecraft as a way of detecting opening/closing of the menus which we can already achieve through calling `SendDialogue()` for opening and implementing `dialogue.Closer` for closing.

Example usage in combination with df-mc/npc where p is a `*player.Player` (modifying Player to implement dialogueProvider):

```go
type TestDialogue struct {
	Button1 dialogue.Button
	Button2 dialogue.Button
	Button3 dialogue.Button
}

func (t TestDialogue) Submit(s dialogue.Submitter, pressed dialogue.Button) {
	fmt.Println("Pressed", pressed)
	s.CloseDialogue() // The client will not close the menu automatically
}

func (t TestDialogue) Close(s dialogue.Submitter) {
	fmt.Println("Closed")
}
```
```go
settings := npc.Settings{
	Name:     "Test NPC",
	Position: p.Position(),
	Scale:    1,
	Skin:     p.Skin(),
}
n := npc.Create(settings, p.World(), nil)

d := dialogue.New(TestDialogue{
	Button1: dialogue.Button{Text: "Button 1"},
	Button2: dialogue.Button{Text: "Button 2"},
	Button3: dialogue.Button{Text: "Button 3"},
}, "Dialogue Title").
	WithBody("Welcome to my dialogue!")
p.SendDialogue(d, n)
```


Sandertv (commented):
I really like how this looks so far. Simple, clean, similar to existing APIs. Thumbs up for that! I've left some suggestions on how to update it to use transactions.

Additionally, I believe we discussed some things on Discord about how the entity should be shown in the dialogue, and how sending (altered) metadata before a dialogue to achieve the settings could solve it. Not sure if you wanted that in this PR too?

DaPigGuy (approved):

Sandertv (approved):

PR #938, opened 2024-11-17
implement beds , respawn anchors

FDUTCH:


server/block/respawn_anchor.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/df-mc/dragonfly/server/world/sound"
+	"github.com/sandertv/gophertunnel/minecraft/text"
+)
+
+// RespawnAnchor is a block , allows players to set there respawn point in the nether.
+type RespawnAnchor struct {
+	solid
+	bassDrum
+	charge int32
```
TwistedAsylumMC, line 15: This field should be exported and be just a regular int

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// EncodeBlock ...
+func (r RespawnAnchor) EncodeBlock() (string, map[string]any) {
+	return "minecraft:respawn_anchor", map[string]any{"respawn_anchor_charge": r.charge}
+}
+
+// BreakInfo ...
+func (r RespawnAnchor) BreakInfo() BreakInfo {
+	return newBreakInfo(35, func(t item.Tool) bool {
```
TwistedAsylumMC, line 35: Hardness should be 50

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,104 @@
...
+		return t.ToolType() == item.TypePickaxe && t.HarvestLevel() >= item.ToolTierDiamond.HarvestLevel
+	}, pickaxeEffective, oneOf(r)).withBlastResistance(6000)
```
TwistedAsylumMC, line 37: Blast resistance should be 1200

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Hash ...
+func (r RespawnAnchor) Hash() (uint64, uint64) {
+	return hashRespawnAnchor, uint64(r.charge)
+}
+
+// LightEmissionLevel ...
+func (r RespawnAnchor) LightEmissionLevel() uint8 {
+	return (uint8(r.charge) + 1) * 3
```
TwistedAsylumMC, line 20: This logic is incorrect based on the wiki:
> The first charge makes the respawn anchor glow with a light level of 3. Each glowstone after the first increases the light level by 4, up to a maximum of 15

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,99 @@
...
+	}, pickaxeEffective, oneOf(r)).withBlastResistance(6000)
+}
+
+// Activate ...
+func (r RespawnAnchor) Activate(pos cube.Pos, clickedFace cube.Face, w *world.World, u item.User, ctx *item.UseContext) bool {
+
+	var nether bool = w.Dimension().WaterEvaporates()
+
+	held, _ := u.HeldItems()
+	_, usingGlowStone := held.Item().(Glowstone)
```
TwistedAsylumMC, line 46: Name it usingGlowstone just to be consistent with the type

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// allRespawnAnchorsItems returns all possible respawn anchors as items.
+func allRespawnAnchorsItems() []world.Item {
+	all := make([]world.Item, 0, 5)
+	for i := int32(0); i < 5; i++ {
+		all = append(all, RespawnAnchor{charge: i})
+	}
+	return all
+}
```
TwistedAsylumMC, line 90: I'd just stick this for loop in the register.go file at the bottom of the second init, no other place has a allXXXItems() method

server/player/player.go
```diff
@@ -910,8 +923,14 @@ func (p *Player) kill(src world.DamageSource) {
 			// We have an actual client connected to this player: We change its position server side so that in
 			// the future, the client won't respawn on the death location when disconnecting. The client should
 			// not see the movement itself yet, though.
-			newPos := w.Spawn().Vec3()
-			p.pos.Store(&newPos)
+			pos, w, blockHasBeenBroken := p.realSpawnPos()
+			if blockHasBeenBroken {
+				//bl := w.Block(pos).(block.SpawnBlock)
+				//bl.Update(pos, p, w)
```
TwistedAsylumMC: Is this meant to be commented out still?

server/player/player.go
```diff
@@ -940,7 +959,14 @@ func (p *Player) dropContents() {
 // Respawn spawns the player after it dies, so that its health is replenished and it is spawned in the world
 // again. Nothing will happen if the player does not have a session connected to it.
 func (p *Player) Respawn() {
-	w := p.World()
+	position, w, ok := p.realSpawnPos()
+	if bl, ok := w.Block(position).(block.SpawnBlock); ok {
```
TwistedAsylumMC, line 962: Should probably just add `&& b.SpawnBlock()` just for consistency

server/player/player.go
```diff
@@ -964,6 +989,26 @@ func (p *Player) Respawn() {
...
 
+// SpawnPos returns spawn position in the current world.
+func (p *Player) SpawnPos() cube.Pos {
+	return p.World().PlayerSpawn(p.uuid)
+}
+
+// SetSpawnPos sets spawn position for the player in the provided world.
+func (p *Player) SetSpawnPos(pos cube.Pos, w *world.World) {
+	w.SetPlayerSpawn(p.UUID(), pos)
+}
```
TwistedAsylumMC, line 999: Does it make sense for SpawnPos() to return the current world and then SetSpawnPos() take in any world? Might be worth making SpawnPos() also take in a world just to be consistent

server/player/player.go
```diff
@@ -964,6 +989,26 @@ func (p *Player) Respawn() {
...
+
+// realSpawnPos returns position and world where player should be spawned.
+func (p *Player) realSpawnPos() (cube.Pos, *world.World, bool) {
```
TwistedAsylumMC, line 1002: a better name for this would be great, maybe just something as simple as "spawnPos". Improving the docs to mention what the returned bool indicates would also be great

server/player/player.go
```diff
@@ -1160,6 +1205,75 @@ func (p *Player) Jump() {
 	}
 }
 
+// Sleep makes the player sleep at the given position. If the position does not map to a bed (specifically the head side),
+// the player will not sleep.
```
TwistedAsylumMC, line 1208: We try not to go over 110 characters in a single line of docs
```suggestion
// Sleep makes the player sleep at the given position. If the position does not map to a bed (specifically
the head side), the player will not sleep.
```

server/world/sound/block.go
```diff
@@ -188,6 +188,24 @@ type PotionBrewed struct{ sound }
 // LecternBookPlace is a sound played when a book is placed in a lectern.
 type LecternBookPlace struct{ sound }
 
+// RespawnAnchorCharge is a sound played when a respawn anchor has been charged.
+type RespawnAnchorCharge struct {
+	Charge int32
+	sound
```
TwistedAsylumMC, line 194: Please embed sound before the Charge field for this struct and the two below, i'd also make Charge an int and cast it where necessary

server/world/tick.go
```diff
@@ -52,6 +52,13 @@ func (t ticker) tick() {
 	}
 
 	rain, thunder, tick, tim := t.w.set.Raining, t.w.set.Thundering && t.w.set.Raining, t.w.set.CurrentTick, int(t.w.set.Time)
+	sleep := false
+	if t.w.set.RequiredSleepTicks > 0 {
+		t.w.set.RequiredSleepTicks--
+		if t.w.set.RequiredSleepTicks-1 <= 0 {
```
TwistedAsylumMC, line 58: Is this logic correct? You're decrementing the ticks and then checking if the tick below is <= 0?

server/world/world.go
```diff
@@ -806,6 +816,41 @@ func (w *World) Entities() []Entity {
...
+func (w *World) Sleepers() []Sleeper {
+	ent := w.Entities()
+	sleepers := make([]Sleeper, 0, len(ent)/40)
+	for _, e := range ent {
+		if s, ok := e.(Sleeper); ok {
+			sleepers = append(sleepers, s)
+		}
+	}
+	return sleepers
+}
```
TwistedAsylumMC, line 829: Is there any reason not to repurpose this method to return only those that are actively sleeping?

server/player/player.go
TwistedAsylumMC, line 999: I also don't see a reason why these methods need to exist if the caller already has access to the player and world? They can just call PlayerSpawn/SetPlayerSpawn directly on the world

server/player/player.go
```diff
@@ -3109,3 +3224,11 @@ func (p *Player) resendBlock(pos cube.Pos, w *world.World) {
...
 	return strings.TrimSuffix(strings.TrimSuffix(fmt.Sprintln(a...), "\n"), "\n")
 }
+
+// DimensionProvider provides access to all 3 dimensions.
+type DimensionProvider interface {
+	World() *world.World
+	Nether() *world.World
+	End() *world.World
+	Players() []*Player
+}
```
TwistedAsylumMC, line 3233: This interface feels like it's very limiting for worlds, what happens when a server has multiworld support?

TwistedAsylumMC (changes requested):
Noticed a few problems after a quick look

server/world/world.go
FDUTCH: check here: https://github.com/df-mc/dragonfly/pull/938/files#diff-91f78c90dff5aa1562e35a0f1760be7de0398829c40343a40e1b973555bceb62R26-R42
and here: https://github.com/df-mc/dragonfly/pull/938/files#diff-30888af30ee265b9debd16a1ae7820be78a1f73450bd8378c914d1e45dc6f47bR832-R853

server/player/player.go
FDUTCH: SetSpawnPos used here: https://github.com/df-mc/dragonfly/pull/938/files#diff-b1e7a7f21487969b0d5980153a20003b2e51eecd41a8aab7d1215f9432cbad40R118
and here: https://github.com/df-mc/dragonfly/pull/938/files#diff-0c77a3c2d3771dbd25a7b738b13d7f35d3871ad3a5f241c192c5fc242eea02e5R58

server/player/player.go
FDUTCH: I agree, but how do I determine which world to teleport the player to after death?

server/player/player.go
TwistedAsylumMC: Yes you can just do w.SetPlayerSpawn(u.UUID(), pos) ?

server/player/player.go
TwistedAsylumMC: I think the default world is appropriate for that, but maybe add a respawn event that lets you modify the new world and position?

FDUTCH: df already has a respawn handler
https://github.com/df-mc/dragonfly/blob/0637012e5ea28136785feb3f437de2fb7c4d2460/server/player/handler.go#L55

FDUTCH: Is there something I left to do?

server/player/player.go
```diff
@@ -1160,6 +1197,75 @@ func (p *Player) Jump() {
...
+
+// Wake forces the player out of bed if they are sleeping.
+func (p *Player) Wake() {
+	if !p.sleeping.CompareAndSwap(true, false) {
+		return
+	}
+
+	w := p.World()
+	w.SetRequiredSleepDuration(0)
```
TwistedAsylumMC, line 1240: Is this correct? Wake() can be called from many places such as a player leaving, a bed being destroyed etc.

server/player/player.go
```diff
@@ -940,7 +961,14 @@ func (p *Player) dropContents() {
...
 func (p *Player) Respawn() {
-	w := p.World()
+	position, w, ok := p.realSpawnPos()
+	if bl, ok := w.Block(position).(block.SpawnBlock); ok {
+		bl.Update(position, p, w)
+	}
+	pos := position.Vec3Middle()
+	if ok {
+		p.Messaget("%tile.bed.notValid")
+	}
```
TwistedAsylumMC, line 971: Should this ok check not be moved up to under line 964 where it's originally defined? Should it not also return?

TwistedAsylumMC (changes requested):
Still a few of my previous comments that have not been addressed, found some more issues after another look through

server/session/handler_player_action.go
```diff
@@ -23,14 +23,15 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
 		return errSelfRuntimeID
 	}
 	switch action {
-	case protocol.PlayerActionRespawn, protocol.PlayerActionDimensionChangeDone:
-		// Don't do anything for these actions.
```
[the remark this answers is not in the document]
FDUTCH: yes

server/world/sleep.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Sleeper represents an entity that can sleep.
+type Sleeper interface {
+	Entity
+
+	UUID() uuid.UUID
+	Name() string
+
+	Message(a ...any)
+	Messaget(key string, a ...string)
```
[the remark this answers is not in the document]
FDUTCH: actually only some of them are not needed

server/player/player.go
FDUTCH: it should not return otherwise the player would not respawn

TwistedAsylumMC: Ah yeah I misunderstood the code, the first haf of my message should still be acted on though

FDUTCH: already done

server/item/item.go
```diff
@@ -166,6 +167,11 @@ type User interface {
 	UsingItem() bool
 	ReleaseItem()
 	UseItem()
+
+	UUID() uuid.UUID
+
+	Messaget(key string, a ...string)
+	Sneaking() bool
```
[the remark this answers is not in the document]
TwistedAsylumMC: This is still not resolved

server/player/player.go
```diff
@@ -907,11 +917,19 @@ func (p *Player) kill(src world.DamageSource) {
 		}
 		if p.Dead() {
 			p.SetInvisible()
+
 			// We have an actual client connected to this player: We change its position server side so that in
 			// the future, the client won't respawn on the death location when disconnecting. The client should
 			// not see the movement itself yet, though.
-			newPos := w.Spawn().Vec3()
-			p.pos.Store(&newPos)
+			pos, w, blockHasBeenBroken, _ := p.realSpawnPos()
```
TwistedAsylumMC, line 924: I think "spawnObstructed" is a better name for this

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,100 @@
...
+
+// allRespawnAnchorsItems returns all possible respawn anchors as items.
+func allRespawnAnchorsItems() []world.Item {
+	all := make([]world.Item, 0, 5)
+	for i := 0; i < 5; i++ {
+		all = append(all, RespawnAnchor{Charge: i})
+	}
+	return all
+}
```
TwistedAsylumMC, line 91: Please remove this like I previously asked, just put this for loop in register.go for items specifically

server/player/player.go
```diff
@@ -964,6 +989,22 @@ func (p *Player) Respawn() {
...
+	pos = w.PlayerSpawn(p.UUID())
+	if b, ok := w.Block(pos).(block.RespawnBlock); ok && b.CanSpawn() {
+		return pos, w, false, previousDimension
+	}
+
+	// We can use the principle here that returning through a portal of a specific dimension inside that dimension will
+	// always bring us back to the overworld.
+	w = w.PortalDestination(w.Dimension())
+	return w.Spawn(), w, true, previousDimension
+}
```
TwistedAsylumMC, line 1006: The name and documentation for this method is slightly misleading since nowhere does it state it actually calls w.Spawn()

server/session/handler_player_action.go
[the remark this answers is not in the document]
TwistedAsylumMC: Could you please commit that change?

TwistedAsylumMC (changes requested):
I have had another look, there are still unresolved comments that I have previously provided. I am not going to review this PR anymore until every single comment has been acted on or you've replied with a valid counter argument. I have gone through and resolved all the comments you have addressed but there are still many that remain

server/block/bed.go
```diff
@@ -0,0 +1,242 @@
...
+	return true
+}
+
+func (Bed) Update(pos cube.Pos, u item.User, w *world.World) {}
+
+// SpawnBlock represents a block using which player can set his spawn point.
+type SpawnBlock interface {
+	SpawnBlock() bool
+	Update(pos cube.Pos, u item.User, w *world.World)
+}
```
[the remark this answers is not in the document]
TwistedAsylumMC: You have changed the method name but the documentation is still needing to be updated

server/world/sleep.go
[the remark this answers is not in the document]
TwistedAsylumMC: Could you remove any that are not needed?

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,100 @@
...
+		return false
+	}
+
+	if r.Charge > 0 {
+		w.SetBlock(pos, nil, nil)
+		ExplosionConfig{
+			Size:      5,
+			SpawnFire: true,
+		}.Explode(w, pos.Vec3Centre())
+	}
```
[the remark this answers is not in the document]
DaPigGuy: The `r.Charge > 0` condition is common, merge and check `r.Charge > 0` first

server/block/bed.go
```diff
@@ -0,0 +1,248 @@
...
+	headSide, headPos, ok := b.head(pos, w)
+	if !ok {
+		return false
+	}
+	if _, ok = w.Liquid(headPos); ok {
+		return false
+	}
+
+	previousSpawn := w.PlayerSpawn(u.UUID())
+	if previousSpawn != pos {
```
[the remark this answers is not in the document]
DaPigGuy: Do not update the player spawn position or send the message if it's the other half of the bed

DaPigGuy (changes requested):

server/player/player.go
FDUTCH: ok, I'll call this function ¨spawnLocation¨

server/player/player.go
```diff
@@ -907,11 +917,19 @@ func (p *Player) kill(src world.DamageSource) {
...
 			// We have an actual client connected to this player: We change its position server side so that in
 			// the future, the client won't respawn on the death location when disconnecting. The client should
 			// not see the movement itself yet, though.
-			newPos := w.Spawn().Vec3()
-			p.pos.Store(&newPos)
+			pos, w, blockHasBeenBroken, _ := p.realSpawnPos()
+			vec := pos.Vec3()
+
+			if blockHasBeenBroken {
+				p.Handler().HandleRespawn(&vec, &w)
```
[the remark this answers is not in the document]
FDUTCH: ¨make a new event, i.e. respawn change¨  this can bee done using custom world.Provider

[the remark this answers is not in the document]
DaPigGuy: The respawn event should not be emitted here. The player is not actually respawning here.

server/block/bed.go
```diff
@@ -0,0 +1,248 @@
...
+	o, ok := w.Block(sidePos).(Bed)
+	return o, sidePos, ok
+}
+
+// allBeds returns all possible beds.
+func allBeds() (beds []world.Block) {
+	for _, d := range cube.Directions() {
+		beds = append(beds, Bed{Facing: d})
+		beds = append(beds, Bed{Facing: d, Head: true})
+	}
```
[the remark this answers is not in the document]
DaPigGuy: Bed color is set through the tile, not a block state. Please undo your change.

If all beds are white in the creative inventory, it is probably the same issue with the creative item entries missing NBT that banners have. See a1c98dab589f4874733dc4ad0c50fcee746e9215

server/player/player.go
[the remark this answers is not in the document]
FDUTCH: soo, should I remove the event call from this function?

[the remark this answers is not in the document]
DaPigGuy: Yes

DaPigGuy:
Please fix merge conflicts

FDUTCH:
> Please fix merge conflicts

bro , I'm dead ☠️
I don't know how to resolve merge conflict

server/block/bed.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// BreakInfo ...
+func (b Bed) BreakInfo() BreakInfo {
+	return newBreakInfo(0.2, alwaysHarvestable, nothingEffective, oneOf(b)).withBreakHandler(func(pos cube.Pos, w *world.Tx, _ item.User) {
+		headSide, _, ok := b.head(pos, w)
+		if !ok {
+			return
+		}
+		headSide.Sleeper.Wake()
```
xNatsuri, line 50: Need to check if Sleeper is nil before calling Wake() to avoid a nil pointer.
```suggestion
                sleeper := headSide.Sleeper
                if sleeper != nil {
                    sleeper.Wake()
		}
```

xNatsuri (changes requested):
most of the pr looks good just from a quick glance saw this issue

server/block/bed.go
```diff
@@ -0,0 +1,238 @@
...
+	})
+}
+
+// UseOnBlock ...
+func (b Bed) UseOnBlock(pos cube.Pos, face cube.Face, _ mgl64.Vec3, w *world.Tx, user item.User, ctx *item.UseContext) (used bool) {
+	if pos, _, used = firstReplaceable(w, pos, face, b); !used {
+		return
+	}
+	if _, ok := w.Block(pos.Side(cube.FaceDown)).Model().(model.Solid); !ok {
```
Sandertv, line 59: I don't think this check is correct. Beds can be placed on half slabs (top slot) for example. I think checking if the top face is solid is more accurate.

server/block/bed.go
```diff
@@ -0,0 +1,238 @@
...
+
+	b.Facing = user.Rotation().Direction()
+
+	side, sidePos := b, pos.Side(b.Facing.Face())
+	side.Head = true
+
+	if !replaceableWith(w, sidePos, side) {
+		return
+	}
+	if _, ok := w.Block(sidePos.Side(cube.FaceDown)).Model().(model.Solid); !ok {
```
Sandertv, line 71: Same as above. Should check if the top face is solid.

server/block/bed.go
```diff
@@ -0,0 +1,238 @@
...
+func (Bed) CanSpawn() bool {
+	return true
+}
+
+func (Bed) SpawnOn(pos cube.Pos, u item.User, w *world.Tx) {}
+
+// RespawnBlock represents a block using which player can set his spawn point.
+type RespawnBlock interface {
+	// CanSpawn defines if player can use this block to respawn.
+	CanSpawn() bool
```
Sandertv, line 235: I would rename this method to make it more clear that this is about respawning.

server/block/bed.go
```diff
@@ -0,0 +1,238 @@
...
+	// SpawnOn is called when a player decides to respawn using this block.
+	SpawnOn(pos cube.Pos, u item.User, tx *world.Tx)
```
Sandertv, line 237: Same as above.

server/player/handler.go
```diff
@@ -122,6 +122,8 @@ type Handler interface {
 	// HandleSignEdit handles the player editing a sign. It is called for every keystroke while editing a sign and
 	// has both the old text passed and the text after the edit. This typically only has a change of one character.
 	HandleSignEdit(ctx *Context, pos cube.Pos, frontSide bool, oldText, newText string)
+	// HandleSleep handles the player going to sleep. ctx.Cancel() may be called to cancel the sleep.
```
Sandertv, line 125: Should make this documentation clearer. Does "going to sleep" mean finishing sleeping or starting sleeping?

server/player/player.go
```diff
@@ -54,8 +55,11 @@ type playerData struct {
 
 	sneaking, sprinting, swimming, gliding, crawling, flying,
 	invisible, immobile, onGround, usingItem bool
-	usingSince time.Time
 
+	sleeping atomic.Bool
+	sleepPos atomic.Pointer[cube.Pos]
+
```
Sandertv, line 61: These should not have to be atomic.

server/session/session.go
```diff
@@ -254,6 +254,12 @@ func (s *Session) close(tx *world.Tx, c Controllable) {
 	s.closeCurrentContainer(tx)
 	s.chunkLoader.Close(tx)
 
+	if ent, ok := s.ent.Entity(tx); ok {
+		if c, ok := ent.(Controllable); ok {
+			c.Wake()
+		}
+	}
+
```
Sandertv, line 262: No reason for this entire block here. You can simply call `c.Wake()` without calling `s.ent.Entity(tx)`. `c` already is the player.

server/world/sleep.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/google/uuid"
+)
+
+// Sleeper represents an entity that can sleep.
+type Sleeper interface {
+	Entity
+
+	Name() string
+	UUID() uuid.UUID
+
```
Sandertv, line 14: I couldn't see any usage of these methods. Try removing any methods from interfaces that aren't used.

Sandertv (commented):
Thanks for the PR! I've left some more comments.

server/world/sleep.go
FDUTCH: {4CD39432-01BB-40D1-BABB-28279562CF7A}
{E97426C8-1488-4E8B-8449-BE6A16C17F29}
but these methods are used

Sandertv: Good point! Thanks for clarifying.

DaPigGuy:
Please fix merge conflicts and update to use the translation changes from #961 for the bed messages

server/block/bed.go
```diff
@@ -0,0 +1,249 @@
...
+		s.Messaget(chat.MessageBedIsOccupied)
+		return true
+	}
+
+	s.Sleep(headPos)
+	return true
+}
+
+// EntityLand ...
+func (b Bed) EntityLand(_ cube.Pos, _ *world.World, e world.Entity, distance *float64) {
```
DaPigGuy, line 144:
```suggestion
func (b Bed) EntityLand(_ cube.Pos, _ *world.Tx, e world.Entity, distance *float64) {
```

server/block/register.go
```diff
@@ -473,6 +476,9 @@ func init() {
 			world.RegisterItem(Copper{Type: c, Oxidation: o, Waxed: true})
 		}
 	}
+	for c := range 5 {
+		world.RegisterItem(RespawnAnchor{Charge: c})
```
DaPigGuy, line 480: We typically do not register items that are never accessible in the vanilla game. Only a RespawnAnchor with Charge 0 should be registered.

server/block/bed.go
```diff
@@ -0,0 +1,249 @@
...
+	return 1
+}
+
+// Model ...
+func (Bed) Model() world.BlockModel {
+	return model.Bed{}
+}
+
+// SideClosed ...
+func (Bed) SideClosed(cube.Pos, cube.Pos, *world.World) bool {
```
DaPigGuy, line 39:
```suggestion
func (Bed) SideClosed(cube.Pos, cube.Pos, *world.Tx) bool {
```

server/block/bed.go
```diff
@@ -0,0 +1,249 @@
...
+
+		sleeper := headSide.Sleeper
+		if sleeper != nil {
+			sleeper.Wake()
+		}
+	})
+}
+
+// UseOnBlock ...
+func (b Bed) UseOnBlock(pos cube.Pos, face cube.Face, _ mgl64.Vec3, w *world.Tx, user item.User, ctx *item.UseContext) (used bool) {
```
DaPigGuy, line 59: Please rename the `w` parameter to `tx` for consistency

server/player/player.go
```diff
@@ -1147,6 +1187,75 @@ func (p *Player) Jump() {
...
+		return
+	}
+	tx := p.tx
+
+	if b, ok := tx.Block(pos).(block.Bed); ok {
+		if b.Sleeper != nil {
+			// The player cannot sleep here.
+			return
+		}
```
DaPigGuy, line 1204:
```suggestion
```

This is the responsibility of `block.Bed`. Exiting early here is also awkard because the sleep event has already been emitted by this point.

server/player/player.go
```diff
@@ -1147,6 +1187,75 @@ func (p *Player) Jump() {
...
+		b.Sleeper = p
```
DaPigGuy, line 1205: This should now be an `world.EntityHandle` due to the world transactions changes.

server/player/player.go
```diff
@@ -1147,6 +1187,75 @@ func (p *Player) Jump() {
 	}
 }
 
+// Sleep makes the player sleep at the given position. If the position does not map to a bed (specifically the head side),
+// the player will not sleep.
+func (p *Player) Sleep(pos cube.Pos) {
+	ctx, sendReminder := event.C(p), true
```
DaPigGuy, line 1193: This should exit early if the player is already sleeping.

DaPigGuy (changes requested):

server/player/player.go
FDUTCH: I don't understand what I should I do, explain me please

FDUTCH: done

server/block/register.go
```diff
@@ -88,6 +88,7 @@ func init() {
 	world.RegisterBlock(ResinBricks{Chiseled: true})
 	world.RegisterBlock(ResinBricks{})
 	world.RegisterBlock(Resin{})
+	world.RegisterItem(RespawnAnchor{})
```
DaPigGuy, line 91: This should be with the other `world.RegisterItem` calls

server/world/tick.go
```diff
@@ -60,6 +60,15 @@ func (t ticker) tick(tx *Tx) {
 	}
 
 	rain, thunder, tick, tim := w.set.Raining, w.set.Thundering && w.set.Raining, w.set.CurrentTick, int(w.set.Time)
+
+	sleep := false
```
DaPigGuy, line 64: This variable is unnecessary

server/player/player.go
```diff
@@ -1147,6 +1187,81 @@ func (p *Player) Jump() {
...
+}
+
+// Wake forces the player out of bed if they are sleeping.
+func (p *Player) Wake() {
+	if !p.sleeping {
+		return
+	}
+	p.sleeping = false
+
+	tx := p.Tx()
```
DaPigGuy, line 1236:
```suggestion
	tx := p.tx
```
for consistency

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/df-mc/dragonfly/server/player/chat"
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/df-mc/dragonfly/server/world/sound"
+)
+
+// RespawnAnchor is a block that allows the player to set their spawn point in the Nether.
+type RespawnAnchor struct {
+	solid
+	bassDrum
+	Charge int
```
DaPigGuy, line 15: Document this

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// LightEmissionLevel ...
+func (r RespawnAnchor) LightEmissionLevel() uint8 {
+	return uint8(max(0, 3+4*(r.Charge-1)))
+}
+
+// EncodeItem ...
+func (r RespawnAnchor) EncodeItem() (name string, meta int16) {
+	return "minecraft:respawn_anchor", int16(r.Charge)
```
DaPigGuy, line 25:
```suggestion
	return "minecraft:respawn_anchor", 0
```

Only meta=0 is registered

server/block/respawn_anchor.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if r.Charge > 0 {
+		if w.Dimension() == world.Nether {
+			previousSpawn := w.PlayerSpawn(sleeper.UUID())
+			if previousSpawn == pos {
+				return false
+			}
+			sleeper.Messaget(chat.MessageRespawnPointSet)
+			w.SetPlayerSpawn(sleeper.UUID(), pos)
+			return false
```
DaPigGuy, line 68:
```suggestion
			return true
```
to prevent block place, etc. when setting spawn

server/block/bed.go
```diff
@@ -0,0 +1,254 @@
...
+	// Velocity returns the current velocity of the entity.
+	Velocity() mgl64.Vec3
+	// SetVelocity sets the velocity of the entity.
+	SetVelocity(mgl64.Vec3)
+}
+
+// NeighbourUpdateTick ...
+func (b Bed) NeighbourUpdateTick(pos, _ cube.Pos, w *world.Tx) {
+	if _, _, ok := b.side(pos, w); !ok {
+		w.SetBlock(pos, nil, nil)
```
DaPigGuy, line 171:
```suggestion
		breakBlockNoDrops(b, pos, tx)
```

for break particles

server/block/bed.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// head returns the head side of the bed. If neither side is a head side, the third return value is false.
+func (b Bed) head(pos cube.Pos, w *world.Tx) (Bed, cube.Pos, bool) {
+	headSide, headPos, ok := b.side(pos, w)
+	if !ok {
+		return Bed{}, cube.Pos{}, false
+	}
+	if b.Head {
+		headSide, headPos = b, pos
```
DaPigGuy, line 210:
```suggestion
		return b, pos, true
```

server/block/register.go
```diff
@@ -209,6 +210,7 @@ func init() {
 	registerAll(allCopperDoors())
 	registerAll(allCopperGrates())
 	registerAll(allCopperTrapdoors())
+	registerAll(allRespawnAnchors())
```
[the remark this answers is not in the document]
FDUTCH: ok, but then all the copper things should be higher too

FDUTCH:
after making Bed.Sleeper as *world.EntityHandle appeared dead locks when placing, breaking and sleeping

server/block/bed.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		sleeper := headSide.Sleeper
+		if sleeper != nil {
+			sleeper.ExecWorld(func(tx *world.Tx, e world.Entity) {
+				sleeper, ok := e.(world.Sleeper)
+				if ok {
+					sleeper.Wake()
+				}
+			})
```
DaPigGuy, line 58: The break handler is being called within a transaction, so instead of opening another transaction,

```go
e, ok := headSide.Sleeper.Entity(tx)
// etc ...
```

FDUTCH: thank you, all deadlocks seems to be disappeared

server/block/bed.go
```diff
@@ -0,0 +1,256 @@
...
+func (b Bed) BreakInfo() BreakInfo {
+	return newBreakInfo(0.2, alwaysHarvestable, nothingEffective, oneOf(b)).withBreakHandler(func(pos cube.Pos, tx *world.Tx, _ item.User) {
+		headSide, _, ok := b.head(pos, tx)
+		if !ok {
+			return
+		}
+
+		sleeper := headSide.Sleeper
+		if sleeper != nil {
+
```
TwistedAsylumMC, line 53: Unnecessary whitespace

server/item/item.go
```diff
@@ -13,7 +13,7 @@ import (
 // MaxCounter represents an item that has a specific max count. By default, each item will be expected to have
 // a maximum count of 64. MaxCounter may be implemented to change this behaviour.
 type MaxCounter interface {
-	// MaxCount returns the maximum number of items that a stack may be composed of. The number returned must
+	// MaxCount returns the maximum number of items a stack may be composed of. The number returned must
```
TwistedAsylumMC, line 16: This change is irrelevant, please remove it

server/player/player.go
```diff
@@ -1167,6 +1207,81 @@ func (p *Player) Jump() {
 	}
 }
 
+// Sleep makes the player sleep at the given position. If the position does not map to a bed (specifically the head side),
+// the player will not sleep.
+func (p *Player) Sleep(pos cube.Pos) {
+
```
TwistedAsylumMC, line 1213: Unnecessary whitespace

server/player/player.go
```diff
@@ -1167,6 +1207,81 @@ func (p *Player) Jump() {
...
+	ctx, sendReminder := event.C(p), true
+	if p.Handler().HandleSleep(ctx, &sendReminder); ctx.Cancelled() {
+		return
+	}
+
+	b.Sleeper = p.H()
+	tx.SetBlock(pos, b, nil)
+
+	tx.World().SetRequiredSleepDuration(time.Second * 5)
```
TwistedAsylumMC, line 1235: This should probably be `time.Millisecond * 5050`
> A player sleeps in a bed for 101 in-game [ticks](https://minecraft.wiki/w/Tick), or 5.05 seconds before the time skips to the next day.

server/block/bed.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	previousSpawn := w.PlayerSpawn(s.UUID())
+	if previousSpawn != pos && previousSpawn != sidePos {
+		w.SetPlayerSpawn(s.UUID(), pos)
+		s.Messaget(chat.MessageRespawnPointSet)
+	}
+
+	time := w.Time() % world.TimeFull
+	if (time < world.TimeNight || time >= world.TimeSunrise) && !tx.ThunderingAt(pos) {
```
TwistedAsylumMC, line 137: This logic does not match the behaviour documented on the wiki:
> A player sleeps by using a bed during a [thunderstorm](https://minecraft.wiki/w/Thunderstorm), or at [night](https://minecraft.wiki/w/Night) (between 12542 and 23459 ticks in clear weather, when stars appear in the sky, or between 12010 and 23991 ticks in rainy weather). Players can sleep during a thunderstorm even if they are in a biome where it does not rain (i.e. desert)

server/player/player.go
```diff
@@ -957,6 +980,23 @@ func (p *Player) respawn(f func(p *Player)) {
...
+		return playerSpawn, w, false, previousDimension
+	}
+
+	// We can use the principle here that returning through a portal of a specific dimension inside that dimension will
+	// always bring us back to the overworld.
+	w = w.PortalDestination(w.Dimension())
+	worldSpawn := w.Spawn()
+	return worldSpawn, w, playerSpawn != worldSpawn, previousDimension
+}
+
```
TwistedAsylumMC, line 999: This function does not return accurate spawn positions. block.RespawnBlock should have a method which returns the position where the player should respawn since it's not at the same place where the bed/respawn anchor is located.

https://minecraft.wiki/w/Bed#Setting_the_spawn_point
https://minecraft.wiki/w/Respawn_Anchor#Respawning

TwistedAsylumMC (changes requested):
I have had another look over and found some inconsistencies with your implementation compared to vanilla, or at least what is documented on the wiki. If any of this is incorrect for Bedrock then I am happy to ignore those. There are also still multiple unresolved comments that myself and @DaPigGuy have left previously, please make sure you scroll through https://github.com/df-mc/dragonfly/pull/938/files and look at any unresolved comment and either make the requested changes or reply with a counter reason

CJMustard1452:
@TwistedAsylumMC are these changes above all that need to be altered before this PR is ready for master?

I might be able to get these done relatively quickly

PR #806, opened 2023-07-17
Introduce entity animations

RoyalMCPE:
The documentation probably needs improvement. Also tie-in pr for the [wiki](https://github.com/df-mc/wiki/pull/7)
## Tests
```go
p.World().PlayAnimation(p, animation.New("animation.player.sit").WithController("controller.animation.player.sit").WithState("sit").WithStopCondition("query.is_jumping"))
```

Test resource pack:  [EntityAnimationTest.zip](https://github.com/df-mc/dragonfly/files/12066015/EntityAnimationTest.zip)

PR #940, opened 2024-11-18
world/viewer.go: Allow for more complex entity animations to be played

TwistedAsylumMC:
Continuation from https://github.com/df-mc/dragonfly/pull/806 with a few changes. I am still undecided whether this deserves to be its own package or if it's something we can move into the entity package and have `entity.NewAnimation(...)`


server/world/world.go
```diff
@@ -655,6 +655,17 @@ func (w *World) AddParticle(pos mgl64.Vec3, p Particle) {
 	}
 }
 
+// PlayEntityAnimation plays an animation on an entity in the world. The animation is played for all viewers
+// of the entity.
+func (w *World) PlayEntityAnimation(e Entity, a EntityAnimation) {
```
Sandertv, line 660: Would move this to `world/tx.go` and make it a method on `*world.Tx`.

server/world/world.go
```diff
@@ -655,6 +655,17 @@ func (w *World) AddParticle(pos mgl64.Vec3, p Particle) {
...
+	if w == nil {
+		return
+	}
```
Sandertv, line 663: This can be removed once you move this to tx.

Sandertv (commented):
Looks good, no comments other than some suggestions for moving this over to transactions.

DaPigGuy (approved):

Sandertv (approved):

PR #955, opened 2024-12-08
Add getter methods for weather states

RoyalMCPE:


Sandertv:
These methods don't already exist because weather is position bound. Any reason the RainingAt/ThunderingAt/SnowingAt methods don't suffice?

RoyalMCPE:
Right. Forgot about snow. We probably can just export those, but they don’t give the rain time/thunder time which is what I need so maybe just separate getters or something for that would be enough.

Sandertv:
What would you need the durations for?

PR #956, opened 2024-12-11
Implement shulker boxes

mmm545:
This PR aims to fully implement shulker boxes.

~~Only one thing needs to be addressed:~~
- [x] ~~Properly implement the bounding box for shulker boxes.~~
  - ~~Shulker boxes have a dynamically changing bounding box, it gradually expands as it opens, pushing entities in its path, and vise versa.~~


server/block/shulker_box_types.go
```diff
@@ -0,0 +1,205 @@
+package block
+
+// ShulkerBoxType represents a type of shulker box.
```
DaPigGuy, line 3: It would be nice if we could avoid this. Candles & bundles will also have a no color option.

server/block/shulker_box.go
```diff
@@ -0,0 +1,322 @@
...
+		target := shulkerBBox.Min().Z() - halfL
+		if z := entityPos.Z(); z > target {
+			living.MoveDelta(mgl64.Vec3{0, 0, target - z}, 0, 0)
+		}
+	}
+}
+
+// BreakInfo ...
+func (s ShulkerBox) BreakInfo() BreakInfo {
+	return newBreakInfo(2, alwaysHarvestable, pickaxeEffective, oneOf(s))
```
FDUTCH, line 254:
```suggestion
	return newBreakInfo(2, alwaysHarvestable, pickaxeEffective, simpleDrops()).withBreakHandler(func(pos cube.Pos, tx *world.Tx, u item.User) {
		if !(u != nil && s.inventory.Empty()) {
			dropItem(tx, item.NewStack(s, 1), pos.Vec3Centre())
		}
	})
```

HashimTheArab: changing to this would not preserve the block NBT which would lose the items

HashimTheArab:
this is ready for another review

server/block/shulker_box.go
FDUTCH: > changing to this would not preserve the block NBT which would lose the items

it won't, but I actually forgot to add gamemode check

HashimTheArab: The break-handler pattern is for blocks that spill contents or have side effects on break (chest
unpair, item frame, etc.). Shulker is the opposite — its drop is the block-with-inventory, which is
exactly oneOf(s). Moving to a handler would mean re-implementing creative/harvestable gating that the
Drops path gives us for free.

Let me know if this is incorrect

server/block/shulker_box.go
```diff
@@ -0,0 +1,322 @@
...
+
+const (
+	shulkerStateClosed int32 = iota
+	shulkerStateOpening
+	shulkerStateOpened
+	shulkerStateClosing
+)
+
+// shulkerLidSteps is the number of animation ticks between fully closed and fully open.
+const shulkerLidSteps int32 = 10
```
RestartFU:
```suggestion
const shulkerLidTicks int32 = 10
```

no?

server/block/shulker_box.go
```diff
@@ -0,0 +1,322 @@
...
+		viewers:         make(map[ContainerViewer]struct{}, 1),
+		progress:        new(atomic.Int32),
+		animationStatus: new(atomic.Int32),
+	}
+
+	s.inventory = inventory.New(27, func(slot int, _, after item.Stack) {
+		s.viewerMu.RLock()
+		defer s.viewerMu.RUnlock()
+		for viewer := range s.viewers {
+			// A shulker box inventory can't store shulker boxes, this is mostly handled by the client.
```
RestartFU: also handled by the server

```suggestion
			// A shulker box inventory can't store shulker boxes.
```

server/block/shulker_box.go
```diff
@@ -0,0 +1,322 @@
...
+	delete(s.viewers, v)
+	if len(s.viewers) == 0 {
+		s.close(tx, pos)
+	}
+}
+
+// Inventory returns the inventory of the shulker box.
+func (s ShulkerBox) Inventory(*world.Tx, cube.Pos) *inventory.Inventory {
+	return s.inventory
+}
```
RestartFU: aren't these all interface-implementing methods? usually we just put a comment like `WithName...`

server/player/player.go
```diff
@@ -2255,6 +2254,30 @@ func (p *Player) Move(deltaPos mgl64.Vec3, deltaYaw, deltaPitch float64) {
...
+
+	p.data.Pos = res
+	p.data.Rot = resRot
+	if deltaPos.Len() <= 3 {
+		// Only update velocity if the player is not moving too fast to prevent potential OOMs.
+		p.data.Vel = deltaPos
+		p.checkBlockCollisions(deltaPos)
+	}
+}
+
```
RestartFU: is this method really necessary? and also why are we not dispatching it to the move handler?

RestartFU: like can't this be done using Move?

HashimTheArab: BDS moves it using this delta packet, adding a handler instead

FDUTCH: maybe we can add check (on the session side) if player is moving (if Move is called from PlayerAuthInput handler), and if Move is called at a time when the player is not moving, send a MoveActorDelta to the player

didntpot (dismissed):

RestartFU (dismissed):

RestartFU (dismissed):

RestartFU (dismissed):

HashimTheArab:
Need to fix some motion values before merging this, only thing left

HashimTheArab:
tested and working

PR #982, opened 2025-01-04
world/world.go: clear viewers map on world close.

Dasciam:


RoyalMCPE:
From what I can tell this is already being done [here](https://github.com/df-mc/dragonfly/blob/0347c6e5b1ab1a6f1c9018400c7c360caf1ac97e/server/world/loader.go#L137). The world loops through the entities and calls the [Close](https://github.com/df-mc/dragonfly/blob/0347c6e5b1ab1a6f1c9018400c7c360caf1ac97e/server/world/world.go#L986) method and if that entity is a player it's session's [Close](https://github.com/df-mc/dragonfly/blob/0347c6e5b1ab1a6f1c9018400c7c360caf1ac97e/server/session/session.go#L252) method calls first method referenced.

Dasciam:
Hmm, that's right

PR #991, opened 2025-01-11
Add explosion handler for affected blocks and entities

Flonja:


FDUTCH:
I think world.Handler should have ability to handle explosions too.
https://github.com/df-mc/dragonfly/blob/master/server/world/handler.go

Flonja:
> I think world.Handler should have ability to handle explosions too. https://github.com/df-mc/dragonfly/blob/master/server/world/handler.go

I initially thought it'd be better to keep these handlers in the `ExplosionConfig`, since it's already a configuration.
But I believe having two handlers would be overkill, I'm however welcome to the idea of shifting these current "handlers" to `world/handler.go`.

PR #1000, opened 2025-01-25
fix: player/type.go: sneaking hitbox height.

Dasciam:


Dasciam:
Asked on Discord, I answer. Yes, this value was taken from the BDS

```go
pk, err := serverConn.ReadPacket()
...
switch pk := pk.(type) {
case *packet.SetActorData:
	if pk.EntityRuntimeID != conn.GameData().EntityRuntimeID {
		break
	}
	height, ok := pk.EntityMetadata[protocol.EntityDataKeyHeight]
	if !ok {
		break
	}
	fmt.Printf("Height: %v\n", height)
}
...
```

Dasciam:
^

PR #1007, opened 2025-01-31
Implement snow layer

AkmalFairuz:


Dasciam:
It seems that snow logging feature is missing

PR #1009, opened 2025-02-09
player.Handler improvements

FDUTCH:
This will simplify many things


Sandertv:
Would be very helpful if you could provide actual rationale for PRs. In what ways would you use this?

FDUTCH:
> Would be very helpful if you could provide actual rationale for PRs. In what ways would you use this?
I need to check every tick whether the player is in the certain blocks, I need to update the bossbar to the player in order to show how many time remained until the end of the battle, and there are many more use cases

FDUTCH:
I guess it never going to be merged

PR #1010, opened 2025-02-10
allow player.Handler to modify chat format

Blackjack200:


server/player/player.go
```diff
@@ -315,12 +315,15 @@ func (p *Player) RemoveBossBar() {
...
+		return fmt.Sprintf("<%v> %v", username, message)
+	}
 	message := format(msg)
 	ctx := event.C(p)
-	if p.Handler().HandleChat(ctx, &message); ctx.Cancelled() {
+	if p.Handler().HandleChat(ctx, &formatter, &message); ctx.Cancelled() {
 		return
 	}
-	_, _ = fmt.Fprintf(chat.Global, "<%v> %v\n", p.Name(), message)
+	_, _ = fmt.Fprintf(chat.Global, formatter(p.Name(), message)+"\n")
```
TwistedAsylumMC, line 326: If formatter is set to nil, this would cause a panic

TwistedAsylumMC (changes requested):
Just one issue I can see. As well as this though, is it *really* any easier than just cancelling the context and printing to the global chat yourself?

PR #1015, opened 2025-02-15
session/player.go: don't send actual walk speed via UpdateAbilities

Dasciam:
This fixes FoV and speed issues.


Sandertv (approved):

PR #1017, opened 2025-02-15
fix effects #1016

### LINKED ISSUE #1016: after re-entering the server the effects were not sent

FDUTCH:
fix #1016

### DIFF
--- a/server/entity/effect.go
+++ b/server/entity/effect.go
@@ -36,9 +36,7 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have negative duration: %v", dur))
 	}
 
-	if m.initialEffects != nil {
-		m.flushInitialEffects(entity)
-	}
+	m.flushInitialEffects(entity)
 
 	t, ok := e.Type().(effect.LastingType)
 	if !ok {
@@ -110,10 +108,7 @@ func (m *EffectManager) Effects() []effect.Effect {
 func (m *EffectManager) Tick(entity Living, tx *world.Tx) {
 	update := false
 
-	for _, eff := range m.initialEffects {
-		m.Add(eff, entity)
-	}
-	m.initialEffects = nil
+	m.flushInitialEffects(entity)
 
 	for i, eff := range m.effects {
 		if m.expired(eff) {
--- a/server/entity/effect.go
+++ b/server/entity/effect.go
@@ -36,7 +36,9 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have negative duration: %v", dur))
 	}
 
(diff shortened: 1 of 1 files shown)


server/entity/effect.go
```diff
@@ -59,6 +62,20 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 	return e
 }
 
+// checkInitials checks if added effect intersects with initial effects.
```
DaPigGuy, line 65: I think instead of just applying the initial effect that matches, just apply all of them

```
fn flushInitialEffects:
    initialEffects = m.initialEffects
    m.initialEffects = nil
    for every initialEffects:
        m.Add(initialEffect)
  ```

That way, we can get rid of the duplicated logic for which effect will take precedence. 
Please also use in Remove() and Tick()

File: server/entity/effect.go [hunk 2]
```diff
@@ -78,7 +95,7 @@ func (m *EffectManager) Effect(e effect.Type) (effect.Effect, bool) {
 // Effects returns a list of all effects currently present in the effect manager. This will never include
 // effects that have expired.
 func (m *EffectManager) Effects() []effect.Effect {
```
DaPigGuy, line 97: Please also fix Effect()

DaPigGuy (changes requested):

server/entity/effect.go
FDUTCH: don't really understand that: "Please also use in Remove() and Tick()"
(and the whole comment seems to be AI generated)

server/entity/effect.go
```diff
@@ -71,14 +89,20 @@ func (m *EffectManager) Remove(e effect.Type, entity Living) {
 // Effect returns the effect instance and true if the entity has the effect. If not found, it will return an empty
 // effect instance and false.
 func (m *EffectManager) Effect(e effect.Type) (effect.Effect, bool) {
+	for _, eff := range m.initialEffects {
```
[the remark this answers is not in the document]
FDUTCH: No need cause if m.initialEffects != nil, m.effects is empty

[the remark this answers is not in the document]
DaPigGuy: Right, Add flushes initial effects now

server/entity/effect.go
```diff
@@ -59,8 +62,21 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
...
+	initialEffects := m.initialEffects
+	m.initialEffects = nil
+	for _, e := range initialEffects {
+		m.Add(e, entity)
+	}
+}
+
 // Remove removes any Effect present in the EffectManager with the type of the effect passed.
 func (m *EffectManager) Remove(e effect.Type, entity Living) {
+	if m.initialEffects != nil {
```
DaPigGuy, line 76: Here too

server/entity/effect.go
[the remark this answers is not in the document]
FDUTCH: I can't do that, if I will than I need to pass entity in the function

PR #1018, opened 2025-02-16
Implement world explosion handler

xNatsuri:
This closes #991 

Co-authored-by: @Flonja

### DIFF
--- a/server/block/explosion.go
+++ b/server/block/explosion.go
@@ -3,6 +3,7 @@ package block
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/block/cube/trace"
+	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/item"
 	"github.com/df-mc/dragonfly/server/world"
 	"github.com/df-mc/dragonfly/server/world/particle"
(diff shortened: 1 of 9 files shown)


server/block/explosion.go
```diff
@@ -97,16 +98,9 @@ func (c ExplosionConfig) Explode(tx *world.Tx, explosionPos mgl64.Vec3) {
...
-		pos := e.Position()
-		dist := pos.Sub(explosionPos).Len()
-		if dist > d || dist == 0 {
-			continue
-		}
-		if explodable, ok := e.(ExplodableEntity); ok {
-			impact := (1 - dist/d) * exposure(tx, explosionPos, e)
-			explodable.Explode(explosionPos, impact, c)
-		}
+		affectedEntities = append(affectedEntities, e)
```
DaPigGuy, line 103: Keep the `dist > d || dist == 0` check here

xNatsuri: made the change

server/block/explosion.go
```diff
@@ -132,6 +132,14 @@ func (c ExplosionConfig) Explode(tx *world.Tx, explosionPos mgl64.Vec3) {
 			}
 		}
 	}
+
+	ctx := event.C(tx)
+	spawnFire := c.SpawnFire
+	itemDropChance := c.ItemDropChance
+	if tx.World().Handler().HandleExplosion(ctx, &affectedEntities, &affectedBlocks, &itemDropChance, &spawnFire); ctx.Cancelled() {
```
DaPigGuy, line 139: The explosion position would also be nice to have. For example, you might want to contain an explosion's effects within a plot which would require you to be able to identify the plot the explosion started from.

PR #1020, opened 2025-02-24
Added infinite duration effects

Superomarking:
Using `/effect Player effect infinite` in BDS gives the player an infinite duration of that effect, reference the picture below:

woah !!

Confirmed that it sends a -1 duration. (from BDS)
`&packet.MobEffect{EntityRuntimeID:0x1, Operation:0x2, EffectType:1, Amplifier:10, Particles:false, Duration:-1, Tick:0x0}`
Tested using `p.AddEffect(effect.New(effect.Speed, 1, -1))`

Let me know if I missed something, or did something wrong!

### DIFF
--- a/server/entity/effect.go
+++ b/server/entity/effect.go
@@ -49,7 +49,8 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 		t.Start(entity, lvl)
 		return e
 	}
-	if existing.Level() > lvl || (existing.Level() == lvl && existing.Duration() > dur) {
+	// Infinite duration has priority over other durations
+	if existing.Level() > lvl || (existing.Level() == lvl && (existing.Duration() > dur || existing.Duration() == -1)) {
 		return existing
 	}
 	m.effects[typ] = e
@@ -92,7 +93,7 @@ func (m *EffectManager) Tick(entity Living, tx *world.Tx) {
 	m.initialEffects = nil
 
 	for i, eff := range m.effects {
-		if m.expired(eff) {
+		if m.expired(eff) && eff.Duration() != -1 {
 			delete(m.effects, i)
 			eff.Type().(effect.LastingType).End(entity, eff.Level())
 			update = true
--- a/server/entity/effect.go
+++ b/server/entity/effect.go
@@ -32,8 +32,8 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 	if lvl <= 0 {
 		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have level of 0 or below: %v", lvl))
 	}
-	if dur < 0 {
-		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have negative duration: %v", dur))
+	if dur < -1 {
+		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have negative duration other than -1 (infinite): %v", dur))
 	}
 	t, ok := e.Type().(effect.LastingType)
 	if !ok {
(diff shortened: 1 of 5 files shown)


DaPigGuy (changes requested):
At minimum, [(EffectManager).expired](https://github.com/df-mc/dragonfly/blob/master/server/entity/effect.go#L113), [(Effect).TickDuration](https://github.com/df-mc/dragonfly/blob/master/server/entity/effect/effect.go#L108), and [(EffectManager).Add](https://github.com/df-mc/dragonfly/blob/master/server/entity/effect.go#L52) should be updated to handle infinite durations.

~~Have you checked how infinite duration is stored in a vanilla world?~~

Superomarking:
What do you mean how's it's stored in a vanilla world? I think it's just sent to the client, I mean it is an effect after all.

I've also sent a BDS snippet of that's what you mean

server/entity/effect.go
```diff
@@ -32,8 +32,8 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 	if lvl <= 0 {
 		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have level of 0 or below: %v", lvl))
 	}
-	if dur < 0 {
-		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have negative duration: %v", dur))
+	if dur < -1 {
```
DaPigGuy, line 35: Having a negative duration is a little odd, perhaps there should be an `infinite` field in `effect.Effect` alongside a `NewInfinite` method.

DaPigGuy (changes requested):
> What do you mean how it's stored in a vanilla world? I think it's just sent to the client, I mean it is an effect after all.

How is the duration for an effect stored for a player in a single-player world?

server/player/playerdb/effect.go
```diff
@@ -14,6 +14,7 @@ func effectsToData(effects []effect.Effect) []jsonEffect {
 			Duration: eff.Duration(),
 			Level:    eff.Level(),
 			Ambient:  eff.Ambient(),
+			Infinite: eff.Infinite(),
```
DaPigGuy, line 17: Update `dataToEffects` too

server/entity/effect.go
```diff
@@ -32,7 +32,7 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 	if lvl <= 0 {
 		panic(fmt.Sprintf("(*EffectManager).Add: effect cannot have level of 0 or below: %v", lvl))
 	}
-	if dur < 0 {
+	if dur < 0 && !e.Infinite() {
```
DaPigGuy, line 35:
```suggestion
	if dur < 0 {
```

server/player/playerdb/json.go
```diff
@@ -138,4 +138,5 @@ type jsonEffect struct {
 	Level    int
 	Duration time.Duration
 	Ambient  bool
+	Infinite bool
```
DaPigGuy, line 141: Not a problem with this PR, but particle hidden does not appear to be saved? Feel free to fix that

server/entity/effect.go
```diff
@@ -49,7 +49,8 @@ func (m *EffectManager) Add(e effect.Effect, entity Living) effect.Effect {
 		t.Start(entity, lvl)
 		return e
 	}
-	if existing.Level() > lvl || (existing.Level() == lvl && existing.Duration() > dur) {
+	// Infinite duration has priority over other durations
+	if existing.Level() > lvl || (existing.Level() == lvl && (existing.Duration() > dur || e.Infinite())) {
```
DaPigGuy, line 53:
```suggestion
	if existing.Level() > lvl || (existing.Level() == lvl && ((existing.Duration() > dur && !e.Infinite()) || existing.Infinite())) {
```

PR #1025, opened 2025-03-05
Improvement of commandline processing

FDUTCH:
with this change you can put as much spaces as you want into the command
something like this: `/tell "Cool  guy"       hello!  ` will not result an error any more

### DIFF
--- a/server/cmd/command.go
+++ b/server/cmd/command.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/csv"
 	"fmt"
 	"github.com/df-mc/dragonfly/server/world"
 	"go/ast"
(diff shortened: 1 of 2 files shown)


Sandertv:
This is not valid in vanilla and I don't really see the need for this to be valid in dragonfly. Any motivation for this change?

Sandertv:
Looks like they've changed this. It is actually valid now. Point taken.

server/cmd/command.go
```diff
@@ -345,3 +335,41 @@ func exportedFields(command reflect.Value) []reflect.StructField {
 	}
 	return fields
 }
+
+// SplitCommandArgs splits str to args for command execution.
+func SplitCommandArgs(str string) []string {
```
Sandertv, line 340: Function should not be exported.

server/cmd/command.go
```diff
@@ -345,3 +335,41 @@ func exportedFields(command reflect.Value) []reflect.StructField {
...
+			}
+			// end of the arg.
+			previousSpace = true
+			args = append(args, arg.String())
+			arg.Reset()
+		default:
+			previousSpace = false
+			arg.WriteRune(char)
+		}
+	}
```
Sandertv, line 370: I'm not a huge fan of re-inventing an entire parser for this. My suggestion would be this: 
- Keep using the csv reader, merge the first value in the slice returned with the second value if `first value == "/"`
- Remove all empty strings from the slice returned, except if the value is last in the slice.

Sandertv (changes requested):

server/cmd/command.go
Sandertv: I realise that this is used in player.go, but we should probably adapt that in one way or another so that this does not require being exported.

FDUTCH:
still it's not fully vanilla behavior cause vanilla allows to write commands like so `/  tell someone something`, but it is better than current system

FDUTCH:
I'm gonna open a new pr later, this one is garbage

PR #1029, opened 2025-03-09
advance tick for non overworld worlds

FDUTCH:
This PR fixes scheduled block updates and other stuff related to the world's tick progression in non overworld dimensions 

before this change:

https://github.com/user-attachments/assets/75fc884c-b9f0-49c0-8698-bd64a08a2373

after this change:

https://github.com/user-attachments/assets/0a4b3549-eec3-42b4-8a3b-5bf95b447f70


Sandertv:
This will advance the tick up to three times for every dimension in the same world folder. This code is explicitly to prevent that.

DaPigGuy:
What exactly is this PR trying to fix?

FDUTCH:
> What exactly is this PR trying to fix?

scheduled block updates in nether & end dimensions

FDUTCH:
bump

HashimTheArab (dismissed):
The per-dimension tick is now exposed consistently through Tx.CurrentTick, with regression coverage for worlds sharing Settings. Full CI passes and the remaining documentation thread has been addressed.

FDUTCH:
bump

HashimTheArab (approved):

PR #1041, opened 2025-04-03
handler_player_auth_input: Add safety check for excessive player move…

AkmalFairuz:
…ment

Source: https://github.com/pmmp/PocketMine-MP/blob/stable/src/player/Player.php#L1366


CJMustard1452:
This change does not appear to be parallel with vanilla minecraft.

Sandertv:
Give yourself strong speed or jump boost and you won't be able to move with this change. Not great.

Sandertv:
> cause lots of undesirable behaviour, like freezes, lag spikes and memory exhaustion due to sync chunk loading and collision checks across large distances.

Chunk loading does not happen based on movement, and collision checks are only performed when the squared distance < 4. These changes have no effect on any of those things.

PR #1043, opened 2025-04-06
fix shoutdown message race condition

FDUTCH:


Sandertv:
Race condition how exactly?

Sandertv:
Also please be more descriptive in future PRs. It is not clear to anyone reading this what this aims to solve - or what the problem even is in the first place.

Sandertv:
That is not a race condition and this change technically also probably does not solve it. That is probably because logs are buffered. If anything, see if you can find a way to flush logs to properly fix that.

Sandertv (approved):
Please, in the future, be descriptive with what exactly the issue is, how it arises, what kind of debugging you've done, etc., so that we actually understand the issue and how your PR fixes it.
Thanks for the PR otherwise!

PR #1046, opened 2025-04-07
fix: spectators is visible to others

Blackjack200:
the removed SetFlag remove the invisible flag for spectators.


TwistedAsylumMC:
How does this fix anything? Surely we need to set them invisible?

Dasciam:
rather the problem is that players shouldn't have an view of the players in the spectator mode

view:
spectator <-> spectator (they can see each other)
creative <- spectator (we should hide spectator from non-spectator mode players)

PR #1051, opened 2025-04-14
fix player can attack a dead entity & player can attack entity when dead

Blackjack200:


server/player/player.go
```diff
@@ -1644,9 +1644,15 @@ func (p *Player) UseItemOnEntity(e world.Entity) bool {
...
 func (p *Player) AttackEntity(e world.Entity) bool {
-	if !p.canReach(e.Position()) {
+	if !p.canReach(e.Position()) || p.Dead() {
+		return false
+	}
+
+	living, ok := e.(entity.Living)
+	if !ok || living.Dead() {
 		return false
 	}
```
TwistedAsylumMC: Moving the living check here means the `HandleAttackEntity` event will not be called. Can you not add the dead check to where the if statement was originally? Or you can keep both checks & change this one here to be `ok && living.Dead()`

Blackjack200: changed to `ok && living.Dead()`

server/player/player.go
```diff
@@ -1647,6 +1647,12 @@ func (p *Player) AttackEntity(e world.Entity) bool {
 	if !p.canReach(e.Position()) {
 		return false
 	}
+
+	living, ok := e.(entity.Living)
+	if ok && living.Dead() {
```
Dasciam: !ok || living.Dead()?

server/player/player.go
```diff
@@ -1661,10 +1667,6 @@ func (p *Player) AttackEntity(e world.Entity) bool {
 	p.SwingArm()
 
 	i, _ := p.HeldItems()
-	living, ok := e.(entity.Living)
-	if !ok {
-		return false
-	}
```
TwistedAsylumMC: Bring back this if statement to keep the old functionality so non-living entities don't get past this point. Don't need the first line though since we already have that check. Might be better to rename the ok to something like isLiving though just so it's more clear

Blackjack200: the original 
```go
living, ok := e.(entity.Living)
if !ok || living.Dead() {
	return false
}
```
does the same thing, when e is not living , it is `true || ?`, when e is living, it is `false || living.Dead()`.

TwistedAsylumMC: The way you've now got this code means the attack event will not be called for non-living entities which is not what the behaviour was before. That's why I'm saying to change it

Blackjack200: > The way you've now got this code means the attack event will not be called for non-living entities which is not what the behaviour was before. That's why I'm saying to change it

got it

PR #1055, opened 2025-04-20
new basic entity actions: link/unlink

k4ties:


xan6x:
penis

TwistedAsylumMC:
How do you plan on this being used, there's no actual exposed API unless you're accessing the session which isn't recommended

PR #1061, opened 2025-05-09
dragonfly: Updated for 1.21.80.

didntpot:


NopeNotDark (approved):
LGTM, Thanks for the PR!

server/world/biome_definition.go
```diff
@@ -53,42 +39,58 @@ func init() {
...
-func BiomeDefinitions() ([]protocol.BiomeDefinition, []string) {
-	return cachedBiomeDefinitions, cachedBiomeStringList
+// int32FromRGBA converts a color.RGBA into an int32. These int32s are present in things such as signs and dyed leather armour.
+func int32FromRGBA(x color.RGBA) int32 {
+	if x.R == 0 && x.G == 0 && x.B == 0 {
+		// Default to black colour. The default (0x000000) is a transparent colour. Text with this colour will not show
+		// up on the sign.
+		return int32(-0x1000000)
+	}
+	return int32(binary.BigEndian.Uint32([]byte{x.A, x.R, x.G, x.B}))
```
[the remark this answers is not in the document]
TwistedAsylumMC: If you do remove that check, the method can probably be removed and just inlined/put into a variable where used

server/world/biome_definition.go
```diff
@@ -1,43 +1,29 @@
...
+	// Spores returns the blue and red spores of the biome.
+	Spores() (blueSpores float64, redSpores float64)
+}
 
-	var rawBiomes []biomeNBT
-	if err := nbt.NewDecoder(bytes.NewReader(biomeDefinitionData)).Decode(&rawBiomes); err != nil {
-		panic(err)
-	}
+// MaxVanillaBiomeID ...
+const MaxVanillaBiomeID = 193
```
TwistedAsylumMC, line 23: I do not like how this is a hard coded number, is there any way we can automatically calculate this?

didntpot: not that i'm aware of, any suggestions?

TwistedAsylumMC: Best I can think of right now is to have a world.FinaliseBiomeRegistry() method, similar to blocks, which sets an unexpected field based on the max ID it has up until that point. Make sure this method can only be called once like the block one.

Where does the 193 actually come from? Since there's only ~90 biomes, is that just the highest ID or something?

didntpot: alright, i'll take a look at that. and 193 is the highest id i believe for the pale garden

TwistedAsylumMC: Ah yeah the max should be fine then. Maybe we could also automate biome ID's for custom biomes? I'm pretty sure they're only ever used at runtime and not stored on disk

didntpot: i've pushed smth, take a look whenever you're able to. i'm off for the night

PR #1062, opened 2025-05-11
implement item.OffHand to offhand items

FDUTCH:
It would be useful to know if you can put item in the offhand

Ps:
 maybe I miss some items

### DIFF
--- a/server/item/inventory/inventory.go
+++ b/server/item/inventory/inventory.go
@@ -21,14 +21,14 @@ type Inventory struct {
 	slots []item.Stack
 
 	f         SlotFunc
-	validator SlotValidator
+	validator SlotValidatorFunc
 }
 
 // SlotFunc is a function called for each item changed in an Inventory.
 type SlotFunc func(slot int, before, after item.Stack)
 
-// SlotValidator is a function that limits changes in the Inventory slot.
-type SlotValidator func(s item.Stack, slot int) bool
+// SlotValidatorFunc is a function that limits changes in the Inventory slot.
+type SlotValidatorFunc func(s item.Stack, slot int) bool
 
 // ErrSlotOutOfRange is returned by any methods on inventory when a slot is passed which is not within the
 // range of valid values for the inventory.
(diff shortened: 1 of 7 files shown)


cjmustard:
arrow, spectral arrow, tipped arrow, firework rocket, totem of undying, shield, map, nautilus shell

I think you have all of the existent ones added

Dasciam:
also you could add validation. if an item can't be placed in the off hand the action should be cancelled

FDUTCH:
> also you could add validation. if an item can't be placed in the off hand the action should be cancelled

I had no plans to validate slots because then I would have to do validation for armor as well

FDUTCH:
maybe I'll implement validation in the next pr

DaPigGuy:
> > also you could add validation. if an item can't be placed in the off hand the action should be cancelled
> 
> I had no plans to validate slots because then I would have to do validation for armor as well

https://github.com/df-mc/dragonfly/blob/master/server/item/inventory/armour.go#L24

FDUTCH:
> > > also you could add validation. if an item can't be placed in the off hand the action should be cancelled
> > 
> > 
> > I had no plans to validate slots because then I would have to do validation for armor as well
> 
> https://github.com/df-mc/dragonfly/blob/master/server/item/inventory/armour.go#L24

right, I overlooked that

server/item/inventory/inventory.go
```diff
@@ -20,13 +20,16 @@ type Inventory struct {
...
-	canAdd func(s item.Stack, slot int) bool
+	f         SlotFunc
+	validator SlotValidator
 }
 
 // SlotFunc is a function called for each item changed in an Inventory.
 type SlotFunc func(slot int, before, after item.Stack)
 
+// SlotValidator is a function that limits changes in the Inventory slot.
+type SlotValidator func(s item.Stack, slot int) bool
```
TwistedAsylumMC: I'd rename this to SlotValidatorFunc

server/item/inventory/inventory.go
```diff
@@ -61,6 +64,13 @@ func (inv *Inventory) SlotFunc(f SlotFunc) {
 	inv.f = f
 }
 
+// SlotValidator changes the function that limits item placement in the inventory slot.
+func (inv *Inventory) SlotValidator(f SlotValidator) {
```
TwistedAsylumMC: Then also rename this method to SlotValidatorFunc

server/player/type.go
```diff
@@ -20,6 +20,14 @@ func (t ptype) Open(tx *world.Tx, handle *world.EntityHandle, data *world.Entity
 		playerData: pd,
 	}
 
+	pd.offHand.SlotValidator(func(s item.Stack, _ int) bool {
+		if s.Empty() {
+			return true
+		}
+		_, allowedInOffhand := s.Item().(item.OffHand)
+		return allowedInOffhand
```
TwistedAsylumMC: I'd also make sure the OffHand() method returns true on the item that is an item.OffHand

PR #1066, opened 2025-05-16
implement totem ignoring for damage source

FDUTCH:


TwistedAsylumMC:
Considering almost all sources *don't* ignore totems, would it make more sense to create an interface that can be implemented for this behaviour?

FDUTCH:
> Considering almost all sources _don't_ ignore totems, would it make more sense to create an interface that can be implemented for this behaviour?

Well then why did you have to make a separate method for the resistance effect or for the armor check? 
Originally I didn't do it this way because I didn't know how to name the interface and where to place it.

PR #1067, opened 2025-05-18
Various fixes for spectator gamemode

didntpot:


didntpot:
i would implement SurvivalSpectator, CreativeSpectator but BDS doesn't let you switch to those gamemodes

PR #1075, opened 2025-05-30
fix session.Conn hooking

FDUTCH:
since this [commit](https://github.com/df-mc/dragonfly/commit/405843c030a778f4cd2c2e49c4f556c070414f6e) it is no longer possible to handle client-bound packets by hooking session.Conn interface, this pr makes it possible again


FDUTCH:
closed because reflection is such bad thing

PR #1084, opened 2025-06-22
feat: Player colours, locator bar functionality.

Dasciam:
# Potential Performance Consideration

may need to remove dynamic colour changes. rapidly updating a player's colour could lead to increased traffic, as the player list entry is resent with each change

also added the toggling of locator bar

### DIFF
--- a/server/session/player.go
+++ b/server/session/player.go
@@ -3,6 +3,9 @@ package session
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/df-mc/dragonfly/server/player/debug"
+	"github.com/go-gl/mathgl/mgl32"
+	"golang.org/x/exp/rand"
 	"image/color"
 	"maps"
 	"math"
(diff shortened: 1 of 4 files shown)


Flonja:
Maybe use the hashed player name? Or store the (randomized) color in `playerData`?

server/session/player.go
```diff
@@ -1053,6 +1075,16 @@ func protocolToSkin(sk protocol.Skin) (s skin.Skin, err error) {
 	return
 }
 
+// randomColour returns random colour.
+func randomColour() color.RGBA {
+	return color.RGBA{
+		R: byte(rand.Int31n(255)),
+		G: byte(rand.Int31n(255)),
+		B: byte(rand.Int31n(255)),
```
DaPigGuy: 256?

Dasciam: no, the number from 0 to 255 (0xFF), isn't it?

DaPigGuy: https://pkg.go.dev/math/rand#Int31n

[0, n) not inclusive

PR #1091, opened 2025-06-26
improve commandline processing

FDUTCH:
with this change you can put as much spaces as you want into the command
something like this: ```/tell "Cool  guy"       hello! ``` will not result an error any more


server/cmd/argument.go
```diff
@@ -185,6 +198,10 @@ func (p parser) float(line *Line, v reflect.Value) error {
 
 // string ...
 func (p parser) string(line *Line, v reflect.Value) error {
+	if p.fields == 1 {
+		return p.restAsString(line, v)
```
DaPigGuy, line 202: Isn't this what the Varargs parameter type is for?

server/cmd/argument.go
```diff
@@ -46,7 +47,12 @@ func (line *Line) Next() (string, bool) {
 	if !ok {
 		return "", false
 	}
-	return v[0], true
+	val := v[0]
+	if val == "" {
+		line.RemoveNext()
+		return line.Next()
+	}
+	return val, true
```
DaPigGuy, line 55: Ignoring the empty arg should probably be done in `NextN` for more consistent behavior? It's a bit odd that one would skip over empty args and the other wouldn't

DaPigGuy: Would go a step further and do the same for `RemoveN`

server/player/player.go
```diff
@@ -332,13 +333,24 @@ func (p *Player) ExecuteCommand(commandLine string) {
...
-	name, ok := strings.CutPrefix(args[0], "/")
+	commandLine, ok := strings.CutPrefix(commandLine, "/")
 	if !ok {
 		return
 	}
 
+	args := strings.Split(commandLine, " ")
+
+	var (
+		name  string
```
DaPigGuy, line 344: Would be good to leave a comment here explaining this is to allow something like `/    command` to work

PR #1100, opened 2025-07-07
Handle block entities with nil nbt

Flonja:
Check if the data in a block entity is nil, so that there won't be an assign to nil map panic

Schematic: https://mcbuild.org/schematics/18554:lighthouse
Test code:
```go
file, err := os.Open("test.schematic")
if err != nil {
	panic(err)
}
s, err := schematic.FromReader(file)
if err != nil {
	panic(err)
}

w := srv.World()
<-w.Exec(func(tx *world.Tx) {
	tx.BuildStructure(cube.Pos{}, s)
})
```


server/world/mcdb/db.go
HashimTheArab, line 456: if buf.Len() == 0 {
	batch.Delete(k.Sum(keyBlockEntities))
	return
}
Can you add this before batch.Put to avoid writing empty values if all block entities were skipped

HashimTheArab:
Preferring this over #1275, this PR is due to bad world data from other server softwares but Dragonfly should not panic
image

Flonja:
Should it not fix broken block entities by correcting the lack of NBT? In my opinion my pr (this one), should've corrected it. A block entity isn't a block entity without nbt and dragonfly should correct this

PR #1101, opened 2025-07-08
Patch vanilla fly

ethaniccc:
Currently, Dragonfly uses the RequestAbility packet to determine if the player is requesting to fly or not. However, the client seems to handle this via. PlayerAuthInput packet instead. This PR allows Dragonfly to have the client sync abilities to prevent them from flying.


server/session/handler_player_auth_input.go
```diff
@@ -135,6 +136,17 @@ func (h PlayerAuthInputHandler) handleInputFlags(flags protocol.Bitset, s *Sessi
 		defer s.swingingArm.Store(false)
 		c.PunchAir()
 	}
+	if flags.Load(packet.InputFlagStartFlying) {
+		if !c.GameMode().AllowsFlying() {
+			s.conf.Log.Debug("process packet: RequestAbility: flying flag enabled while unable to fly")
```
didntpot: you forgot to change this

didntpot (changes requested):

TwistedAsylumMC:
Do we need to remove any of the existing logic or does this work on top of the old stuff?

ethaniccc:
No existing logic has to be changed, this PR can work with the RequestAbility handler as well

PR #1111, opened 2025-08-16
improve chunk NetworkDecode

FDUTCH:
this pr allows to read chunk data left after decoding


server/world/chunk/decode.go
```diff
@@ -3,20 +3,26 @@ package chunk
...
 	"bytes"
 	"fmt"
+
 	"github.com/df-mc/dragonfly/server/block/cube"
 )
 
 // StateToRuntimeID must hold a function to convert a name and its state properties to a runtime ID.
 var StateToRuntimeID func(name string, properties map[string]any) (runtimeID uint32, found bool)
 
-// NetworkDecode decodes the network serialised data passed into a Chunk if successful. If not, the chunk
```
TwistedAsylumMC, line 12: I'd bring back the same docs for this function, so it's shown in IDE's and on the go docs page. It's a short comment anyways so no harm having it twice. Maybe add that it creates a new buffer and calls NetworkDecodeBuffer or something like that

PR #1113, opened 2025-09-02
Prevent arbitrary block-resend requests that can exhaust server RAM

AkmalFairuz:


server/player/player.go
```diff
@@ -3134,19 +3134,29 @@ func (p *Player) viewers() []world.Viewer {
...
 }
 
-// resendBlock resends the block at a cube.Pos in the world.World passed.
-func (p *Player) resendBlock(pos cube.Pos) {
+// resendNearbyBlock resends a block at cube.Pos if it is within the player's render distance.
+func (p *Player) resendNearbyBlock(pos cube.Pos) {
+	if p.session() == session.Nop {
+		return
+	}
+	if p.Position().Sub(pos.Vec3()).Len() > float64(p.session().ChunkRadius()*16) {
```
DaPigGuy: Not sure this is quite right for chunks diagonal from the current chunk

didntpot (approved):

PR #1117, opened 2025-09-18
item/stack.go: add option to apply incompatible enchantments

FDUTCH:
I think it is better to have this ability


server/item/stack.go
```diff
@@ -275,6 +275,13 @@ func (s Stack) WithEnchantments(enchants ...Enchantment) Stack {
 	return s
 }
 
+// WithIncompatibleEnchantment returns the current stack with the passed enchantment (it ignores item compatability with enchantment).
+func (s Stack) WithIncompatibleEnchantment(enchant Enchantment) Stack {
```
Dasciam: I'd refactor it like this:
```go
func (s Stack) WithIncompatibleEnchantments(enchants ...Enchantment) Stack
```

should accept variadic arguments like `WithEnchantments` do

Dasciam (changes requested):

server/item/stack.go
FDUTCH: maybe you are right

server/item/creative/creative.go
```diff
@@ -164,7 +164,7 @@ func itemStackFromEntry(data creativeItemEntry) (item.Stack, bool) {
 					invalid = true
 					break
 				}
-				st = st.WithEnchantments(item.NewEnchantment(t, int(nbtconv.Int16(v, "lvl"))))
+				st = st.WithIncompatibleEnchantments(item.NewEnchantment(t, int(nbtconv.Int16(v, "lvl"))))
```
TwistedAsylumMC: When do the vanilla creative items ever have incompatible enchantments?

server/item/stack.go
```diff
@@ -275,6 +275,15 @@ func (s Stack) WithEnchantments(enchants ...Enchantment) Stack {
 	return s
 }
 
+// WithIncompatibleEnchantments returns the current stack with the passed enchantments (it ignores item compatability with enchantments).
```
TwistedAsylumMC: Wrap the comment like everything else, I'd also change it slightly
```go
// WithIncompatibleEnchantments returns the current stack with the passed enchantments, allowing incompatible
// enchantments to be applied.
```

TwistedAsylumMC (dismissed):
> I think it is better to have this ability
Do you have an actual use case for this feature? Could you give an example?

server/item/creative/creative.go
FDUTCH: yeah, you are right

Dasciam (approved):

FDUTCH:
> > I think it is better to have this ability
> > Do you have an actual use case for this feature? Could you give an example?

I'm not going to argue, I know it's pretty pointless, but I just need it

Dasciam:
we do not have any need to restrict the developer from making items with incompatible enchantments anyway

TwistedAsylumMC:
I think that's fair enough, but I do feel like this method deserves a better name. Currently it sounds like every enchantment you provided *has* to be incompatible. What do you think of something like `WithIncompatibleEnchantmentsAllowed()` or something similar?

FDUTCH:
> I think that's fair enough, but I do feel like this method deserves a better name. Currently it sounds like every enchantment you provided _has_ to be incompatible. What do you think of something like `WithIncompatibleEnchantmentsAllowed()` or something similar?

maybe, I'm not too good at naming

FDUTCH:
> or have the `WithEnchantments` method be `WithEnchantments(ignoreIncompatibles bool, enchantments Enchantment...)`. You would break the original usage for little gain tho...

I think it's better to create a separate method for this

FDUTCH:
I guess it is ready to be merged

Dasciam (approved):

FDUTCH:
maybe merge? :thinking:

FDUTCH:
it's been a while for such a small pr...

HashimTheArab (dismissed):

server/item/stack.go
```diff
@@ -275,6 +275,16 @@ func (s Stack) WithEnchantments(enchants ...Enchantment) Stack {
 	return s
 }
 
+// ForceEnchantments returns the current stack with the passed enchantments, allowing incompatible
+// enchantments to be applied.
+func (s Stack) ForceEnchantments(enchants ...Enchantment) Stack {
```
RestartFU: to stay consistent maybe?

```suggestion
func (s Stack) WithForcedEnchantments(enchants ...Enchantment) Stack {
```

server/internal/nbtconv/read.go
```diff
@@ -225,7 +226,7 @@ func readEnchantments(m map[string]any, s *item.Stack) {
 	}
 	for _, ench := range enchantments {
 		if t, ok := item.EnchantmentByID(int(Int16(ench, "id"))); ok {
-			*s = s.WithEnchantments(item.NewEnchantment(t, int(Int16(ench, "lvl"))))
+			*s = s.ForceEnchantments(item.NewEnchantment(t, int(Int16(ench, "lvl"))))
```
HashimTheArab:
```suggestion
			*s = s.WithForcedEnchantments(item.NewEnchantment(t, int(Int16(ench, "lvl"))))
```

server/item/stack.go
```diff
@@ -275,6 +275,16 @@ func (s Stack) WithEnchantments(enchants ...Enchantment) Stack {
 	return s
 }
 
+// ForceEnchantments returns the current stack with the passed enchantments, allowing incompatible
+// enchantments to be applied.
```
HashimTheArab:
```suggestion
// WithForcedEnchantments returns the current stack with the passed enchantments applied,
// bypassing compatibility checks that would normally prevent incompatible enchantments
// from being applied together.
```

HashimTheArab (dismissed):

HashimTheArab (approved):

RestartFU (approved):

PR #1118, opened 2025-09-19
server: Fix persona premium skins by setting correct skin full id.

didntpot:


server/server.go
```diff
@@ -606,6 +606,8 @@ func (srv *Server) parseSkin(data login.ClientData) skin.Skin {
 	playerSkin.Model, _ = base64.StdEncoding.DecodeString(data.SkinGeometry)
 	playerSkin.ModelConfig, _ = skin.DecodeModelConfig(skinResourcePatch)
 	playerSkin.PlayFabID = data.PlayFabID
+	// The clients full skin identifier is provided in SkinID during login.
```
TwistedAsylumMC: I don't think this comment is necessary here but you could move it into the Skin struct and also make a comment about why it's needed

PR #1119, opened 2025-09-19
Entity rideable feature

Studgi:
This PR is an updated version of #330 but I made several changes according to the feedback given in that PR.

- If a rideable has 2 or more riders; if any dismounts occurred, the API should not be responsibly for reordering the seat index of the remaining rider- it should be changed outside of dragonfly.
- It's much simpler and voids confusion between <Rideable>.AddRider() and <Rider>.MountEntity()
- It's still very flexible since you can change the seat position during runtime via the HandleMount handler (seatPos *Vec3)
- You are able to control who the driver is without using their seat position to do so

If you want to suggest changes, please let me know.


server/entity/riding.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/go-gl/mathgl/mgl32"
+)
+
+// Rider is an interface for entities that can ride other entities.
+type Rider interface {
+	world.Entity
+	// RidingEntity returns the entity the player is currently riding.
+	RidingEntity() Rideable
+	// SeatPosition returns the position of where the player is sitting.
```
TwistedAsylumMC: Rephrase the word "player" here

server/entity/riding.go
```diff
@@ -0,0 +1,26 @@
...
+	MountEntity(e Rideable, position mgl32.Vec3, driver bool)
+	// DismountEntity dismounts the rider from the entity they are currently riding.
+	DismountEntity()
+}
+
+// Rideable is an interface for entities that can be ridden.
+type Rideable interface {
+	world.Entity
+	Driver() Rider
+	Move(vector mgl32.Vec2, yaw, pitch float32)
```
TwistedAsylumMC: These names could probably be more clear, i.e. ControllingRider() and MoveInput() or something along those lines, just so it's obvious these methods are related to entity riding when not looking at this interface

server/player/handler.go
```diff
@@ -138,6 +140,12 @@ type Handler interface {
 	// ctx.Cancel() may be called to prevent the player from dropping the entity.Item passed on the ground.
 	// e.Item() may be called to obtain the item stack dropped.
 	HandleItemDrop(ctx *Context, s item.Stack)
+	// HandleMount handles when a player mounts an entity. ctx.Cancel() may be called to cancel the player mounting
+	// an entity.
+	HandleMount(ctx *Context, r entity.Rideable, seatPos *mgl32.Vec3, driver *bool)
+	// HandleDismount handles when a player mounts an entity. ctx.Cancel() may be called to force the player
+	// to re-mount the entity.
+	HandleDismount(ctx *Context, r entity.Rideable)
```
TwistedAsylumMC: I'd rename these to HandleMountEntity() and HandleDismountEntity() just to be consistent with other methods (i.e. HandleAttackEntity())

server/entity/riding.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/go-gl/mathgl/mgl32"
+)
+
+// Rider is an interface for entities that can ride other entities.
+type Rider interface {
+	world.Entity
+	// RidingEntity returns the entity the player is currently riding.
+	RidingEntity() Rideable
+	// SeatPosition returns the position of where the player is sitting.
+	SeatPosition() mgl32.Vec3
```
TwistedAsylumMC: This should return a mgl64.Vec3, we try to always use float64 in the API and convert to a float32 for the protocol (see [vec64to32](https://github.com/df-mc/dragonfly/blob/master/server/server.go#L635))

server/player/player.go
```diff
@@ -54,6 +55,9 @@ type playerData struct {
 	armour                       *inventory.Armour
 	heldSlot                     *uint32
 
+	rideable     entity.Rideable
+	seatPosition mgl32.Vec3
```
TwistedAsylumMC: This should also be a mgl64.Vec3

server/player/player.go
```diff
@@ -2554,6 +2558,56 @@ func (p *Player) SetMaxAirSupply(duration time.Duration) {
 	p.updateState()
 }
 
+// RidingEntity returns the entity the player is currently riding.
+func (p *Player) RidingEntity() entity.Rideable {
+	return p.rideable
+}
+
+// SeatPosition returns the position of where the player is sitting.
+func (p *Player) SeatPosition() mgl32.Vec3 {
```
TwistedAsylumMC: This should also return a mgl64.Vec3

server/player/player.go
```diff
@@ -2554,6 +2558,56 @@ func (p *Player) SetMaxAirSupply(duration time.Duration) {
...
+}
+
+// MountEntity mounts the player to an entity if the entity is rideable and if there is a seat available.
+func (p *Player) MountEntity(r entity.Rideable, seatPos mgl32.Vec3, driver bool) {
+	ctx := event.C(p)
+	if p.h.HandleMount(ctx, r, &seatPos, &driver); ctx.Cancelled() {
+		return
+	}
+
+	if rd := p.RidingEntity(); rd != nil {
```
TwistedAsylumMC: How come you're not just using p.rideable here?

server/player/player.go
```diff
@@ -2554,6 +2558,56 @@ func (p *Player) SetMaxAirSupply(duration time.Duration) {
...
+	for _, v := range p.viewers() {
+		v.ViewEntityMount(p, r, driver)
+	}
+}
+
+// DismountEntity dismounts the player from an entity.
+func (p *Player) DismountEntity() {
+	ctx := event.C(p)
+	r := p.RidingEntity()
+	if r == nil {
```
TwistedAsylumMC: How come you're not just using p.rideable here?

server/world/viewer.go
```diff
@@ -51,6 +51,11 @@ type Viewer interface {
 	ViewEntityState(e Entity)
 	// ViewEntityAnimation starts viewing an animation performed by an Entity.
 	ViewEntityAnimation(e Entity, a EntityAnimation)
+	// ViewEntityMount views one entity mounting another. It is called when any entity mounts another or
+	// changes its role (passenger or driver).
+	ViewEntityMount(r Entity, rd Entity, driver bool)
+	// ViewEntityDismount views one entity dismounting another. It is called when any entity is dismounted.
+	ViewEntityDismount(r Entity, rd Entity)
```
TwistedAsylumMC: What's r and rd here? I feel like these variable could do with better names or docs describing what they are. It's quite hard to tell otherwise

TwistedAsylumMC (changes requested):
Thank you for picking up on the old PR, it's very appreciated. I've left a few comments in the code but I do have one other concern.

There's currently no way for the ridden entity to control how many riders it can have but also where the seats are, which I feel like it's something that should happen. From what I can tell, any player can go up to the same entity and start riding it, ending up in the same place as everyone else unless a player handler changes this? Rideable() should most likely have some sort of method that allows controlling this, providing the seat positions etc.

server/player/player.go
Studgi: RidingEntity() previously contained mutex logic which was no longer needed so I missed this

server/entity/riding.go
Studgi: Ohh okay

server/session/handler_inventory_transaction.go
```diff
@@ -158,6 +159,11 @@ func (h *InventoryTransactionHandler) handleUseItemOnEntityTransaction(data *pro
 	var valid bool
 	switch data.ActionType {
 	case protocol.UseItemOnEntityActionInteract:
+		if rideable, ok := e.(entity.Rideable); ok {
+			if nextSeatIndex, ok := rideable.NextFreeSeatIndex(); ok {
```
DaPigGuy: Could we pass the ClickedPosition so that it is possible to allow choosing a seat? e.g. the train in the Hive's hub

Studgi: Ofc bbg will do today

server/player/player.go
```diff
@@ -2554,6 +2557,78 @@ func (p *Player) SetMaxAirSupply(duration time.Duration) {
...
+	}
+}
+
+// SeatPosition returns the position of the seat the player is currently sitting on. If the player is not
+// currently riding an entity, the second return value is false.
+func (p *Player) SeatPosition() (mgl64.Vec3, bool) {
+	if p.rideable == nil || p.seatIndex == -1 || len(p.rideable.SeatPositions()) <= p.seatIndex {
+		return mgl64.Vec3{}, false
+	}
+	return p.rideable.SeatPositions()[p.seatIndex], true
```
TwistedAsylumMC: This could technically cause a panic if the size of SeatPositions() changes and the player was not ejected

server/player/player.go
```diff
@@ -54,6 +54,9 @@ type playerData struct {
 	armour                       *inventory.Armour
 	heldSlot                     *uint32
 
+	rideable  entity.Rideable
```
TwistedAsylumMC: I'd call this "riding" instead of "rideable", it currently doesn't sound like it's the entity the player is riding

TwistedAsylumMC (changes requested):
I'm happy with the direction this has gone, just a few minor comments. I assume this has been tested with the most recent changes? Would be nice to see some minimal code showing an example entity being created with seats etc.

Studgi:
> I'm happy with the direction this has gone, just a few minor comments. I assume this has been tested with the most recent changes? Would be nice to see some minimal code showing an example entity being created with seats etc.
Yep it's tested. Also where do I put the example code?

server/player/player.go
Studgi: Can we go with rideableEntity instead? I am concerned they might confuse "riding" with "rider"

server/player/player.go
Studgi: `len(p.rideable.SeatPositions()) <= p.seatIndex` would've become true though right? Meaning it'll return Vec3{}, and false

Also even though seatIndex will be considered invalid, it won't cause any panics with any logic that uses it

TwistedAsylumMC: Yes you're completely right, I am just blind

server/player/player.go
TwistedAsylumMC: Still not clear imo, I think you need to have "riding" in there somewhere. If you want you can also add a doc to explain but i don't think that's necessary with just "riding" or "ridingEntity". You could also do "riddenEntity" or something like that. The "able" just sounds like it's something that can be ridden, not actively being ridden

server/session/entity_metadata.go
```diff
@@ -70,7 +70,7 @@ func (s *Session) addSpecificMetadata(e any, m protocol.EntityMetadata) {
 		m.SetFlag(protocol.EntityDataKeyFlags, protocol.EntityDataFlagInvisible)
 	}
 	if b, ok := e.(baby); ok && b.IsBaby() {
```
didntpot: i think `Baby()` would look better here, also is this supposed to be part of this PR?

server/player/player.go
```diff
@@ -55,6 +55,9 @@ type playerData struct {
 	armour                       *inventory.Armour
 	heldSlot                     *uint32
 
+	riddenEntity entity.Rideable
```
AkmalFairuz: I think we should use `*world.EntityHandle` here, since the entity may become invalid, for example if it moves to another world.

AkmalFairuz (changes requested):

server/session/entity_metadata.go
Studgi: Nope. The query name for this property is q.is_baby, so IsBaby() suits this better
Also no it's not supposed to 💀 I didn't know pushing to my repo Studgi/dragonfly would add onto this PR..

server/player/player.go
Studgi: When you move to another world in vanilla (e.g. through nether portal), the link should break anyway
But yeah I do agree it should be *world.EntityHandle. I'll change it soon

server/session/entity_metadata.go
Studgi: Several other people explained why this was needed for consistency, so I updated the code. Thanks for pointing this out!

HashimTheArab:
updated this to latest df and fixed some issues

PR #1123, opened 2025-10-01
dragonfly: Updated for 1.21.111.

didntpot:


server/block/chain.go
```diff
@@ -41,12 +41,12 @@ func (c Chain) BreakInfo() BreakInfo {
 
 // EncodeItem ...
 func (Chain) EncodeItem() (name string, meta int16) {
```
TwistedAsylumMC: I'd rename this file to iron_chain and the struct to IronChain, since we are going to be adding CopperChain also

TwistedAsylumMC (changes requested):
Just one minor thing

PR #1135, opened 2025-11-18
parity vanilla: reduce durability every time player attacks living entity

Dasciam:


didntpot (approved):
LGTM!

Dasciam:
> I'm sorry, the issue this PR aims to fix isn't entirely clear to me? What behaviour does this fix?

In vanilla, when you attack an entity with something durable, it's durability reduces by 1 every time you hit with it.

This is described by Minecraft bedrock developers as intended behaviour, their explanation is: if shield is losing durability every time you hit to it, so your weapon should do it either. Same as with entities

HashimTheArab (approved):

PR #1140, opened 2025-12-06
fix consumables, chargeables items

FDUTCH:


server/player/player.go
```diff
@@ -1400,22 +1406,18 @@ func (p *Player) UseItem() {
...
-			if !usable.ReleaseCharge(p, p.tx, useCtx) {
-				// If the item was not charged yet, start charging.
-				p.usingSince, p.usingItem = time.Now(), true
-			}
-			p.handleUseContext(useCtx)
-			p.updateState()
+		if p.usingItem {
+			p.tryChargeItem(usable, false)
+			return
+		} else if time.Since(p.serverChargedItem) <= time.Millisecond*100 {
```
DaPigGuy, line 1522: Is there no way at all to distinguish the client wants to finish charging a crossbow vs. release it?

server/player/player.go
```diff
@@ -1400,22 +1406,18 @@ func (p *Player) UseItem() {
...
-		useCtx := p.useContext()
-		if !p.usingItem {
-			if !usable.ReleaseCharge(p, p.tx, useCtx) {
-				// If the item was not charged yet, start charging.
-				p.usingSince, p.usingItem = time.Now(), true
-			}
-			p.handleUseContext(useCtx)
-			p.updateState()
+		if p.usingItem {
+			p.tryChargeItem(usable, false)
```
DaPigGuy, line 1520: Is there even a reason to care about the client finishing the charge anymore if we are handling it on player tick? I believe players get ticked before the transactions executed by packet handlers are handled, so player tick would beat UseItem in most cases unless the client sends too early, in which case the charge finish would be rejected anyway.

But if the UseItem charge attempt does not get rejected, would there not be an issue of crossbow releases sometimes having an artificial delay of 100ms and sometimes not?

server/player/player.go
```diff
@@ -1296,6 +1299,9 @@ func (p *Player) SetHeldSlot(to int) error {
 		return nil
 	}
 	*p.heldSlot = uint32(to)
+	if p.usingItem {
+		defer p.updateState()
+	}
 	p.usingItem = false
```
DaPigGuy:
```suggestion
	if p.usingItem {
		p.usingItem = false
		p.updateState()
	}
```

server/player/player.go
```diff
@@ -1471,11 +1442,11 @@ func (p *Player) UseItem() {
 // ReleaseItem either aborts the using of the item or finished it, depending on the time that elapsed since
 // the item started being used.
 func (p *Player) ReleaseItem() {
-	if !p.usingItem || !p.canRelease() || !p.GameMode().AllowsInteraction() {
-		p.usingItem = false
+	defer p.updateState()
+	p.usingItem = false
```
DaPigGuy:
```suggestion
	p.usingItem = false
	p.updateState()
```

Good catch

server/player/player.go
FDUTCH: idk if we should care about client finishing the charge, but I think it would be better to remove this option

I found that vanilla has a slight delay

DaPigGuy: Ok we can see if any players end up complaining about that delay later then

FDUTCH:
bump

HashimTheArab:
```
Found one high-severity issue:

  - [P1] Unsolicited release packets can fire a fully charged bow — server/player/player.go:1444

    ReleaseItem no longer checks p.usingItem before releasing the held item:

    p.usingItem = false
    p.updateState()
    if !p.canRelease() || !p.GameMode().AllowsInteraction() {
        return
    }

    A client can send a release transaction while holding a bow without first starting item use. Because usingSince is zero or
    stale, useDuration() returns an extremely long duration, and Bow.Release treats it as a fully charged shot. This bypasses the
    minimum bow draw time.

    Preserve the prior state check before clearing it:

    if !p.usingItem {
        return
    }
    p.usingItem = false
    p.updateState()

    Then perform the remaining release validation. A regression test should call ReleaseItem without a preceding UseItem and
    assert that no projectile, durability damage, or ammunition consumption occurs.
```

FDUTCH:
> ```
> Found one high-severity issue:
> 
>   - [P1] Unsolicited release packets can fire a fully charged bow — server/player/player.go:1444
> 
>     ReleaseItem no longer checks p.usingItem before releasing the held item:
> 
>     p.usingItem = false
>     p.updateState()
>     if !p.canRelease() || !p.GameMode().AllowsInteraction() {
>         return
>     }
> 
>     A client can send a release transaction while holding a bow without first starting item use. Because usingSince is zero or
>     stale, useDuration() returns an extremely long duration, and Bow.Release treats it as a fully charged shot. This bypasses the
>     minimum bow draw time.
> 
>     Preserve the prior state check before clearing it:
> 
>     if !p.usingItem {
>         return
>     }
>     p.usingItem = false
>     p.updateState()
> 
>     Then perform the remaining release validation. A regression test should call ReleaseItem without a preceding UseItem and
>     assert that no projectile, durability damage, or ammunition consumption occurs.
> ```

fixed

PR #1147, opened 2025-12-23
Implement candles and candle cakes

xNatsuri:
This PR implements candles and candle cakes.


server/block/cake.go
```diff
@@ -46,12 +60,48 @@ func (c Cake) NeighbourUpdateTick(pos, _ cube.Pos, tx *world.Tx) {
...
+			c.Candle = true
+			c.Colour = candle.Colour
+			tx.SetBlock(pos, c, nil)
+			tx.PlaySound(pos.Vec3Centre(), sound.ItemUseOn{Block: c})
+			ctx.SubtractFromCount(1)
+			return true
+		}
+	}
+
+	if _, ok := held.Item().(item.FlintAndSteel); ok {
```
DaPigGuy: Does `Activate` even get called in this case?

server/block/cake.go
```diff
@@ -46,12 +60,48 @@ func (c Cake) NeighbourUpdateTick(pos, _ cube.Pos, tx *world.Tx) {
...
+		return false
+	}
+
+	if c.Candle {
+		if c.Lit {
```
DaPigGuy: Combine these two conditions

server/block/cake.go
```diff
@@ -63,18 +113,49 @@ func (c Cake) Activate(pos cube.Pos, _ cube.Face, tx *world.Tx, u item.User, _ *
...
+	c.Lit = true
+	tx.SetBlock(pos, c, nil)
+	tx.PlaySound(pos.Vec3(), sound.Ignite{})
+	return true
+}
+
 // BreakInfo ...
 func (c Cake) BreakInfo() BreakInfo {
+	if c.Candle {
+		return newBreakInfo(0.5, alwaysHarvestable, nothingEffective, silkTouchOnlyDrop(Candle{Colour: c.Colour}))
```
DaPigGuy: Doesn't look like you need silk touch in vanilla for the candle?

server/block/candle.go
```diff
@@ -0,0 +1,131 @@
...
+	Colour OptionalColour
+	// Candles is the number of candles.
+	Candles int
+	// Lit is whether the candles are lit.
+	Lit bool
+}
+
+// BreakInfo ...
+func (c Candle) BreakInfo() BreakInfo {
+	return newBreakInfo(0.1, alwaysHarvestable, shearsEffective, oneOf(c))
```
DaPigGuy: Why `shearsEffective`?

server/block/cake.go
```diff
@@ -16,6 +16,20 @@ type Cake struct {
 
 	// Bites is the amount of bites taken out of the cake.
 	Bites int
+	// Colour is the colour of the candle.
+	Colour item.OptionalColour
```
Sandertv: Would probably name this CandleColour.

server/block/cake.go
```diff
@@ -16,6 +16,20 @@ type Cake struct {
...
+	// Candle is true if the cake has a candle on top.
+	Candle bool
+	// Lit is whether the candle is lit.
+	Lit bool
```
Sandertv: And this maybe CandleLit.

server/item/optional_colour.go
```diff
@@ -0,0 +1,36 @@
...
+func NewOptionalColour(c Colour) OptionalColour {
+	return OptionalColour(c.Uint8() + 1)
+}
+
+// Colour returns the Colour of the OptionalColour.
+func (oc OptionalColour) Colour() (Colour, bool) {
+	if oc == 0 {
+		return Colour{}, false
+	}
+	return Colours()[(oc - 1)], true
```
Sandertv: Might be a good idea to store `Colours()` in a package level variable to avoid having to allocate these for every `Colour()` call. We don't have to worry about someone modifying the slice returned by Colours() here.

server/item/optional_colour.go
```diff
@@ -0,0 +1,36 @@
+package item
+
+// OptionalColour is an optional Colour for certain items and blocks.
+type OptionalColour uint8
```
Sandertv, line 8: You could consider adding some sort of function to simplify the encoding of an optional colour to something like `candle_cake`.

Maybe something like this:
```go
func (oc OptionalColour) Prepend(str string) string {
  if oc != 0 {
    return colours[(oc - 1)] + "_" + str
  }
  return str
}
```

server/item/optional_colour.go
```diff
@@ -0,0 +1,36 @@
...
+// Colour returns the Colour of the OptionalColour.
+func (oc OptionalColour) Colour() (Colour, bool) {
+	if oc == 0 {
+		return Colour{}, false
+	}
+	return Colours()[(oc - 1)], true
+}
+
+// Empty returns an empty OptionalColour.
+func (oc OptionalColour) Empty() OptionalColour {
```
Sandertv: Would consider removing this method and just creating a new `OptionalColour{}` where it's used.

Sandertv (commented):
Got a couple of ideas for this PR, but the idea with the OptionalColour is exactly what I had in mind. Did you run into any problems doing it this way?

xNatsuri:
All requested changes have been made.

server/block/cake.go
```diff
@@ -46,12 +60,48 @@ func (c Cake) NeighbourUpdateTick(pos, _ cube.Pos, tx *world.Tx) {
...
+			return true
+		}
+	}
+
+	if _, ok := held.Item().(item.FlintAndSteel); ok {
+		return false
+	}
+
+	if c.Candle {
+		if c.CandleLit {
```
Sandertv: Would merge this into `c.Candle && c.CandleLit`

server/block/candle.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// UseOnBlock ...
+func (c Candle) UseOnBlock(pos cube.Pos, face cube.Face, _ mgl64.Vec3, tx *world.Tx, user item.User, ctx *item.UseContext) (used bool) {
+	if existing, ok := tx.Block(pos.Add(cube.Pos{0, 1, 0})).(Candle); ok {
+		if existing.Colour.Uint8() != c.Colour.Uint8() {
+			return false
+		}
+
+		if existing.Candles >= 3 {
```
Sandertv: Would merge these into `existing.Colour != c.Colour || existing.Candles >= 3`
Also does not require comparing the Uint8s.

server/block/candle.go
```diff
@@ -0,0 +1,120 @@
...
+
+	place(tx, pos, c, user, ctx)
+	return placed(ctx)
+}
+
+// NeighbourUpdateTick ...
+func (c Candle) NeighbourUpdateTick(pos, _ cube.Pos, tx *world.Tx) {
+	if liquid, ok := tx.Liquid(pos); ok {
+		if liquid.LiquidType() == "water" {
+			if !c.Lit {
```
Sandertv: I believe this check should be `if c.Lit {}`.

I would also merge these checks like so:
```go
liquid, _ := tx.Liquid(pos)
if _, ok := liquid.(Water); ok && c.Lit {
    // ...
}
```

server/block/candle.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Ignite ...
+func (c Candle) Ignite(pos cube.Pos, tx *world.Tx, _ world.Entity) bool {
+	if c.Lit {
+		return false
+	}
+
+	if _, ok := tx.Liquid(pos); ok {
+		return false
+	}
```
Sandertv, line 118: Could also merge these checks if you would like.

Sandertv (changes requested):
Got a couple of final comments from my end, mostly style stuff, looks good to me otherwise!

Sandertv (dismissed):

server/block/cake.go
```diff
@@ -16,6 +16,20 @@ type Cake struct {
 
 	// Bites is the amount of bites taken out of the cake.
 	Bites int
+	// CandleColour is the colour of the candle.
+	CandleColour item.OptionalColour
+	// Candle is true if the cake has a candle on top.
+	Candle bool
```
DaPigGuy:
```suggestion
	// Candle is true if the cake has a candle on top.
	Candle bool
	// CandleColour is the colour of the candle.
	CandleColour item.OptionalColour
```

Probably makes more sense to flip the order of these two

server/block/cake.go
```diff
@@ -63,18 +114,48 @@ func (c Cake) Activate(pos cube.Pos, _ cube.Face, tx *world.Tx, u item.User, _ *
...
+		if flammable.OnFireDuration() > 0 {
+			c.Ignite(pos, tx, e)
+		}
+	}
+}
+
 // BreakInfo ...
 func (c Cake) BreakInfo() BreakInfo {
+	if c.Candle {
+		return newBreakInfo(0.5, alwaysHarvestable, nothingEffective, silkTouchOnlyDrop(Candle{Colour: c.CandleColour}))
```
DaPigGuy:
```suggestion
		return newBreakInfo(0.5, alwaysHarvestable, nothingEffective, oneOf(Candle{Colour: c.CandleColour}))
```

server/block/cake.go
DaPigGuy, line 1: Missing water log extinguish behavior

server/block/candle.go
```diff
@@ -0,0 +1,140 @@
...
+
+// Candle is a dyeable block that emits light when lit with a flint and steel.
+// Up to four of the same color of candle can be placed in one block space, which affects the amount of light produced.
+type Candle struct {
+	transparent
+	sourceWaterDisplacer
+
+	// Colour is the colour of the candle.
+	Colour item.OptionalColour
+	// Candles is the number of candles.
```
DaPigGuy: This doesn't seem entirely accurate since `Candle{Candles: 0}` actually has one candle? Perhaps `AdditionalCandles`?

server/block/candle.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// BreakInfo ...
+func (c Candle) BreakInfo() BreakInfo {
+	return newBreakInfo(0.1, alwaysHarvestable, nothingEffective, simpleDrops(item.NewStack(c, c.Candles+1)))
+}
+
+// Model ...
+func (c Candle) Model() world.BlockModel {
+	return model.Candle{Count: c.Candles}
```
DaPigGuy: Shouldn't this be `c.Candles + 1` if in the model there is a case for count of 4?

server/block/candle.go
```diff
@@ -0,0 +1,140 @@
...
+			return true
+		}
+		return false
+	}
+
+	if added := addCandle(pos); added {
+		return placed(ctx)
+	}
+
+	if added := addCandle(pos.Side(cube.FaceUp)); added && face == cube.FaceUp {
```
DaPigGuy: Why only for Up face?

server/block/candle.go
```diff
@@ -0,0 +1,140 @@
...
+		return placed(ctx)
+	}
+
+	pos, _, used = firstReplaceable(tx, pos, face, c)
+	if !used {
+		return false
+	}
+
+	place(tx, pos, c, user, ctx)
```
DaPigGuy, line 85: Should check if there is not air under the position first, i.e. when placing against a wall

PR #1152, opened 2025-12-24
session: Detect enum changes and promote to soft enums dynamically, fix suffixes.

### LINKED ISSUE #623: Command enum type does not show up
I am trying to implement a /give command, and I want to have the suffix of the itemType be 'item' so it would look like `<itemType: Item>`
However, no matter what I do it always seems to be string.
image

My command looks like this
```go
type Give struct {
	Player []cmd.Target
	Item   Item `cmd:"itemType,Item"`
	Count  cmd.Optional[uint]
	Data   cmd.Optional[int16]
}
```
and the Item enum type like this
```go
type Item string

func (i Item) Type() string {
	return "Item"
}

func (i Item) Options(source cmd.Source) []string {
	return []string{"apple"}
}
```
Neither the suffix in the struct field tag nor the one returned in Type() seem to have any effect.

didntpot:
Resolves #623


Sandertv (approved):
Thanks for the PR, very much appreciated. All looks good.

PR #1156, opened 2025-12-25
add knockback handler

FDUTCH:
not every knockback is a result of player been punched, so I think we need ability to control every knockback


FDUTCH:
> I suggest combining HandleEntityAttack and this handler into a single version. Perhaps also specifying the cause of the knockback (e.g. KnockBackCause)

I like the idea of having as much information as possible when handling events, but it seems that the guys here prefer simplicity, so I think we can avoid adding KnockBackCause (as I described in discord, there is a way to get knockback cause)

FDUTCH:
@Sandertv any suggestions?

FDUTCH:
I need opinion on this pr

PR #1164, opened 2026-01-04
update loader to allow custom load strategies

schphe:


server/world/loader.go
```diff
@@ -1,19 +1,173 @@
...
 	"math"
 	"sync"
+
+	"github.com/go-gl/mathgl/mgl64"
 )
 
+// LoadStrategy determines which chunks should be loaded and when they should be unloaded.
+type LoadStrategy interface {
+	// Load returns the chunks that should be loaded, ordered by priority (closest first).
+	Load(pos ChunkPos) []ChunkPos
```
HashimTheArab: "ordered by priority" should be documented on the implementation method or the struct not the interface

schphe: i've updated the comments

PR #1168, opened 2026-01-09
Implement Armour Stand

AkmalFairuz:
- [x] Poses
- [x] Sounds (Place, Land, Break, Hit)
- [x] Add or modify the armours/held Items

image


AkmalFairuz:
Bump

HashimTheArab:
Findings

  1. High: armour_stand_behaviour.go:129 only matches concrete item.Helmet / item.Chestplate / item.Leggings / item.Boots values after already deciding the stack is item.Armour. Dragonfly accepts armour by
     slot interfaces elsewhere, so valid items like item.Elytra, item.TurtleShell, carved pumpkins, and skulls fall into the armour branch, hit no case, and still execute ctx.SubtractFromCount(1). Result: the player loses the item and the stand
     does not equip it.
  2. High: armour_stand.go:42 assumes the Armor NBT list has four elements and indexes armours[i] unconditionally. nbtconv.Slice returns nil or short slices when the tag is missing or incomplete, so loading
     an armour stand with absent or truncated armour NBT will panic the server.
  3. Medium: armour_stand_behaviour.go:162 plays sound.ArmourStandPlace{} on hit. The PR adds a dedicated hit sound path in armour_stand.go:6 and server/session/world.go:890, but it is never used, so hit feedback is wrong and the new sound support is dead code.
  
  
Additional findings

  1. High: /server/entity/armour_stand_behaviour.go:167 destroys the entity without ever dropping the armour stand item itself. destroy() only calls dropAll(), and /server/entity/armour_stand_behaviour.go:179 only drops equipped armour and held
     items. Breaking a placed stand therefore permanently deletes the stand instead of returning item.ArmourStand{}.
  2. High: /server/item/armour_stand.go:15 places the stand at pos.Side(face).Vec3Middle() with no replaceable-block, clearance, or face validation. That means you can place armour stands by clicking walls or ceilings, and you can spawn them
     into occupied space even though their bbox is almost two blocks tall (/server/entity/armour_stand.go:31). The placement rules are much looser than the entity’s geometry supports.
  3. Medium: The new interaction API is not slot-aware. /server/entity/ent.go:21 defines ItemAcceptor.AcceptItem without any clicked-position data, and /server/player/player.go:1743 forwards only the player, tx, and use context. As a result, /
     server/entity/armour_stand_behaviour.go:105 can only infer behavior from the held item type, not where the player clicked on the stand. In practice this blocks core armour-stand interactions like removing equipment with an empty hand and
     targeting distinct held-item slots.

Silentmussle913:
bump

PR #1183, opened 2026-02-17
server/block: Add block state accessors and cmd/blocksetter

AkmalFairuz:
Introduces `WithColour`, `WithAxis`, and `WithFacing` support for blocks so we can properly handle state-aware transformations in the WorldEdit server. The main use case is implementing a `//rotate` command that rotates selected regions while preserving block properties. Rotation only applies to blocks that depend on orientation, specifically those using Axis and Facing. By exposing these modifiers directly on the block interface, we can transform block states cleanly and deterministically during region edits without relying on hardcoded type switches.

Previously, the alternative approach relied on `reflect` to modify block state dynamically. That worked functionally, but the performance overhead was noticeable, especially when rotating large selections. Reflection adds unnecessary runtime cost and reduces type safety. With explicit `WithAxis` and `WithFacing` methods, rotation becomes faster.

Example:
```go
for pos, b := range blocks {
  if a, ok := b.(block.HasFacing); ok {
    b = a.WithFacing(targetFace)
    tx.SetBlock(pos, b)
  }
}
```


Silentmussle913:
bump

HashimTheArab:
Pushed 42f98f76 with some cleanup:

- Generator now uses `go/packages` (like `cmd/blockhash`) instead of the deprecated `parser.ParseDir`, and emits via a `text/template` + setter-spec table with `format.Source`, which cuts it roughly in half.
- It skips blocks that already define a hand-written method with the same name instead of emitting a duplicate-method compile error, and excludes the generated file itself when scanning.
- Dropped `HasColour`/`WithColour` for now — colour isn't affected by `//rotate`, and on blocks like `Bed`/`Banner` it's item/NBT metadata rather than encoded block state, so it deserves its own PR if there's a concrete use case. Easy to re-add as one more spec entry.
- Added docs noting the known gaps: `cube.Face`-oriented blocks (torches, levers, hoppers, item frames, …) don't implement `HasFacing`, and the setters don't touch the other half of multi-block structures (beds, doors, chest pairs) — callers rotating regions own that.

Generated `setter.go` output is byte-identical for the 26 facing + 12 axis blocks apart from the updated doc comments.

HashimTheArab:
@AkmalFairuz @didntpot review please thanks its cleaned up

HashimTheArab:
Restored `HasColour`/`WithColour` — it's useful for recolouring edits on its own, and with the spec table it's a six-line entry. Kept a doc note that on beds/banners the colour is stored as block entity data rather than encoded block state.

Also merged your master merge and re-ran `go generate`, which picked up `BambooBlock` and `InfestedDeepslate` for `WithAxis`.

HashimTheArab:
Added getters to the interfaces so state can be read through them too, e.g. `b.WithAxis(b.PillarAxis().RotateRight())`.

A method can't share its name with a struct field, so `Facing()`/`Axis()`/`Colour()` aren't possible on these blocks. Rather than `Get*` (not used anywhere in dragonfly), the names follow the existing `world.Liquid` convention — `Water` has `Depth`/`Falling` fields and exposes them as `LiquidDepth()`/`LiquidFalling()` next to `WithDepth(...)`:

- `FacingDirection() cube.Direction` — matches the vanilla `facing_direction`/`minecraft:cardinal_direction` state names
- `PillarAxis() cube.Axis` — matches the vanilla `pillar_axis` state name
- `DyeColour() item.Colour`

HashimTheArab:
Two guard rails added:

- The generator now **fails** (instead of skipping) if a block defines a hand-written `WithFacing`/`FacingDirection`/etc., so the semantics of these methods can't quietly diverge per block. If a block ever legitimately needs custom behaviour here, the right escape hatch would be a `//blocksetter:` directive like `cmd/blockhash` uses, not a silent override.
- The PR/push workflows get a **Verify generated code** step: `go generate ./server/block/` + `git diff --exit-code`. This enforces `setter.go` freshness and, as a bonus, catches a stale `hash.go` whenever someone adds a block without running go generate — both generators are deterministic, so a clean tree stays clean.

Note custom blocks outside `server/block` can still implement `HasFacing` etc. by hand — Go interfaces are structural and that's desirable (WorldEdit can rotate third-party blocks). The guard only covers the generated package, where a manual copy would be drift.

HashimTheArab (dismissed):

didntpot (dismissed):

didntpot (approved):

.github/workflows/pr.yml
```diff
@@ -20,3 +20,8 @@ jobs:
 
       - name: Test
         run: go test ./...
+
+      - name: Verify generated code
```
RestartFU, line 24:
```suggestion
      - name: Generate Block code
```

.github/workflows/push.yml
```diff
@@ -21,6 +21,11 @@ jobs:
       - name: Test
         run: go test ./...
 
+      - name: Verify generated code
```
RestartFU, line 24:
```suggestion
      - name: Generate Block code
```

RestartFU (approved):

PR #1189, opened 2026-02-22
feat(blocks): Implement cobweb

RestartFU:


server/block/cobweb.go
```diff
@@ -0,0 +1,60 @@
+package block
+
+import (
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// Cobweb is a block that drastically slows down the movement of most entities touching it.
+type Cobweb struct {
```
DaPigGuy, line 10: Implement LiquidRemovable

server/block/cobweb.go
```diff
@@ -0,0 +1,60 @@
...
+
+// BreakInfo ...
+func (c Cobweb) BreakInfo() BreakInfo {
+	return newBreakInfo(4, alwaysHarvestable, func(t item.Tool) bool {
+		return t.ToolType() == item.TypeShears || t.ToolType() == item.TypeSword
+	}, func(t item.Tool, enchantments []item.Enchantment) []item.Stack {
+		if t.ToolType() == item.TypeShears || hasSilkTouch(enchantments) {
+			return []item.Stack{item.NewStack(c, 1)}
+		}
+		return nil
```
DaPigGuy, line 57: TODO: Add String item drop

server/block/register.go
```diff
@@ -248,6 +249,7 @@ func init() {
 	world.RegisterItem(ChiseledQuartz{})
 	world.RegisterItem(Clay{})
 	world.RegisterItem(Coal{})
+	world.RegisterItem(Cobweb{})
```
DaPigGuy: Make alphabetical

server/block/register.go
```diff
@@ -24,6 +24,7 @@ func init() {
 	world.RegisterBlock(Calcite{})
 	world.RegisterBlock(Clay{})
 	world.RegisterBlock(Coal{})
+	world.RegisterBlock(Cobweb{})
```
DaPigGuy: Make alphabetical

RestartFU: done

server/block/register.go
RestartFU: done

server/block/cobweb.go
RestartFU: added a todo, not sure if that's what you meant, or to implement strings aswell

server/block/cobweb.go
RestartFU: done

server/block/cobweb.go
```diff
@@ -0,0 +1,69 @@
...
+// EntityInside ...
+func (Cobweb) EntityInside(_ cube.Pos, _ *world.Tx, e world.Entity) {
+	if fallEntity, ok := e.(fallDistanceEntity); ok {
+		fallEntity.ResetFallDistance()
+	}
+	if v, ok := e.(velocityEntity); ok {
+		vel := v.Velocity()
+		vel[0] *= 0.25
+		vel[1] *= 0.05
+		vel[2] *= 0.25
```
TwistedAsylumMC, line 30: From reading the wiki, this logic only applies to players. We should probably have some sort of way to handle this for all entity types too

server/block/cobweb.go
```diff
@@ -0,0 +1,69 @@
...
+	return false
+}
+
+// HasLiquidDrops ...
+func (c Cobweb) HasLiquidDrops() bool {
+	return true
+}
+
+// BreakInfo ...
+func (c Cobweb) BreakInfo() BreakInfo {
```
TwistedAsylumMC, line 46: Cannot remember off the top of my head if dragonfly supports it, but using a sword to break cobwebs should deduct durability by 2 instead of 1

server/block/cobweb.go
TwistedAsylumMC: Also worth noting https://minecraft.wiki/w/Weaving can reduce the movement speed, so something will need to be done to make that possible to implement, even if you don't actually implement that here

TwistedAsylumMC (changes requested):

server/block/cobweb.go
TwistedAsylumMC: This is already a thing, ignore me!

server/block/cobweb.go
RestartFU: yep, I plan on implementing weaving in another PR

TwistedAsylumMC: Does it not make sense to put the base in this PR so you're not re-doing all the work you've already done? Technically this physics behaviour isn't correct as it stands. It works for normal players, but it's different for other entities and even more different for item entities. Have you tested to see if this works correctly with jumping behaviour too?

> the player can move at a speed of about 25% of the normal [walking](https://minecraft.wiki/w/Player#Movement) speed, and their [jumping](https://minecraft.wiki/w/Jumping) height is severely reduced

From the wiki, but could just mean the same 25% reduction

RestartFU: > Does it not make sense to put the base in this PR so you're not re-doing all the work you've already done? Technically this physics behaviour isn't correct as it stands. It works for normal players, but it's different for other entities and even more different for item entities. Have you tested to see if this works correctly with jumping behaviour too?
> 
> > the player can move at a speed of about 25% of the normal [walking](https://minecraft.wiki/w/Player#Movement) speed, and their [jumping](https://minecraft.wiki/w/Jumping) height is severely reduced
> 
> From the wiki, but could just mean the same 25% reduction

I don't see the point of implementing the base of weaving in this PR, if that's what you mean by base?

TwistedAsylumMC: I mean the base of having different multipliers for velocity, not specifically for weaving

PR #1192, opened 2026-02-25
chunk/light.go: optimize light calculations (4400x fewer allocs)

HashimTheArab:
- Replaced linked-list BFS queue with a ring-buffer queue for light propagation (Fill/Spread) to remove per-node heap churn.
- Removed iterator-based neighbor traversal in propagation; now uses direct 6-neighbor checks.
- Removed small slice allocations in world light orchestration (chunk() and spreadLight()).

# Benchmark results
image

image

# Benchmark Code
```go
package chunk

import (
	"testing"

)

func BenchmarkLightAreaFill(b *testing.B) {
	prepareLightBenchTables()

	c := benchmarkChunk()
	area := LightArea([]*Chunk{c}, 0, 0)

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		area.Fill()
	}
}

func BenchmarkLightAreaSpread(b *testing.B) {
	prepareLightBenchTables()

	chunks := make([]*Chunk, 0, 9)
	fillAreas := make([]*lightArea, 0, 9)
	for z := -1; z <= 1; z++ {
		for x := -1; x <= 1; x++ {
			c := benchmarkChunk()
			chunks = append(chunks, c)
			fillAreas = append(fillAreas, LightArea([]*Chunk{c}, x, z))
		}
	}
	spread := LightArea(chunks, -1, -1)

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		for _, area := range fillAreas {
			area.Fill()
		}
		spread.Spread()
	}
}

func prepareLightBenchTables() {
	if len(LightBlocks) < 4 {
		LightBlocks = make([]uint8, 4)
	}
	if len(FilteringBlocks) < 4 {
		FilteringBlocks = make([]uint8, 4)
	}

	// Runtime IDs used in benchmarkChunk():
	// 0: air, 1: full blocker, 2: emitter, 3: partial diffuser.
	LightBlocks[0], LightBlocks[1], LightBlocks[2], LightBlocks[3] = 0, 0, 14, 0
	FilteringBlocks[0], FilteringBlocks[1], FilteringBlocks[2], FilteringBlocks[3] = 0, 15, 0, 2
}

func benchmarkChunk() *Chunk {
	c := New(0, cube.Range{0, 63})

	// Opaque floor.
	for x := uint8(0); x < 16; x++ {
		for z := uint8(0); z < 16; z++ {
			c.SetBlock(x, 0, z, 0, 1)
		}
	}

	// A few emitters spread through the chunk and on the edges to exercise both
	// fill and neighbour spreading.
	for x := uint8(1); x < 16; x += 4 {
		for z := uint8(1); z < 16; z += 4 {
			c.SetBlock(x, 2, z, 0, 2)
		}
	}
	for _, x := range [...]uint8{0, 15} {
		for _, z := range [...]uint8{0, 15} {
			c.SetBlock(x, 2, z, 0, 2)
		}
	}

	// A partially diffusing strip to force filtered propagation paths.
	for x := uint8(0); x < 16; x += 2 {
		c.SetBlock(x, 3, 7, 0, 3)
	}
	return c
}

```


server/world/chunk/light_area.go
```diff
@@ -16,6 +14,56 @@ type lightArea struct {
 	r            cube.Range
 }
 
+type lightQueue struct {
```
TwistedAsylumMC, line 20: If you could document the lightQueue that would be great

server/world/world.go
```diff
@@ -1213,19 +1214,21 @@ func (w *World) calculateLight(centre ChunkPos) {
...
 				return
 			}
-			c = append(c, neighbour.Chunk)
+			chunks[i] = neighbour.Chunk
+			i++
 		}
 	}
 	// All chunks surrounding the current one are present, so we can spread.
-	chunk.LightArea(c, int(pos[0])-1, int(pos[1])-1).Spread()
+	chunk.LightArea(chunks[:], int(pos[0])-1, int(pos[1])-1).Spread()
```
Sandertv: The changes in this file don't seem consequential? No allocations should be saved by replacing the `append`s here since the slice is pre-allocated, and doing `[:]` and passing it to a function should still allocate a slice itself too.

Sandertv: Even if anything changes, any improvement would be so minimal that I don't think it's worth polluting the code over it.

server/world/chunk/light_area.go
HashimTheArab: done

server/world/chunk/light_area.go
```diff
@@ -30,11 +78,11 @@ func LightArea(c []*Chunk, baseX, baseY int) *lightArea {
 // individual chunks within the lightArea itself, without light crossing chunk borders.
 func (a *lightArea) Fill() {
 	a.initialiseLightSlices()
-	queue := list.New()
+	queue := newLightQueue(512)
```
Sandertv: Are these pre-allocated sizes based on anything? Could we do some tests with a real world to see what a suitable cap might be if that wasn't done yet?

Sandertv (commented):
Good changes, a ring buffer will definitely help bring down allocations. Got a couple of minor comments

server/world/world.go
HashimTheArab: reverted

server/world/chunk/light_area.go
HashimTheArab: will need @AkmalFairuz to do this

cjmustard: @AkmalFairuz any word?

Silentmussle913 (approved):

T14Raptor (dismissed):

RestartFU (dismissed):

cqdetdev (dismissed):
lgtm

server/world/chunk/light_area.go
cqdetdev: After doing further testing, 512 looks to be best size (after doing a simulated test). The results show 1024 could be a good candidate but that would require more detailed ingame metrics that just wouldn't be worth for the minimal performance gains we could get from going to 1024. I did test past 2048 we had memory degradation and it didn't improve runtime so 512 looks good.

HashimTheArab:
I did some more testing after cadet and updated the value to 1024, and named it as a constant to be more clear and easier to change if needed. For anything more than a basic server such as vanilla or skyblock 512 will almost always grow. The difference between 512 and 1024 is negligible (24KiB). 1024 is the safer value, anything past that and the benefit drops quickly. If a less common case needs more than 1024 the queue can still grow normally

RestartFU (approved):

cqdetdev (approved):

PR #1201, opened 2026-03-16
feat: implement wind charges

HashimTheArab:
## Summary

- implement wind charges and their block interactions
- match Java launch inheritance, spread, collision, burst offset, exposure, and knockback physics
- allow projectile collision filters so wind charges can collide with non-living entities without changing other projectile behavior


server/block/wood_fence_gate.go
```diff
@@ -80,7 +80,7 @@ func (f WoodFenceGate) shouldBeLowered(pos cube.Pos, tx *world.Tx) bool {
 // Activate ...
 func (f WoodFenceGate) Activate(pos cube.Pos, _ cube.Face, tx *world.Tx, u item.User, _ *item.UseContext) bool {
 	f.Open = !f.Open
-	if f.Open && f.Facing.Opposite() == u.Rotation().Direction() {
+	if f.Open && u != nil && f.Facing.Opposite() == u.Rotation().Direction() {
```
TwistedAsylumMC, line 83: Is this change necessary?

server/entity/wind_charge.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// Toggle interactive blocks at the impact point.
+	if r, ok := target.(trace.BlockResult); ok {
+		toggleWindChargeBlock(r.BlockPosition(), r.Face(), tx)
+	}
+}
+
+// toggleWindChargeBlock toggles a block at pos if it is a door, trapdoor or
+// fence gate. Other activatable blocks (chests, crafting tables, etc.) are not
```
TwistedAsylumMC: I personally wouldn't consider chests and crafting tables to be "activatable"

server/entity/wind_charge.go
```diff
@@ -0,0 +1,119 @@
...
+// affected.
+func toggleWindChargeBlock(pos cube.Pos, face cube.Face, tx *world.Tx) {
+	b := tx.Block(pos)
+	switch b.(type) {
```
TwistedAsylumMC: Probably worth adding a TODO comment for Buttons, Levers, Bells and Candles

server/entity/wind_charge.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+var windChargeConf = ProjectileBehaviourConfig{
+	Gravity:  0,
+	Drag:     0,
+	Damage:   -1,
+	Particle: particle.WindExplosion{},
+	Sound:    sound.WindChargeBurst{},
+	Hit:      windChargeBurst,
+}
```
TwistedAsylumMC, line 41: Does this account for the distance attenuation when going through lava/water?

server/item/wind_charge.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// WindCharge is a throwable item that creates a burst of wind on impact, knocking back nearby entities and
+// toggling certain blocks such as doors, trapdoors and fence gates.
+type WindCharge struct{}
+
+// Use ...
+func (WindCharge) Use(tx *world.Tx, user User, ctx *UseContext) bool {
+	create := tx.World().EntityRegistry().Config().WindCharge
+	opts := world.EntitySpawnOpts{Position: eyePosition(user), Velocity: user.Rotation().Vec3().Mul(3.0)}
```
TwistedAsylumMC: Should this account for the randomness stated on the wiki?:
> There is a random offset in a wind charge's trajectory over a cone-shaped region, spreading along the aimed trajectory, which can cause a wind charge to miss its target or hit something else.

TwistedAsylumMC (commented):
Got a few comments on the code, after testing them for a minute they are definitely not anything close to how they behave in vanilla. Would not say "opus 1 shot this" personally

server/block/wood_fence_gate.go
HashimTheArab: Yea check toggleWindChargeBlock

server/entity/wind_charge.go
HashimTheArab: Looking at the codebase, no other projectile handles water/lava drag either — every projectile (snowball, egg, ender pearl, arrow) uses a single  uniform Drag value that applies identically in air, water, and lava. There's no per-medium drag support in ProjectileBehaviour at all.

server/item/wind_charge.go
HashimTheArab: According to the wiki this applies to all projectiles, and no other projectiles in dragonfly implement this
https://minecraft.wiki/Projectile
image

server/entity/wind_charge.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+}
+
+// toggleWindChargeBlock toggles a block at pos if it is a door, trapdoor or
+// fence gate.
+// TODO: Buttons, levers, bells and candles should also be toggled.
+func toggleWindChargeBlock(pos cube.Pos, face cube.Face, tx *world.Tx) {
+	b := tx.Block(pos)
+	switch b.(type) {
+	case block.WoodDoor, block.CopperDoor, block.WoodTrapdoor, block.CopperTrapdoor, block.WoodFenceGate:
```
DaPigGuy: some sort of WindChargeAffected interface might be nicer

server/item/wind_charge.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Cooldown ...
+func (WindCharge) Cooldown() time.Duration {
+	return time.Millisecond * 500
+}
+
+// MaxCount ...
+func (WindCharge) MaxCount() int {
+	return 64
```
DaPigGuy: This is the default

HashimTheArab: fixed

server/entity/wind_charge.go
HashimTheArab: fixed

HashimTheArab:
anything else?

server/block/block.go
```diff
@@ -21,6 +21,22 @@ type Activatable interface {
...
+	// should be applied. Multi-block structures use this to collapse multiple
+	// affected positions into one interaction.
+	WindChargeInteractionPos(pos cube.Pos) cube.Pos
+	// WindChargeInteract applies the block's wind charge interaction.
+	WindChargeInteract(pos cube.Pos, tx *world.Tx)
+}
+
+// ProjectileIgniter represents a block or liquid that sets projectiles passing
+// through it on fire.
+type ProjectileIgniter interface {
```
FDUTCH, line 36: there is no need to check if block can ignite projectile... if block can ignite entities - block can ignite projectiles, I have pull request that handles such behavior: #1212

HashimTheArab: The dedicated check is still needed here. Wind charges move up to 1.5 blocks per tick, and `igniteProjectileAlongPath` sweeps every crossed voxel so they cannot tunnel through fire or lava. #1212 currently checks only the entity bounding box at its tick position and is not merged, so replacing this with `EntityInside` would miss intermediate igniters. Keeping this implementation for the projectile-specific swept-path behaviour.

FDUTCH: Sure, but are you sure that's how it works in the vanilla? I'm not against this change, but maybe we can check for block insiders instead?

PR #1202, opened 2026-03-18
Move crafting logic into player and add HandleCraftItem

AkmalFairuz:
## Description
This PR moves normal crafting logic out of `server/session` and into `server/player`.

**Notes:** HandleCraftItem is only used for crafting-table crafting (3x3), including the player 2x2 grid. Smithing, loom, grindstone, enchanting, etc. still use their own handlers

## API Changes
- added `HandleCraftItem(ctx *Context, recipes []item.Stack, results []item.Stack)` to `player.Handler`.
- moved crafting plan generation for normal craft, auto-craft, and dynamic craft into `server/player/crafting.go`
- kept session responsible for packet handling and applying inventory/stack response changes
- introduced shared `recipe.CraftingPlan` / `CraftingSlotChange`

## Tests
image


Silentmussle913:
bump

PR #1203, opened 2026-03-19
server/world: use slices.Clone to safely iterate and fix panic in world.entitiesWithin

Dasciam:


AkmalFairuz (dismissed):

cjmustard (dismissed):

server/world/world.go
FDUTCH:
```suggestion
ent, ok := handle.Entity(tx)
if ok && !yield(ent) {
	return
}
```

FDUTCH: I hate github, but I guess you get the idea

Dasciam: I don't get why this code needs to be changed like that

FDUTCH: If you gonna remove entity from the world while iterating, mustEntity will panic

Dasciam: > If you gonna remove entity from the world while iterating, mustEntity will panic

mustEntity will not panic here, the point of the PR is to clone the slice, because removing an entity while iterating the slice nils elements that were before removing an entity

FDUTCH: image
image
:scissors:

HashimTheArab: fdutch is correct, it is a different panic case, its so handles that are removed after the snapshot are skipped. If you can apply this we'll merge the PR just click "apply suggestion". I verified with a test and AI review

Calling handle.mustEntity on an entity that was removed after taking the slices.Clone snapshot will panic

HashimTheArab (approved):

AkmalFairuz (approved):
LGTM!

PR #1205, opened 2026-03-20
Implement Sweet Berries

AkmalFairuz:


AkmalFairuz:
Duplicate #1197

server/block/sweet_berries.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/go-gl/mathgl/mgl64"
+)
+
+// SweetBerries are edible berries that can also be planted to grow a sweet berry bush.
+type SweetBerries struct {
+	transparent
+	empty
+
+	// Growth is the current growth stage of the bush when placed as a block. 7 is fully grown.
```
DaPigGuy: 3 is fully grown, but there are 4 additional growth states because Mojang..

DaPigGuy: Usually we don't register those extra states

server/block/sweet_berries.go
```diff
@@ -0,0 +1,175 @@
...
+
+	s.Growth = 0
+	place(tx, pos, s, user, ctx)
+	return placed(ctx)
+}
+
+// Pick ...
+func (SweetBerries) Pick() item.Stack {
+	return item.NewStack(SweetBerries{}, 1)
+}
```
DaPigGuy: Should no longer be needed after merging the block and item

server/block/sweet_berries.go
```diff
@@ -0,0 +1,175 @@
...
+	} else {
+		s.Growth++
+	}
+	tx.SetBlock(pos, s, nil)
+	return true
+}
+
+// FlammabilityInfo ...
+func (SweetBerries) FlammabilityInfo() FlammabilityInfo {
+	return newFlammabilityInfo(60, 100, false)
```
DaPigGuy, line 69:
```suggestion
	return newFlammabilityInfo(30, 100, false)
```

30 on Bedrock according to wiki

server/block/sweet_berries.go
```diff
@@ -0,0 +1,175 @@
...
+		return false
+	}
+
+	count := rand.IntN(2) + 1
+	if s.Growth == 7 {
+		count++
+	}
+	dropItem(tx, item.NewStack(SweetBerries{}, count), pos.Vec3Centre())
+
+	s.Growth = 2
```
DaPigGuy:
```suggestion
	s.Growth = 1
```

Update to state with no berries

DaPigGuy (changes requested):

server/block/sweet_berries.go
```diff
@@ -0,0 +1,170 @@
...
+		vel[1] *= 0.75
+		vel[2] *= 0.8
+		v.SetVelocity(vel)
+	}
+
+	if s.Growth < 1 {
+		return
+	}
+	if math.Abs(movement[0]) >= 0.003 || math.Abs(movement[2]) >= 0.003 {
+		living.Hurt(0.5, DamageSource{Block: s})
```
DaPigGuy: Shouldn't this be 1?

Silentmussle913:
bump

server/block/sweet_berries.go
HashimTheArab: we should still map them for the sake of decoding (from maps / chunks from other servers via gopher) so they dont end up as unknown blocks

HashimTheArab (dismissed):

PR #1206, opened 2026-03-20
Implement Hanging Sign

AkmalFairuz:
This PR adds hanging sign support across block registration, placement, block actor handling, and sign editing.

## Changes
  - Added a new HangingSign block implementation in `server/block/hanging_sign.go`.
  - Registered hanging sign block permutations and item forms in `server/block/register.go`.
  - Added hanging sign hashing support in server/block/hash.go and updated the hash generator in `cmd/blockhash/main.go` to handle `cube.Orientation`.
  - Updated sign editing in `server/player/player.go` so players can edit both regular signs and hanging signs.
  - Updated `server/session/handler_block_actor_data.go` to handle `HangingSign` block actor packets and validate the expected sign type.


Silentmussle913:
bump

PR #1211, opened 2026-03-21
Implement Bell

AkmalFairuz:
This PR adds initial Bell support 

## API Changes
- `world.Viewer{}.ViewBlockUpdate` now takes an additional `nbtOnly bool` parameter:
  - This allows callers to send block entity NBT updates without forcing a full block runtime update.
  - Bell: sending an `packet.UpdateBlock` on every ring tick causes the ringing animation to look weird.


server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+}
+
+// SideClosed ...
+func (Bell) SideClosed(cube.Pos, cube.Pos, *world.Tx) bool {
+	return false
+}
+
+// ring starts or refreshes the Bell ring state.
+func (b Bell) ring(pos cube.Pos, tx *world.Tx, face cube.Face) bool {
```
TwistedAsylumMC, line 193: Should this be exposed so bells can be triggered through code?

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+	tx.SetBlock(pos, b, &world.SetOpts{DisableBlockUpdates: true, DisableLiquidDisplacement: true})
+	tx.PlaySound(pos.Vec3Centre(), sound.BellRing{})
+	return true
+}
+
+// multipleSupportFaces returns the faces checked for a Bell attached between two side supports.
+func (b Bell) multipleSupportFaces() (cube.Face, cube.Face) {
+	face := b.Facing.Face()
+	return face, face.Opposite()
+}
```
TwistedAsylumMC, line 205: Does this need to be a whole method if it's used just once?

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+
+// bellSupportSolid reports whether the block on the given side of the Bell can support it.
+func bellSupportSolid(tx *world.Tx, pos cube.Pos, face cube.Face) bool {
+	supportPos := pos.Side(face)
+	return tx.Block(supportPos).Model().FaceSolid(supportPos, face.Opposite(), tx)
+}
+
+// bellFace returns a cube.Face from the given int32, or the fallback if the int32 is not a valid horizontal face.
+func bellFace(v int32, fallback cube.Face) cube.Face {
```
TwistedAsylumMC, line 214: Why does this take an int32 and cast to a cube.Face when all but one callers go from a cube.Face to an int?

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if c.ItemDropChance < 0 {
+		return
+	}
+	if c.ItemDropChance < 1 && c.ItemDropChance <= rand.Float64() {
+		return
+	}
+	for _, drop := range b.BreakInfo().Drops(item.ToolNone{}, nil) {
+		dropItem(tx, drop, pos.Vec3Centre())
+	}
```
TwistedAsylumMC, line 184: Should there be some sort of generic method on the explosion config if this logic has to be done for every block that handles explosion drops individually? Or is some of the logic here unique to bells

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+func (b Bell) DecodeNBT(data map[string]any) any {
+	b.Toggle = nbtconv.Bool(data, "Ringing")
+	b.ringTicks = int(nbtconv.Int32(data, "Ticks"))
+	b.ringDirection = bellFace(nbtconv.Int32(data, "Direction"), b.Facing.Face())
+	return b
+}
+
+// Explode ...
+func (b Bell) Explode(_ mgl64.Vec3, pos cube.Pos, tx *world.Tx, c ExplosionConfig) {
+	b.ring(pos, tx, b.Facing.Face())
```
TwistedAsylumMC, line 169: This doesn't seem right, I cannot replicate this in vanilla

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+	if !b.Toggle {
+		return
+	}
+	b.ringTicks++
+	if b.ringTicks < 50 {
+		tx.SetBlock(pos, b, &world.SetOpts{DisableBlockUpdates: true, DisableLiquidDisplacement: true})
+		return
+	}
+	b.Toggle, b.ringTicks = false, 0
+	tx.SetBlock(pos, b, &world.SetOpts{DisableBlockUpdates: true, DisableLiquidDisplacement: true})
```
TwistedAsylumMC, line 132: Can probably simplify this?
```
b.ringTicks++
if b.ringTicks >= 50 {
	b.Toggle, b.ringTicks = false, 0
}
tx.SetBlock(pos, b, &world.SetOpts{DisableBlockUpdates: true, DisableLiquidDisplacement: true})
```

server/block/bell.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// BreakInfo ...
+func (b Bell) BreakInfo() BreakInfo {
+	return newBreakInfo(5, alwaysHarvestable, pickaxeEffective, oneOf(b)).withBlastResistance(5)
+}
+
+// Activate ...
+func (b Bell) Activate(pos cube.Pos, face cube.Face, tx *world.Tx, _ item.User, _ *item.UseContext) bool {
+	return b.ring(pos, tx, face)
```
TwistedAsylumMC, line 45: It should only ring when interacting with the correct axis

server/world/viewer.go
```diff
@@ -61,7 +61,7 @@ type Viewer interface {
 	ViewSound(pos mgl64.Vec3, s Sound)
 	// ViewBlockUpdate views the updating of a block. It is called when a block is set at the position passed
 	// to the method.
```
TwistedAsylumMC, line 63: Please document what nbtOnly is supposed to do

TwistedAsylumMC (changes requested):

Silentmussle913:
bump

PR #1212, opened 2026-03-21
Destroy item entity on contact with fire or cactus

FDUTCH:


server/entity/ent.go
```diff
@@ -123,14 +133,25 @@ func (e *Ent) Tick(tx *world.Tx, current int64) {
 		_ = e.Close()
 		return
 	}
-	e.SetOnFire(e.OnFireDuration() - time.Second/20)
 
-	if m := e.Behaviour().Tick(e, tx); m != nil {
+	beh := e.Behaviour()
+	e.SetOnFire(e.OnFireDuration() - time.Second/20)
+	e.checkBlockCollisions(beh)
+	if m := beh.Tick(e, tx); m != nil {
```
TwistedAsylumMC: Small nitpick, I'd call this `b` instead of `beh`

server/entity/ent.go
```diff
@@ -123,14 +133,25 @@ func (e *Ent) Tick(tx *world.Tx, current int64) {
...
 }
 
+// checkBlockCollisions checks block collisions for the entity.
+func (e *Ent) checkBlockCollisions(beh Behaviour) {
+	if bl, ok := beh.(interface {
+		CheckBlocks(e *Ent, tx *world.Tx) bool
+	}); ok && bl.CheckBlocks(e, e.tx) {
+		CheckEntityInsiders(e.tx, e.H().Type().BBox(e).Translate(e.Position()), e)
+	}
+}
```
TwistedAsylumMC, line 153: Running this interface check for every entity every tick isn't the prettiest, can you think of any other way to achieve this?

server/block/cactus.go
```diff
@@ -108,6 +109,11 @@ func allCactus() (b []world.Block) {
 	return
 }
 
+// PrickableEntity represents an entity that can be pricked.
+type PrickableEntity interface {
+	Prick(pos cube.Pos)
```
TwistedAsylumMC, line 115: Add a simple doc to the Prick method & explain what the provided position is

server/block/cactus.go
```diff
@@ -69,9 +70,9 @@ func (c Cactus) canGrowHere(pos cube.Pos, tx *world.Tx, recursive bool) bool {
 }
 
 // EntityInside ...
-func (c Cactus) EntityInside(_ cube.Pos, _ *world.Tx, e world.Entity) {
-	if l, ok := e.(livingEntity); ok {
-		l.Hurt(0.5, DamageSource{Block: c})
```
TwistedAsylumMC, line 74: There's a small regression with this change, you've made it now so only players get the 0.5 damage and not any other living entity

server/entity/item_behaviour.go
```diff
@@ -74,9 +75,18 @@ func (i *ItemBehaviour) Item() item.Stack {
 // Tick moves the entity, checks if it should be picked up by a nearby collector
 // or if it should merge with nearby item entities.
 func (i *ItemBehaviour) Tick(e *Ent, tx *world.Tx) *Movement {
+	if e.OnFireDuration() > 0 {
+		//TODO: replicate vanilla behavior.
```
TwistedAsylumMC, line 79: What needs to be done to replicate vanilla here?

TwistedAsylumMC (changes requested):

server/entity/ent.go
FDUTCH: I was thinking about simply adding a boolean field called "checkBlocks" to Ent, or making CheckBlocks part of the Behavior interface, what you think about that?

TwistedAsylumMC: With a better name, I would not be against it

server/block/cactus.go
FDUTCH: we already have explosion damage calculation only for player and I don't know if you okay with passing damage arg in Prick()

server/entity/item_behaviour.go
FDUTCH: in vanilla item entities do have health and items can survive short period of time burning

server/entity/ent.go
FDUTCH: are you okay with adding method or both options are fine?

FDUTCH: ...

PR #1216, opened 2026-03-28
server/world: add concurrent chunk loading, generation, and light updates

FDUTCH:
This pr implements concurrent chunk generation with no additional locks.
But it is not ready for production use and it requires testing on a multiplayer server.

PS. No AI was used...


FDUTCH:
okay, now it feels really stable

FDUTCH:
should I also call world.Provider asynchronously?

HashimTheArab:
take reference from this PR for some things https://github.com/HashimTheArab/dragonfly/pull/7

FDUTCH:
> take reference from this PR for some things [HashimTheArab#7](https://github.com/HashimTheArab/dragonfly/pull/7)

I don't really like how this is done in your brunch: 
1) it is too over complicated
2) it has additional locks
3) I don't like workers concept

I was planing to make pr in the future that will allow you to swap low level systems of the world to your own, like how world is handling ticks, chunk requests and such...

Silentmussle913:
bump

HashimTheArab:
## Concurrent chunk loading and generation

Loads and generates chunks on background workers instead of the world's transaction goroutine, keeping disk reads and lighting off the tick path.

**Features**
- Loads/generates chunks off the transaction goroutine on a background worker — the tick path no longer blocks on disk I/O, generation, or lighting (active by default)
- `Config.ChunkLoadWorkers` (default 1) — raise above 1 to generate multiple chunks in parallel; requires a concurrency-safe generator
- Per-chunk light `Fill` now runs on the worker, not the owner goroutine
- Provider calls serialised internally (providers need not be concurrency-safe)

**Behaviour changes**
- `Loader.Load` is now asynchronous — chunks appear after it returns
- `Tx.Light` returns 0 for unloaded chunks instead of generating them
- `World.HighestLightBlocker` is now off-owner only; use `Tx.HighestLightBlocker` in a transaction

**Fixes**
- `populateLoadQueue` no longer drops the outermost ring of the load radius (off-by-one)
- Chunk-distance no longer overflows `int32` for far-apart positions (e.g. long teleports)

HashimTheArab (dismissed):

NopeNotDark (approved):
LGTM, Thanks for the PR!

cqdetdev (dismissed):
LGTM!

HashimTheArab (dismissed):

HashimTheArab (approved):

RestartFU (approved):

PR #1217, opened 2026-03-29
server/block: Fix liquid-removable blocks dropping items when removed by lava

hydzilla:
Fixed liquid-removable blocks incorrectly dropping items when destroyed by lava. Only water should cause item drops, matching vanilla behaviour.


server/block/liquid.go
```diff
@@ -169,7 +170,7 @@ func flowInto(b world.Liquid, src, pos cube.Pos, tx *world.Tx, falling bool) boo
 		if _, air := existing.(Air); !air {
 			tx.SetBlock(pos, nil, nil)
 		}
-		if removable.HasLiquidDrops() {
+		if _, water := b.(Water); water && removable.HasLiquidDrops() {
```
RestartFU: I would do this instead, check if it isn’t lava rather than check if it’s water.

I know there’s only water and lava liquids, but I believe this is just more future-proof and defensive,
if another liquid ever gets added, this won’t silently break or behave incorrectly. It also makes the intent clearer: we’re excluding lava specifically, rather than assuming water is the only other liquid case.

```suggestion
		if _, lava := b.(Lava); !lava && removable.HasLiquidDrops() {
```

RestartFU (commented):
just a small NIT

RestartFU (approved):

RestartFU:
Seb suggested a different fix, I agree with him, I think his proposition is a better fix:
image
https://mcsrc.dev/1/26.1.2/net/minecraft/world/level/material/WaterFluid#L81-84
https://mcsrc.dev/1/26.1.2/net/minecraft/world/level/material/LavaFluid#L145-147

hydzilla:
Should be fine now.

HashimTheArab:
The gating in `flowInto` makes the fizz fix only half-work:

```go
if removable.HasLiquidDrops() {
    b.LiquidRemoveBlock(pos, tx, existing)
}
```

Several `LiquidRemovable` blocks return `HasLiquidDrops() == false`:

- `Vines` (`server/block/vine.go:40`)
- `Coral` (`server/block/coral.go:47`)
- `NetherSprouts` (`server/block/nether_sprouts.go:39`)
- `Fire` (`server/block/fire.go:238`)
- `Air` (`server/block/air.go:11`)

Vanilla fizzes when lava consumes vines, coral, nether sprouts, etc. Under this PR, lava removing any of those is silent because the call site short-circuits before `Lava.LiquidRemoveBlock` runs. The fizz only fires for the subset of removables that also drop items.

### Suggested fix

Remove the `HasLiquidDrops` if statement and let each liquid gate its own behavior:

```go
if isRemovable {
    if _, air := existing.(Air); !air {
        tx.SetBlock(pos, nil, nil)
    }
    b.LiquidRemoveBlock(pos, tx, existing)
}
```

Then move the `HasLiquidDrops` check into `Water.LiquidRemoveBlock`:

```go
func (Water) LiquidRemoveBlock(pos cube.Pos, tx *world.Tx, removed world.Block) {
    r, ok := removed.(LiquidRemovable)
    if !ok || !r.HasLiquidDrops() {
        return
    }
    b, ok := removed.(Breakable)
    if !ok {
        panic("liquid drops should always implement breakable")
    }
    for _, d := range b.BreakInfo().Drops(item.ToolNone{}, nil) {
        dropItem(tx, d, pos.Vec3Centre())
    }
}
```

`Lava.LiquidRemoveBlock` then fizzes unconditionally (skipping air, since lava \"destroying\" air shouldn't make sound):

```go
func (Lava) LiquidRemoveBlock(pos cube.Pos, tx *world.Tx, removed world.Block) {
    if _, air := removed.(Air); air {
        return
    }
    tx.PlaySound(pos.Vec3Centre(), sound.Fizz{})
}
```

That gives full vanilla parity for both behaviors.

### Other notes
- Minor: `LiquidRemoveBlock` reads as a verb but it's a callback after removal. `OnBlockRemoved` or `HandleBlockRemoved` would read more naturally. Not blocking.

hydzilla:
I've applied the requested changes, however it may be better to apply the air check before calling `LiquidRemoveBlock`. Let me know.

HashimTheArab (approved):

cjmustard (approved):

PR #1222, opened 2026-04-26
dragonfly: Updated for 1.26.20.

didntpot:


hydzilla:
Is this pending due to furnace recipe changes?

server/player/debug/shape.go
```diff
@@ -130,15 +130,29 @@ type Sphere struct {
...
+	BackgroundColour color.RGBA
 	// Position is the origin position of the shape in the world.
 	Position mgl64.Vec3
 	// Scale is the size of the text. If zero, it will default to 1.0.
 	Scale float64
 	// Text is the text to be displayed on the shape. The background automatically scales to fit the text.
 	Text string
+	// NoRotation specifies whether the text should stay static instead of rotating to always face the
+	// camera. If false, the text will rotate to face the camera.
+	NoRotation bool
```
TwistedAsylumMC: I'd maybe call this "Lock Rotation". Probably worth noting that when this is true, it uses the Rotation field which should also be added to this type

TwistedAsylumMC (changes requested):

HashimTheArab (approved):

voiddev20:
@TwistedAsylumMC 
can you merge this pull request?

PR #1227, opened 2026-04-27
dragonfly: Simplify CI workflows and add vulnerability check

HashimTheArab:
merge this after the 1.26 bump PR since this diff currently contains both changes

- **chore: bump Go to 1.26.0**
- **use go-version-file**
- **simplify workflow**


.github/workflows/pr.yml
```diff
@@ -14,10 +14,9 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
-        id: go
 
-      - name: Build
-        run: go build -o dragonfly_exe -v .
+      - name: Test
+        run: go test ./...
```
RestartFU: we don't have any tests yet, and I think verifying build should stay

RestartFU (dismissed):

RestartFU (commented):
since we decided we were not adding the vuln check, can you strip it and only remove the build step, since that's already checked by golangci govet

RestartFU (approved):

didntpot (approved):

PR #1231, opened 2026-04-30
server/player: Implement cross-world transfer support

hydzilla:
Added `Player.TransferToWorld` and `Tx.TransferEntity` for moving entities between worlds, and a cancellable `HandleWorldTransfer` event. This will be required to implement portal functionality.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -2183,6 +2183,10 @@ func (p *Player) TransferToWorld(w *world.World, pos mgl64.Vec3) *world.EntityHa
 	p.ResetFallDistance()
 
 	handle := p.tx.RemoveEntity(p)
+	if handle == nil {
+		// The player was already removed from its world (e.g. a previous transfer this tick).
+		return p.handle
+	}
 	w.Exec(func(tx *world.Tx) {
 		np := tx.AddEntity(handle).(*Player)
 		np.Teleport(pos)
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -2161,6 +2161,37 @@ func (p *Player) PickBlock(pos cube.Pos) {
 	p.SetHeldItems(pickedItem, offhand)
 }
 
+// TransferToWorld transfers the player to a different world, teleporting them to the position
+// passed. If the destination world is the same as the player's current one, the call behaves
+// like Teleport. ctx.Cancel() may be used in HandleWorldTransfer to cancel the transfer.
+// Calling TransferToWorld may lead to the player being removed from its world and being added
+// to a new world. This means that p cannot be assumed to be valid after a call to TransferToWorld.
+func (p *Player) TransferToWorld(w *world.World, pos mgl64.Vec3) *world.EntityHandle {
+	if w == p.tx.World() {
+		p.Teleport(pos)
+		return p.handle
+	}
+	ctx := event.C(p)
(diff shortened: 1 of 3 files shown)


HashimTheArab:
It might be better to include this with the nether portals PR since they might influence this design

server/player/player.go
```diff
@@ -2161,6 +2161,41 @@ func (p *Player) PickBlock(pos cube.Pos) {
 	p.SetHeldItems(pickedItem, offhand)
 }
 
+// TransferToWorld transfers the player to a different world, teleporting them to the position
+// passed. If the destination world is the same as the player's current one, the call behaves
+// like Teleport. ctx.Cancel() may be used in HandleWorldTransfer to cancel the transfer.
+// Calling TransferToWorld may lead to the player being removed from its world and being added
+// to a new world. This means that p cannot be assumed to be valid after a call to TransferToWorld.
+func (p *Player) TransferToWorld(w *world.World, pos mgl64.Vec3) *world.EntityHandle {
```
AkmalFairuz, line 2169: I prefer to use the target world's transaction rather than the world itself, so the caller must be inside the target world's transaction. I don't see why you're returning `p.handle`.

server/world/tx.go
```diff
@@ -206,6 +206,22 @@ func (tx *Tx) RemoveEntity(e Entity) *EntityHandle {
 	return tx.World().removeEntity(e, tx)
 }
 
+// TransferEntity removes the Entity from the World tied to this Tx and re-adds
+// it to the World passed at the position passed. The destination World may be
+// the same as the current one, in which case this acts like a teleport
+// scheduled via World.Exec. After transferring the Entity from the World, the
+// Entity is no longer usable.
+func (tx *Tx) TransferEntity(e Entity, w *World, pos mgl64.Vec3) {
```
AkmalFairuz, line 214: This too, I prefer the target world's transaction.

server/player/player.go
```diff
@@ -2161,6 +2161,41 @@ func (p *Player) PickBlock(pos cube.Pos) {
...
+	}
+
+	p.ResetFallDistance()
+
+	handle := p.tx.RemoveEntity(p)
+	if handle == nil {
+		// The player was already removed from its world (e.g. a previous transfer this tick).
+		return p.handle
+	}
+	w.Exec(func(tx *world.Tx) {
```
AkmalFairuz, line 2190: Based on my review before, I prefer using the target world's transaction from the caller. This could cause an inconsistent state, since this code is not guaranteed to execute.

There are two good options. Choose one:
- Use the target world's transaction rather than the world itself, or
- Use `<-w.Exec()`.

server/world/tx.go
```diff
@@ -206,6 +206,22 @@ func (tx *Tx) RemoveEntity(e Entity) *EntityHandle {
...
+	handle := tx.RemoveEntity(e)
+	if handle == nil {
+		return
+	}
+	handle.data.Pos = pos
+	w.Exec(func(tx *Tx) {
```
AkmalFairuz, line 220: Same as before, this would cause an inconsistent state.

AkmalFairuz (changes requested):

cjmustard:
portals should be implementing this logic instead of having a standalone

PR #1237, opened 2026-05-04
add DefaultSpawn() to generators and fix incorrect superflat spawn

HashimTheArab:


NopeNotDark (approved):

server/world/generator/flat.go
```diff
@@ -35,15 +33,21 @@ func NewFlat(biome world.Biome, layers []world.Block) Flat {
...
+	n := int16(len(f.layers))
 
-	for x := uint8(0); x < 16; x++ {
-		for z := uint8(0); z < 16; z++ {
-			for y := int16(0); y <= max; y++ {
-				if y < f.n {
-					chunk.SetBlock(x, min+y, z, 0, f.layers[f.n-y-1])
+	for x := range uint8(16) {
+		for z := range uint8(16) {
+			for y := range int16(max) {
```
cjmustard: This changes behavior from the old logic. `range int16(max)` only iterates through `max - 1`, while the previous `y <= max` loop included `max`. Should this be `range max + 1` to preserve the old behavior?

cjmustard (changes requested):

PR #1240, opened 2026-05-06
fix incorrect player prone eye height

HashimTheArab:


RestartFU:
where'd you get that from

RestartFU (dismissed):

cjmustard:
I feel like this is a pretty large change and may mess up some dependencies' legacy code. Maybe we should wait to merge this on a larger update/tag?

server/player/player.go
```diff
@@ -2902,12 +2902,13 @@ func (p *Player) OnGround() bool {
...
+// EyeHeight returns the player's eye offset above its feet, in blocks. The value depends on the
+// player's current pose: 1.62 when standing, 1.26 while sneaking, and 0.4 while swimming, gliding
+// or crawling.
 func (p *Player) EyeHeight() float64 {
 	switch {
 	case p.swimming || p.crawling || p.gliding:
-		return 0.52
+		return 0.4
 	case p.sneaking:
 		return 1.26
```
TwistedAsylumMC: Did you want to change sneaking to 1.27 also? Source: https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/Avatar#L34. Can also confirm this is the same as bedrock

TwistedAsylumMC (changes requested):

TwistedAsylumMC:
> I feel like this is a pretty large change and may mess up some dependencies' legacy code. Maybe we should wait to merge this on a larger update/tag?
How much depends on eye height specifically? This is not touching actual bounding boxes and shouldn't have too much of an impact fixing to match vanilla?

TwistedAsylumMC (approved):

cqdetdev (approved):

PR #1241, opened 2026-05-06
Fix immune-period player hits applying second knockback

HashimTheArab:
## Summary

Fixes player attacks during damage immunity so stronger hits only apply the vanilla damage delta without behaving like a fresh hit.

Previously, Dragonfly correctly ignored equal/lower immune-period damage and applied only the delta for higher damage, but still reset attack immunity and applied hurt animation/knockback. This made high-CPS combat feel like a single click registered as two hits.

This now matches PMMP/vanilla behavior more closely:
- higher damage during immunity may still deal only the damage difference
- immunity is not reset by that delta hit
- hurt animation and knockback only apply to cold hits


RestartFU (changes requested):
I believe this behavior is intentional, and it's fixable through the handlers

PR #1247, opened 2026-05-09
Implement Respawn Anchors

cqdetdev:
This PR implements respawn anchors. End crystal support was split into https://github.com/df-mc/dragonfly/pull/1344 for independent review.

- Adds respawn anchor block states for 0-4 charges, glowstone charging, light levels, Nether spawn setting, safe-position resolution, and charge consumption after successful respawn insertion.
- Creates vanilla power-5 incendiary explosions when anchors are used outside the Nether, including the six-face water suppression behavior.
- Persists and sends the `respawnblocksexplode` world rule, applying it to both respawn anchors and beds.
- Persists current dimension-aware spawn fields while retaining compatibility with older Dragonfly and vanilla player data.
- Updates `world.Provider` to load and save `world.PlayerSpawn`, including its dimension.
- Supports Nether anchors when returning from the End without consuming a charge.
- Resolves respawn positions safely around handlers, including cross-dimensional fallback and handler-mutated worlds or blocks.

`ExplosionSize` was removed because placed blocks retain only encoded block-state properties, so a per-block Go field would silently reset. The vanilla power of 5 is used directly instead.

Provider implementations must replace `LoadPlayerSpawnPosition`/`SavePlayerSpawnPosition` with `LoadPlayerSpawn`/`SavePlayerSpawn`.

Verification:
- `go test ./...`
- `make lint`
- `git diff --check`


server/entity/damageable.go
HashimTheArab, line 1: this was deduplicated out of player

server/block/explosion.go
```diff
@@ -20,6 +20,12 @@ import (
 type ExplosionConfig struct {
 	// Size is the size of the explosion, it is effectively the radius which entities/blocks will be affected within.
 	Size float64
+	// EndCrystal makes the explosion behave like an End crystal explosion, preventing block damage below the
+	// explosion origin's Y level if the crystal is on obsidian or bedrock.
+	EndCrystal bool
```
FDUTCH: not how it works, blocks below end crystal does not break because of obsidian (or bedrock) and explosion position is just above that block

server/block/explosion.go
```diff
@@ -20,6 +20,12 @@ import (
...
+	// DisableEntityDamage disables entity impact from the explosion while still applying block effects, sound and
+	// particles.
+	DisableEntityDamage bool
```
FDUTCH: as I know, if explosion position is inside water block, it doesn't dial any damage to blocks, but it does dial 0 damage to entities around (similar to an explosion blocked by blocks, or damage from snowball)

server/block/explosion.go
```diff
@@ -219,9 +233,23 @@ func exposure(tx *world.Tx, origin mgl64.Vec3, e world.Entity) float64 {
...
+		return 0
+	}
 	return misses / checks
 }
 
+// endCrystalSupported checks if the End crystal explosion has a supporting block
+// that protects blocks below it from being destroyed.
+func endCrystalSupported(tx *world.Tx, explosionPos mgl64.Vec3) bool {
+	switch tx.Block(cube.PosFromVec3(explosionPos).Side(cube.FaceDown)).(type) {
+	case Obsidian, Bedrock:
```
FDUTCH: I'd make it an interface

server/entity/end_crystal_behaviour.go
```diff
@@ -0,0 +1,93 @@
...
+		if _, ok := tx.Block(pos).(block.Fire); !ok {
+			flame := block.Fire{}
+			tx.SetBlock(pos, flame, nil)
+			tx.ScheduleBlockUpdate(pos, flame, time.Duration(30+rand.IntN(10))*time.Second/20)
+		}
+	}
+	return nil
+}
+
+func (endCrystalBehaviour) Explode(e *Ent, _ mgl64.Vec3, impact float64, _ block.ExplosionConfig) {
```
FDUTCH: does not contain necessary logic (tip: сlose the entity before the explosion)

server/block/explosion.go
HashimTheArab: I considered this but I'm not sure if its worth it because crystals are explicitly only placable on obsidian and bedrock and it's unlikely to change, but I realize this doesn't check for crying obsidian

server/entity/damageable.go
HashimTheArab: Kept this deduplicated at the entity layer: `HurtEntity` and `DamageableEntity` are the shared path used by player attacks and projectiles, so the player does not regain a separate damageability implementation.

server/block/explosion.go
HashimTheArab: The obsolete crystal-specific block clipping was removed. The crystal now explodes from its entity position above the support block, while normal blast resistance protects obsidian/bedrock.

server/block/explosion.go
HashimTheArab: Addressed in `62031e8a`: End-crystal explosions opt into liquid-blocked entity exposure and skip zero-impact dispatch, so entities behind water are unaffected while legacy generic explosion behavior is unchanged.

server/block/explosion.go
HashimTheArab: This code is outdated and the special-case support clipping it referred to is gone. Crystal placement retains the vanilla obsidian/bedrock allowlist; standard explosion resistance handles the supporting block.

server/entity/end_crystal_behaviour.go
HashimTheArab: Addressed: `explodeEndCrystal` closes the entity before invoking the explosion, preventing the crystal from being included in or retriggered by its own blast.

server/entity/end_crystal_behaviour.go
```diff
@@ -0,0 +1,81 @@
...
+
+// explodeEndCrystal closes the End crystal and creates a radius-6,
+// non-incendiary explosion at its base, if the crystal was not closed yet.
+func explodeEndCrystal(e *Ent) {
+	if _, ok := e.H().Entity(e.tx); !ok {
+		return
+	}
+	pos := e.Position()
+	_ = e.Close()
+	block.ExplosionConfig{Size: 6, SuppressUnderwaterImpact: true}.Explode(e.tx, pos)
```
FDUTCH: would be really nice if we could configure explosion size via config (and maybe implement same thing for tnt)

HashimTheArab: this should be a separate PR

HashimTheArab:
this pr is ready

PR #1249, opened 2026-05-11
feat(world/view_layer.go): implement block view layer

RestartFU:
## Summary

- add world-scoped, transaction-owned block overrides to view layers
- apply overrides copy-on-write to chunk/sub-chunk delivery, including cache-aware height advertisements and block entities
- keep public effects scoped to viewers of the affected block while private effects remain session-only
- preserve public/private break mode through start, continue, and finish, and correct Bedrock's adjacent placement prediction on rejected private interactions
- restore public solid and liquid layers when an override is removed

## API changes

- `player.Handler.HandleBlockBreak` now receives a `private bool`; private breaks expose empty drops/XP and ignore mutations to them
- `session.Controllable` adds `BreakVisibleBlock` so network-driven breaks target the block the client sees while direct `Player.BreakBlock` retains public-world semantics
- block override mutations require the active `*world.Tx`; entity-only `ViewLayerUpdater` implementations remain compatible through the optional `ViewLayerBlockUpdater`

## Verification

- `make test`
- `make lint`
- `go test -race ./server/player -count=1`
- `codex review --base upstream/master` (clean)


cqdetdev (dismissed):

cqdetdev (dismissed):

cqdetdev (dismissed):
I've tested this in game, and it LGTM

PR #1259, opened 2026-05-20
feat(server/world): Implement armor visiblity to per-viewer view layers

cqdetdev:
A caveat to https://github.com/df-mc/dragonfly/commit/b9a408ac6b28efed82a9a051d983cfbb827ae60d is that armor is not hidden when a player goes out of a view. This means that a player could still be visible if they are wearing armor which may be an unintended side effect.

This PR implements the ability to manually choose to display armor as visible.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -2612,12 +2612,17 @@ func (p *Player) ViewVisibility(entity world.Entity, level world.VisibilityLevel
 	p.session().ViewVisibility(entity, level)
 }
 
-// ViewArmour overrides whether the public armour of the entity is visible for this player.
-func (p *Player) ViewArmour(entity world.Entity, visible bool) {
-	p.session().ViewArmour(entity, visible)
+// ViewArmour overrides the public armour of the entity for this player.
+func (p *Player) ViewArmour(entity world.Entity, helmet, chestplate, leggings, boots item.Stack) {
+	p.session().ViewArmour(entity, helmet, chestplate, leggings, boots)
 }
 
-// ViewPublicArmour removes the armour visibility override of the entity for this player.
+// ViewNoArmour hides the armour of the entity for this player.
+func (p *Player) ViewNoArmour(entity world.Entity) {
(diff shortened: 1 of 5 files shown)


RestartFU:
I think we should also be able to view specific armour, not just be either hidden or visible

cqdetdev:
Good idea, i'm just not sure the use case for individual pieces but I can add it. Just need to make it clean

RestartFU:
> Good idea, i'm just not sure the use case for individual pieces but I can add it. Just need to make it clean

well I don't think it should be based on visibility, I think we should be able to set specific armour pieces, and if we want no armour, we can just have 4 empty stacks (or an empty inventory)

cqdetdev:
That's what I meant. What I mean is like if a user wants a player invisible in the first place 99% of the time they'd want their armor invisible too and not probably only enable helmet but I guess you implemented that in one of your things so IG there is a usecase

cqdetdev:
@RestartFU Have a look

RestartFU:
I like that better, I'll take a deeper look tonight

cqdetdev:
@RestartFU review

HashimTheArab:
```
 Found one actionable issue:

  - [P2] Armor overrides leak after entities despawn — server/session/view_layer.go:65

    viewLayerArmour stores entries separately from world.ViewLayer. Existing despawn cleanup only calls ViewLayer.remove, so it
    never deletes the corresponding armor entry. The player-disconnect path has the same problem: sessionList.Remove calls
    other.viewLayer.Remove(entity) directly.

    Each overridden entity therefore leaves its EntityHandle and four item.Stacks retained until the viewing session closes. A
    long-running server that applies overrides to temporary NPCs/entities will grow this map indefinitely.

    Armor overrides should participate in the existing ViewLayer lifecycle, or both non-player despawn and player-session removal
    must explicitly clean viewLayerArmour without sending a refresh for the removed entity.
    ```

PR #1260, opened 2026-05-20
perf(server/world): Track dirty column sections to avoid redundant world saves

cqdetdev:
Dragonfly currently tracks modifications to a loaded column but always treats it as modified until it is saved. This means repeated autosaves can occur redundantly even if a proper save already occurred. 

This PR adds dirty-section tracking to chunk columns so the world and provider can distinguish clean columns and dirty ones. This should reduce overhead for worlds to not commit to a save if we already know the chunk column is clean.

 ## Benchmarks

  Local benchmark results on Windows amd64, Ryzen 7 7730U.

  ### Clean save path

  Before:
  - direct clean `saveChunk`: ~617.6 ns/op, 96 B/op, 1 alloc/op
  - public `World.Save()` with LevelDB: ~163.5 us/op, ~20 KB/op, 131 allocs/op

  After:
  - direct clean `saveChunk`: ~6.49 ns/op, 0 B/op, 0 allocs/op
  - public `World.Save()` with LevelDB: ~3.32 us/op, 184 B/op, 5 allocs/op

  ### Provider dirty-signal path

  Before:
  - unchanged immediate store: ~11.2-13.2 us/op
  - unchanged safe store: ~6.0-6.4 us/op
  - unchanged dense safe store: ~35.1-36.7 us/op

  After:
  - unchanged immediate store: ~132.9 ns/op, 0 allocs
  - unchanged safe store: ~138.1 ns/op, 0 allocs
  - unchanged dense safe store: ~149.1 ns/op, 0 allocs

  Normal write performance stayed in the same range:
  - immediate store: ~24.7 us/op
  - dense store: ~10.2 us/op
  - LevelDB store in same targeted pass: ~88.5 us/op****

### DIFF
--- a/server/world/chunk/chunk.go
+++ b/server/world/chunk/chunk.go
@@ -26,6 +26,9 @@ type Chunk struct {
 	sub []*SubChunk
 	// biomes is an array of biome IDs. There is one biome ID for every column in the chunk.
 	biomes []*PalettedStorage
+	// dirty holds which persisted chunk sections have changed since the chunk
+	// was last marked clean.
+	dirty DirtyFlags
 }
 
 // New initialises a new chunk and returns it, so that it may be used.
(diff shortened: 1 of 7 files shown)


cqdetdev:
@RestartFU @didntpot review pls

PR #1269, opened 2026-06-08
Read full level.dat data with io.ReadFull

TrippleAWap:
Replace a direct r.Read call with io.ReadFull when reading the level.dat data to ensure the full FileLength bytes are read (r.Read can return partial reads). Also adjust imports ordering to group standard library and external packages.

### DIFF
--- a/server/world/mcdb/leveldat/level_dat.go
+++ b/server/world/mcdb/leveldat/level_dat.go
@@ -4,9 +4,10 @@ import (
 	"bufio"
 	"encoding/binary"
 	"fmt"
-	"github.com/sandertv/gophertunnel/minecraft/nbt"
 	"io"
 	"os"
+
+	"github.com/sandertv/gophertunnel/minecraft/nbt"
 )
 
 // LevelDat implements the encoding and decoding of level.dat files. An empty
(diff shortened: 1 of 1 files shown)


RestartFU:
do you have an example .dat file where this is an issue?

TrippleAWap:
Yeah here, had to upload as a text file as GitHub doesn't support .dat files.
[level_dat.txt](https://github.com/user-attachments/files/28710960/level_dat.txt)

Silentmussle913:
I think it would be worth adding a `level_dat_test.go` for this change so reviewers can immediately see the issue being fixed and prevent regressions in the future.

Something along these lines:

```go
func TestReadPartialReader(t *testing.T) {
	var buf bytes.Buffer

	require.NoError(t, binary.Write(&buf, binary.LittleEndian, header{
		StorageVersion: 10,
		FileLength:     4,
	}))

	buf.Write([]byte{1, 2, 3, 4})

	ld, err := Read(&partialReader{data: buf.Bytes(), max: 1})

	require.NoError(t, err)
	require.Len(t, ld.data, 4)
	require.Equal(t, []byte{1, 2, 3, 4}, ld.data)
}
```

This would demonstrate why `io.ReadFull()` is needed, since a regular `Read()` is allowed to return fewer bytes than requested.
``THIS IS AN AI MSG LOL i was testing ai pr managing if u need to use it feel free but ignore this if not needed.``

TrippleAWap:
bump

TrippleAWap:
Can you merge this so I don't need to maintain a fork with the fix.

PR #1275, opened 2026-06-21
session/chunk.go: skip nil block entity NBT

RestartFU:
## Summary
- initialize nil block entity NBT maps before adding coordinates during subchunk encoding
- apply the same nil-map handling to cached and uncached chunk sends


Flonja:
Resolves #1100 as well then

cqdetdev (approved):
I'm OK with these changes, but defer to @TwistedAsylumMC on final say.

HashimTheArab:
This would significantly increase allocations

cqdetdev:
@HashimTheArab It'll only increase allocations for a small edge case, would make sense to risk a few extra allocations of an empty map than panic lol

PR #1279, opened 2026-06-23
world/chunk/decode.go: Fix panic on out-of-range sub chunk index in NetworkDecode

### LINKED ISSUE #1024: chunk.NetworkDecode and (*chunk.Chunk).SetBlock panic on certain data
# SetBlock Panic
```go
func (w *World) SetBlockByRuntimeID(pos cube.Pos, blockID uint32) {
	if cube.Pos(pos).OutOfBounds(cube.Range(w.Dimension().Range())) {
		return
	}

	chunkPos := protocol.ChunkPos{int32(pos[0]) >> 4, int32(pos[2]) >> 4}
	c := w.Chunk(chunkPos) // This returns a dragonfly *chunk.Chunk
	if c == nil {
		return
	}

	c.SetBlock(uint8(pos[0]), int16(pos[1]), uint8(pos[2]), 0, blockID) // This line panics
}
```
```go
case *packet.UpdateBlock:
	rid := pk.NewBlockRuntimeID
	_, ok := world.BlockByRuntimeID(rid)
	if !ok {
		rid = airRuntimeID
	}

	pos := cube.Pos{int(pk.Position.X()), int(pk.Position.Y()), int(pk.Position.Z())}
	p.player.World().SetBlockByRuntimeID(pos, rid)

```

github.com/df-mc/dragonfly/server/world/chunk in (*Chunk).SetBlock at line 88
runtime error: index out of range [-1]
https://github.com/df-mc/dragonfly/blob/1171b7e8d91e40388267e04014de2ad9f941c730/server/world/chunk/chunk.go#L88

# chunk.NetworkDecode panic

```go
case *packet.LevelChunk:
	dim := dimensionFromNetworkID(pk.Dimension)

	c, err := chunk.NetworkDecode(airRuntimeID, pk.RawPayload, int(pk.SubChunkCount), dim.Range())
	if err != nil {
	 	c = chunk.New(airRuntimeID, dim.Range())
	}
```

I have many logs of chunk data from these crashes, here's a log of *packet.LevelChunk 
chunk.NetworkDecode panic: runtime error: index out of range [109] with length 24
https://gist.github.com/HashimTheArab/1ebab2a39a22d8af009987d96b598a24

root-nat:
This fixes a denial-of-service panic when decoding chunk data received over the network.

A version 9 sub chunk encodes its own Y index in the network payload. That value is attacker-controlled and was translated into an index into `c.sub` without any bounds check, so a crafted `LevelChunk` packet (for example from a malicious proxy) could index `c.sub` out of range and crash the server with a runtime panic (`index out of range [109] with length 24`).

`NetworkDecodeBuffer` now captures the decoded sub chunk, validates the resulting index against `len(c.sub)` and returns an error instead of panicking, letting the caller fall back to an empty chunk.

This follows the fix suggested in the issue thread, with the off-by-one corrected (`>= len(c.sub)` rather than `> n`, so an index exactly equal to the length is also rejected).

Fixes #1024.

### DIFF
--- a/server/world/chunk/decode.go
+++ b/server/world/chunk/decode.go
@@ -23,16 +23,21 @@ func NetworkDecode(br BlockRegistry, data []byte, count int, r cube.Range) (*Chu
 // The sub chunk count passed must be that found in the LevelChunk packet.
 // noinspection GoUnusedExportedFunction
 func NetworkDecodeBuffer(br BlockRegistry, buf *bytes.Buffer, count int, r cube.Range) (*Chunk, error) {
-	var (
-		c   = New(br, r)
-		err error
-	)
+	c := New(br, r)
 	for i := 0; i < count; i++ {
 		index := uint8(i)
-		c.sub[index], err = decodeSubChunk(buf, c, &index, NetworkEncoding)
+		sub, err := decodeSubChunk(buf, c, &index, NetworkEncoding)
 		if err != nil {
 			return nil, err
 		}
+		// A version 9 sub chunk encodes its own Y index, which is read from the
+		// untrusted network data and translated to a sub chunk index. A malicious
+		// sender may craft an index outside the chunk's range, so guard against it
+		// before indexing c.sub to avoid an out of range panic.
+		if int(index) >= len(c.sub) {
+			return nil, fmt.Errorf("sub chunk index %v out of range (length %v)", index, len(c.sub))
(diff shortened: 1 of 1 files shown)


HashimTheArab (approved):

HashimTheArab:
This is no longer an issue

PR #1281, opened 2026-06-23
server/session: Resend inventories on item release to keep the client in sync

### LINKED ISSUE #1001: Canceling HandleItemRelease when using a Bow will decrease arrows in the player inventory (client-side)
Since the client predicts the arrow will be decreased when using the Bow, the server needs to resend the item to sync the inventory correctly

root-nat:
When a player releases an item such as a bow, the client predicts the result and, for example, decrements the arrow it expects to be consumed. If the server does not actually consume anything (the release is cancelled, the bow is not drawn long enough, or there are no arrows), it never corrected the client, leaving the displayed inventory out of sync with the server.

The inventories are now resent before handling the release in both release paths, mirroring what `handleUseItemTransaction` already does for item use. The slot updates from an actual consumption are still sent afterwards, so a successful shot ends up with the correct count either way.

Fixes #1001.

### DIFF
--- a/server/session/handler_inventory_transaction.go
+++ b/server/session/handler_inventory_transaction.go
@@ -45,7 +45,7 @@ func (h *InventoryTransactionHandler) Handle(p packet.Packet, s *Session, tx *wo
 
 	switch data := pk.TransactionData.(type) {
 	case *protocol.NormalTransactionData:
-		h.resendInventories(s)
+		s.resendInventories()
 		// Always resend inventories with normal transactions. Most of the time we do not use these
 		// transactions, so we're best off making sure the client and server stay in sync.
 		if err := h.handleNormalTransaction(pk, s, c); err != nil {
@@ -54,7 +54,7 @@ func (h *InventoryTransactionHandler) Handle(p packet.Packet, s *Session, tx *wo
 		return
 	case *protocol.MismatchTransactionData:
 		// Just resend the inventory and don't do anything.
-		h.resendInventories(s)
+		s.resendInventories()
 		return
 	case *protocol.UseItemOnEntityTransactionData:
 		if err = s.VerifyAndSetHeldSlot(int(data.HotBarSlot), stackToItem(s.br, data.HeldItem.Stack), c); err != nil {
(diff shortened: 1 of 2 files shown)


server/session/handler_inventory_transaction.go
```diff
@@ -199,7 +199,11 @@ func (h *InventoryTransactionHandler) handleUseItemTransaction(data *protocol.Us
 }
 
 // handleReleaseItemTransaction ...
-func (h *InventoryTransactionHandler) handleReleaseItemTransaction(c Controllable) error {
+func (h *InventoryTransactionHandler) handleReleaseItemTransaction(s *Session, c Controllable) error {
+	// The client predicts the result of releasing an item, such as an arrow being removed from the
```
HashimTheArab, line 203: this comment could be 1 line

server/session/handler_player_action.go
```diff
@@ -47,6 +47,11 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
 	case protocol.PlayerActionStartItemUseOn:
 		// TODO: Properly utilize these actions.
 	case protocol.PlayerActionStopItemUseOn:
+		// The client predicts the result of releasing an item, such as an arrow being removed from
```
HashimTheArab, line 50: this comment could be 1 line

server/session/handler_inventory_transaction.go
```diff
@@ -199,7 +199,11 @@ func (h *InventoryTransactionHandler) handleUseItemTransaction(data *protocol.Us
...
+	// inventory when a bow is fired. Resend the inventories so that the client stays in sync if the
+	// release does not actually consume the predicted items, for example when it is cancelled.
+	s.resendInventories()
```
FDUTCH, line 206: wrong logic, It won't fix anything, it will only make things worse...

server/session/handler_player_action.go
```diff
@@ -47,6 +47,11 @@ func handlePlayerAction(action int32, face int32, pos protocol.BlockPos, entityR
...
+		// the inventory when a bow is fired. Resend the inventories so that the client stays in sync
+		// if the release does not actually consume the predicted items, for example when it is
+		// cancelled.
+		s.resendInventories()
```
FDUTCH, line 54: wrong logic...

root-nat:
You're right — resending before the release was the wrong placement (it re-sends the pre-release inventory on every successful shot and relies on the slot update to re-correct it, causing a redundant resend and a one-frame flicker). Reworked in #1289: it now resyncs only when the release consumes nothing, and leaves successful shots to the existing inventory slot-change callback (`broadcastInvFunc`).

PR #1282, opened 2026-06-23
server/block: Implement the snow layer block

### LINKED ISSUE #871: feature: Snow Layer block missing
The snow layer block is currently not implemented in dragonfly

root-nat:
Implements the snow layer block (`minecraft:snow_layer`), a thin partial covering of snow that can be stacked up to eight layers high.

- Scales its collision box with the number of layers, with a single layer being passable, matching vanilla.
- Can be placed on top of any block with a solid upward face, and stacks an extra layer when more snow is used on it. At eight layers it stays a snow layer rather than becoming a full snow block.
- Breaks for one to four snowballs depending on the number of layers, or the layers themselves with Silk Touch (a full eight-layer block drops a snow block).
- Melts a layer at a time in biomes that are too warm to sustain snow.

The block follows the existing patterns of similar blocks (`Cake`, `SeaPickle`, `Carpet`). Only the `covered_bit=false` states are registered; snowlogging (`covered_bit`), gravity, and melting from a nearby block light source are noted as follow-ups (the last because the world does not currently expose the block light level separately from the sky light).

Fixes #871.


Silentmussle913:
gl but this aint gettin accepted 🥀

RestartFU:
> gl but this aint gettin accepted 🥀 

tbh it doesn't look bad at all

I'll test the PR out later today or tomorrow

PR #1283, opened 2026-06-23
player/player.go: Only deal fall damage on an actual vertical collision

### LINKED ISSUE #1003: Fall damage when getting close to blocks while falling
https://github.com/user-attachments/assets/cfefa11c-3c07-4e81-8451-9fc51bb4e4df

root-nat:
### Problem

When a player falls close to a block at its side it takes premature fall damage (#1003). The cause is `checkOnGround`: it extended the probe box opposite to the whole movement (`deltaPos.Mul(-1.0)`), which while falling pushes the box upwards and sideways, so a block next to the fall path intersects it and the player is wrongly reported on the ground.

### Fix

Extend the probe straight **down** only — by 0.05 plus the tick's descent — keeping the player's exact horizontal footprint:

```go
.Extend(mgl64.Vec3{0, -0.05}).Extend(mgl64.Vec3{0, math.Min(deltaPos[1], 0)})
```

A block at the side is no longer mistaken for ground, while a stepped or fast descent still finds the ground beneath the footprint. `updateFallState` is left untouched. This is the root-cause fix suggested in the previous review, rather than patching the fall-damage path.

### Testing

Verified in-game on 1.26.30:
- sprinting and walking down a 15-step staircase deals no fall damage (#1044 preserved);
- a ~15-block open fall deals the expected ~6 hearts;
- falling flush alongside a wall deals the same damage, and only on landing (#1003 fixed).

Fixes #1003.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -538,7 +538,9 @@ func (p *Player) Heal(health float64, source world.HealingSource) {
 // updateFallState is called to update the entities falling state.
 func (p *Player) updateFallState(distanceThisTick float64) {
 	switch {
-	case p.OnGround():
+	// Require a vertical collision so brushing a block sideways mid-fall does
+	// not register as landing and deal premature fall damage.
+	case p.OnGround() && p.collidedVertically:
 		if p.fallDistance > 0 {
 			p.fall(p.fallDistance)
 			p.ResetFallDistance()


TwistedAsylumMC:
We should fix the on ground being set to true when falling past blocks rather than this

root-nat:
Thanks for the steer — agreed, patching `updateFallState` was treating the symptom.

I've reworked it to fix `checkOnGround` directly. The probe box was extended opposite to the whole movement (`deltaPos.Mul(-1.0)`), which while falling pushes it upwards and sideways, so a block next to the fall path intersects it and the player reports as grounded. The rework extends the probe straight **down** only — by 0.05 plus the tick's descent — keeping the player's exact horizontal footprint:

```go
.Extend(mgl64.Vec3{0, -0.05}).Extend(mgl64.Vec3{0, math.Min(deltaPos[1], 0)})
```

A block at the side is no longer mistaken for ground, while a stepped or fast descent still finds the ground beneath the footprint, so the sprinting-down-stairs fix from #1044 should hold (the 0.6-wide footprint always straddles the step it is leaving, and the descent term keeps the downward reach in step with the gap). The idle-tick call, which passes a zero delta, is unchanged.

Does this match what you had in mind? If so I'll reopen with it — I'll give the stairs/slabs case a run in-game first either way.

root-nat:
Continued in #1286 with the root-cause `checkOnGround` rework (this PR could not be reopened after the branch was force-pushed).

PR #1286, opened 2026-06-25
player/player.go: Stop treating blocks passed while falling as ground

### LINKED ISSUE #1003: Fall damage when getting close to blocks while falling

root-nat:
Reworked from #1283 per @TwistedAsylumMC's review: this fixes the root cause in `checkOnGround` rather than patching the fall-damage path. (The original PR could not be reopened because the branch was force-pushed while it was closed.)

### Problem

When a player falls close to a block at its side it takes premature fall damage (#1003). The cause is `checkOnGround`: it extended the probe box opposite to the whole movement (`deltaPos.Mul(-1.0)`), which while falling pushes the box upwards and sideways, so a block next to the fall path intersects it and the player is wrongly reported on the ground.

### Fix

Extend the probe straight **down** only — by 0.05 plus the tick's descent — keeping the player's exact horizontal footprint:

```go
.Extend(mgl64.Vec3{0, -0.05}).Extend(mgl64.Vec3{0, math.Min(deltaPos[1], 0)})
```

A block at the side is no longer mistaken for ground, while a stepped or fast descent still finds the ground beneath the footprint. `updateFallState` is left untouched.

### Testing

Verified in-game on 1.26.30:
- sprinting and walking down a 15-step staircase deals no fall damage (#1044 preserved);
- a ~15-block open fall deals the expected ~6 hearts;
- falling flush alongside a wall deals the same damage, and only on landing (#1003 fixed).

Fixes #1003.

### DIFF
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -2866,7 +2866,9 @@ func (p *Player) checkEntitySteppers() {
 
 // checkOnGround checks if the player is currently considered to be on the ground.
 func (p *Player) checkOnGround(deltaPos mgl64.Vec3) bool {
-	box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(deltaPos.Mul(-1.0))
+	// Probe straight down (0.05 plus this tick's descent) instead of along the
+	// whole movement, so a block the player falls past at its side is not ground.
+	box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(mgl64.Vec3{0, math.Min(deltaPos[1], 0)})
 	b := box.Grow(1)
 
 	epsilon := mgl64.Vec3{mgl64.Epsilon, mgl64.Epsilon, mgl64.Epsilon}


server/player/player.go
```diff
@@ -2866,7 +2866,9 @@ func (p *Player) checkEntitySteppers() {
 
 // checkOnGround checks if the player is currently considered to be on the ground.
 func (p *Player) checkOnGround(deltaPos mgl64.Vec3) bool {
-	box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(deltaPos.Mul(-1.0))
+	// Probe straight down (0.05 plus this tick's descent) instead of along the
+	// whole movement, so a block the player falls past at its side is not ground.
+	box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(mgl64.Vec3{0, math.Min(deltaPos[1], 0)})
```
FDUTCH, line 2871: missunderstanding the cause... (won't work) 👏

root-nat: Thanks for taking a look — could you share a concrete failing scenario? I'd like to make sure I'm not missing something.

The change keeps the player's exact X/Z footprint and extends the probe box straight **down** by 0.05 plus this tick's descent (`math.Min(deltaPos[1], 0)`), instead of the old `.Extend(deltaPos.Mul(-1.0))`. While falling, `deltaPos[1]` is negative, so the old `-deltaPos` extended the box *upward* and *sideways* (opposite the movement) — that is what made it intersect a wall the player was sliding past and report `onGround` every tick, continuously resetting `fallDistance` (the #1003 bug). The new box only grows downward under the feet, so a block beside the fall path is no longer treated as ground while ground actually under the feet still is.

I tested this in-game on 1.26.30: sprinting/walking down stairs deals no fall damage (#1044 stays fixed), a ~15-block open fall deals ~6 hearts, and falling flush against a wall now deals damage only on landing (#1003 fixed). I also checked the edge cases (standing, ledges, jumping where `math.Min` clamps to 0, and fast falls where `checkBlockCollisions` has already clamped the position onto the floor).

If your point is that `onGround` should ideally come from the move's vertical-collision result (like vanilla) rather than a probe box — I agree in principle, but `checkBlockCollisions` only runs when `deltaPos.Len() <= 3` and `collidedVertically` is also set on ceiling hits, so switching to it is a riskier refactor I'd rather keep separate from this fix. Happy to dig into any specific repro you have.

SpaceLostC: Absolutely! Here's an example scenario of this happening: https://www.youtube.com/watch?v=IlZ-Lclu54E

PR #1290, opened 2026-06-27
Implement first-class Nethernet support

cqdetdev:
## Summary

Adds opt-in NetherNet support next to the existing RakNet listener.

- Updates gophertunnel and adds go-nethernet support
- RakNet stays the default: `Network.Address` keeps its meaning and existing configs need no changes
- `Network.Transport` lists the built-in transports to run, e.g. `["raknet", "nethernet"]`
- `[Network.NetherNet]` configures the plaintext HTTP signaling endpoint (shares `Network.Address` unless its own `Address` is set), the persistent P-384 identity key, the advertised identity-provider domain and the UDP port or range used for player connections
- Exposes `NetherNetConfig` for embedders that need custom signaling server settings or ICE credentials, and `ListenNetwork` for custom `minecraft.Network` transports through `Config.Listeners`

## Notes

The built-in signaling endpoint serves plaintext HTTP. Operators that require HTTPS should terminate TLS with a reverse proxy.

The NetherNet UDP mux defaults to port 19133 because it cannot share UDP 19132 with RakNet on the same host.


HashimTheArab:
Maybe the tls cert should be left to the reverse proxy instead of having dragonfly handle it, most people will probably have that setup anyways

HashimTheArab:
Mostly LGTM tho i dont understand the logistics of "both" and how the client handles it

lactyy2:
> Mostly LGTM tho i dont understand the logistics of "both" and how the client handles it

i think there could be use cases where you'd want to listen on port 19132 for RakNet and port 80 for NetherNet?
The client likely prefers RakNet if both are on the same port, so maybe that's worth documenting in config.toml

lactyy2 (approved):

PR #1292, opened 2026-06-28
Implement bamboo building set

MEMOxiiii:
## Summary

This PR implements the bamboo building block family using Dragonfly's existing wood, stairs, slab, door, fence, trapdoor, and sign systems.

## Added blocks and items

- `bamboo_planks`
- `bamboo_stairs`
- `bamboo_slab`
- `bamboo_fence`
- `bamboo_fence_gate`
- `bamboo_door`
- `bamboo_trapdoor`
- `bamboo_sign`
- `bamboo_block`
- `stripped_bamboo_block`
- `bamboo_mosaic`
- `bamboo_mosaic_stairs`
- `bamboo_mosaic_slab`

## Changes

- Add bamboo as a wood type.
- Add bamboo block and stripped bamboo block with axis support.
- Support stripping bamboo blocks.
- Add bamboo mosaic and its stairs and slab variants.
- Register the complete bamboo building family.
- Avoid registering bamboo as a log or regular wood block.
- Add the required block hashes.


server/block/register.go
```diff
@@ -392,6 +394,9 @@ func init() {
 	world.RegisterItem(Stone{})
 	world.RegisterItem(String{})
 	world.RegisterItem(SugarCane{})
+	world.RegisterItem(BambooBlock{})
+	world.RegisterItem(BambooBlock{Stripped: true})
+	world.RegisterItem(BambooMosaic{})
```
HashimTheArab: Insert these after Andesite to be consistent with the rest of the list being in alphabetical order. Besides this everything looks good!

HashimTheArab (commented):
Looks good except 1 comment

server/block/register.go
MEMOxiiii: Done :)

server/block/wood.go
```diff
@@ -101,6 +101,9 @@ func (w Wood) EncodeBlock() (name string, properties map[string]any) {
 // allWood returns a list of all possible wood states.
 func allWood() (wood []world.Block) {
```
RestartFU, line 103: maybe we should name this something else? because it doesn't include all kinds of wood anymore

HashimTheArab: A comment is fine its an internal method. "of all possible wood states excluding bamboo"
Why does it skip bamboo tho?

MEMOxiiii: Bamboo is skipped here because it doesn't have bamboo_wood variants

 It uses bamboo_block and stripped_bamboo_block instead , which are registered separately in allBambooBlocks

HashimTheArab: got it, yea just update the comment and we can merge this

HashimTheArab (dismissed):

server/block/wood.go
MEMOxiiii: Done

server/block/bamboo_block.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	b.Axis = face.Axis()
+
+	place(tx, pos, b, user, ctx)
+	return placed(ctx)
+}
+
+// Strip ...
+func (b BambooBlock) Strip() (world.Block, world.Sound, bool) {
+	return BambooBlock{Axis: b.Axis, Stripped: true}, nil, !b.Stripped
```
RestartFU, line 51: if stripped is always true, and then strip is triggered again, it's gonna return false, is that intended?

that would mean that the BambooBlock would have Stripped: true, while the Strip() call would return false

```suggestion
func (b BambooBlock) Strip() (world.Block, world.Sound, bool) {
	return BambooBlock{Axis: b.Axis, Stripped: true}, nil, true
```

server/block/bamboo_block.go
```diff
@@ -0,0 +1,77 @@
...
+	return "minecraft:bamboo_block", 0
+}
+
+// EncodeBlock ...
+func (b BambooBlock) EncodeBlock() (name string, properties map[string]any) {
+	if b.Stripped {
+		return "minecraft:stripped_bamboo_block", map[string]any{"pillar_axis": b.Axis.String()}
+	}
+	return "minecraft:bamboo_block", map[string]any{"pillar_axis": b.Axis.String()}
+}
```
RestartFU, line 69:
```suggestion
func (b BambooBlock) EncodeBlock() (name string, properties map[string]any) {
    meta := map[string]any{"pillar_axis": b.Axis.String()}
	if b.Stripped {
		return "minecraft:stripped_bamboo_block", meta
	}
	return "minecraft:bamboo_block", meta
}
```

server/block/bamboo_block.go
MEMOxiiii: I tested both behaviours in game. 
Returning true causes the interaction to succeed again on an already stripped bamboo block and damages the axe even though the block doesn't change. 
With !b.Stripped, it correctly returns false after the first strip, which also matches Log.Strip and Wood.Strip. 

so the current behaviour makes more sense.

RestartFU (approved):

HashimTheArab (approved):

MEMOxiiii:
@HashimTheArab @RestartFU Thx u very much

HashimTheArab:
Thanks for the PR! We're looking into https://github.com/df-mc/dragonfly/pull/1291 to try to get that merged first and will merge this immediately after

MEMOxiiii:
> Thanks for the PR! We're looking into #1291 to try to get that merged first and will merge this immediately after

OK

PR #1294, opened 2026-07-01
block/break_info.go: Account for status effects and environment in block break duration

RestartFU:
## Summary

`block.BreakDuration` only implemented the base hardness/tool/efficiency terms and left a literal `// TODO: Account for haste etc here.`. `Player.breakTime` worked around this by scaling the **already-rounded** `BreakDuration` result by the airborne/underwater penalties and by per-effect multipliers. That approach has two problems:

1. **Rounding happens before the penalties are applied.** Because `BreakDuration` rounds to a whole tick before returning, multiplying afterwards produces off-by-one break times (e.g. diamond pickaxe on stone while airborne is 29 ticks in-game, but scaling the rounded 6-tick result by 5 gives 30).
2. **The per-effect multipliers don't match Bedrock.** `effect.Haste.Multiplier`/`MiningFatigue.Multiplier` are rough single-factor approximations rather than the Bedrock breaking formula.

This PR implements the Bedrock Edition breaking calculation properly and routes `Player.breakTime` through it.

## Changes

- Add `block.BreakContext` describing the effect/environment state (Haste, Conduit Power, Mining Fatigue levels; underwater, aqua affinity, airborne).
- Add `block.BreakDurationContext(b, i, ctx)` implementing the [Bedrock Edition calculation](https://minecraft.wiki/w/Breaking#Calculation):
  - tool speed applies only to harvestable blocks (a correct-type but wrong-tier tool gets no bonus),
  - Haste/Conduit Power and Mining Fatigue are applied to **both** the mining speed and the final destroy progress (this is Bedrock-specific and is why Bedrock mining is faster than Java under Haste — see GeyserMC/Geyser#3987, GeyserMC/Geyser#1621),
  - the underwater and airborne penalties are applied to the speed **before** the final `ceil`-to-tick, so the returned duration matches the number of ticks the client actually spends.
- `BreakDuration(b, i)` keeps its signature and now delegates to `BreakDurationContext` with a zero context (grounded, dry, no effects).
- `Player.breakTime` builds a `BreakContext` from the player's effects/armour/environment instead of scaling the rounded result.
- Add a table-driven test covering the reference values (efficiency, haste, mining fatigue, aqua affinity, underwater, airborne, wrong-tier).

## Notes

- The math follows the **Bedrock** branch of the wiki pseudocode. In particular Mining Fatigue is **not** clamped at level 4 (that clamp is Java-only) and the instant-break threshold is `damage >= 1`.
- The test uses `stretchr/testify`, which adds a dependency. Happy to rewrite it against the standard library if you'd prefer to avoid the dep.


NopeNotDark (approved):
LGTM, Thanks for the PR!

icynutella:
Its wrong.

RestartFU:
gonna open another one

PR #1296, opened 2026-07-02
server/player, server/session: Fix fall state handing

TrippleAWap:
Fixes issue of fallDistance not resetting while flying.
Steps to reproduce:
* Start falling.
* Start flying before you hit the ground.
* Go down to a point where u shouldn't take fall damage.
* Stop flying.


TrippleAWap:
Should be ready, let me know if there's any issues.

PR #1299, opened 2026-07-05
add complete vanilla lodestone support

MEMOxiiii:
## Summary

This adds full vanilla support for lodestones and their interaction with compasses

## Testing

Manually tested linking/relinking a compass to a lodestone, verified needle direction across dimensions, and confirmed inventory sync doesn't desync after linking.

Screenshot 2026-07-06 001441


MEMOxiiii:
@HashimTheArab 

the PR is ready for review

server/block/lodestone.go
```diff
@@ -0,0 +1,94 @@
...
+type Lodestone struct {
+	solid
+	bassDrum
+
+	trackingHandle int32
+}
+
+// BreakInfo returns the lodestone's hardness, blast resistance and drops.
+func (l Lodestone) BreakInfo() BreakInfo {
+	return newBreakInfo(3.5, pickaxeHarvestable, pickaxeEffective, oneOf(Lodestone{})).withBlastResistance(17.5)
```
HashimTheArab, line 23: Bedrock lodestone blast resistance is `3.5`, not `17.5`. As written this also equals the `newBreakInfo` default (`hardness*5`), so it's a no-op documenting the wrong value.

```suggestion
	return newBreakInfo(3.5, pickaxeHarvestable, pickaxeEffective, oneOf(Lodestone{})).withBlastResistance(3.5)
```

server/item/compass.go
```diff
@@ -1,9 +1,55 @@
...
 	return "minecraft:compass", 0
 }
+
+// Glinted returns true if the compass is linked to a lodestone.
+func (c Compass) Glinted() bool { return c.TrackingHandle != 0 }
+
+// EncodeNBT encodes the position tracking handle understood by the Bedrock client.
+func (c Compass) EncodeNBT() map[string]any {
+	return map[string]any{"trackingHandle": c.TrackingHandle}
+}
```
HashimTheArab, line 27: `nbtconv.WriteItem` merges this into every compass stack on both the wire and disk, so unlinked compasses ship a `trackingHandle:0` tag the vanilla client can misread as lodestone-tracking mode (→ spin instead of pointing to spawn). Return `nil` when unlinked:

```suggestion
func (c Compass) EncodeNBT() map[string]any {
	if c.TrackingHandle == 0 {
		return nil
	}
	return map[string]any{"trackingHandle": c.TrackingHandle}
}
```

server/block/lodestone.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/df-mc/dragonfly/server/internal/nbtconv"
+	"github.com/df-mc/dragonfly/server/item"
+	"github.com/df-mc/dragonfly/server/world"
+	"github.com/df-mc/dragonfly/server/world/sound"
+)
+
+// Lodestone is a block that compasses may be linked to in any dimension.
+type Lodestone struct {
+	solid
+	bassDrum
```
HashimTheArab: The lodestone infobox on minecraft.wiki lists the note-block instrument as Default (Harp), not bass drum. Without the `bassDrum` embed, `Note.instrument` falls back to `sound.Piano()` (harp) automatically. Please double-check against a live Bedrock note block first (the wiki's instrument row can be Java-leaning), but if it's harp:

```suggestion
	solid
```

server/block/lodestone.go
MEMOxiiii: i think 17.5 is correct here. minecraft wiki lists lodestone blast resistance as 3.5, but dragonfly seems to store these values multiplied by 5 in breakinfo

for example:
stone is 6 on the wiki, but dragonfly uses 30
basalt is 4.2 on the wiki, but dragonfly uses 21
obsidian is 1200 on the wiki, but dragonfly uses 6000

so lodestone should be 3.5 * 5 = 17.5. using 3.5 here would make it much weaker than the other blocks using the same convention

HashimTheArab: You're correct, thanks!

server/world/world.go
```diff
@@ -193,6 +193,9 @@ func (w *World) blockInChunk(c *Column, pos cube.Pos) Block {
 	if w.conf.Blocks.NBTBlock(rid) {
 		// The block was also a block entity, so we look it up in the block entity map.
 		if b, ok := c.BlockEntities[pos]; ok {
+			if tracked, ok := b.(PositionTrackingBlock); ok && tracked.TrackingHandle() != 0 {
+				w.TrackPosition(pos, tracked.TrackingHandle())
```
HashimTheArab: Reentrant-mutex deadlock: `TrackPosition` takes `w.set.Lock()`, but `ticker.tick` already holds it when it calls `highestObstructingBlock → block → blockInChunk`. `sync.Mutex` isn't reentrant, so the world goroutine hangs permanently. Trigger: spawn Y at the vanilla max sentinel (fresh/imported worlds) + a linked lodestone in the spawn column → freeze at startup. Root issue: a block-read primitive shouldn't acquire `w.set`. Re-register handles on chunk decode, or give the tracker its own mutex.

server/block/lodestone.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// ScheduledTick sends the delayed position tracking update for newly linked
+// lodestone compasses.
+func (l Lodestone) ScheduledTick(pos cube.Pos, tx *world.Tx, _ *rand.Rand) {
+	viewers := tx.Viewers(pos.Vec3Centre())
+	if len(viewers) == 0 {
+		return
+	}
+	tx.World().ObservePositionTracking(l.trackingHandle)
```
HashimTheArab: This marks every handle observed ~1 tick after linking, so `UntrackPosition` always takes the permanent-delete branch and the retain-and-reuse path is dead. Result: compass in a chest → break lodestone → replace in the same spot → retrieve → spins forever. Per the [wiki](https://minecraft.wiki/w/Lodestone) that should stay paired, which is what the `TrackPosition` comment intends. Vanilla never deletes handles — the destroy broadcast is a client spin signal, not DB deletion. Keep the association while inactive and drop the eager delete.

server/world/position_tracking.go
```diff
@@ -0,0 +1,178 @@
...
+	s.Lock()
+	defer s.Unlock()
+	t := positionTracker{byHandle: map[int32]trackedPosition{}, byPosition: map[[4]int]int32{}, observed: map[int32]struct{}{}}
+	for _, entry := range entries {
+		if entry.Handle == 0 {
+			continue
+		}
+		t.byHandle[entry.Handle] = trackedPosition{pos: entry.Position, dim: entry.Dimension, active: entry.Active}
+		t.byPosition[[4]int{entry.Dimension, entry.Position[0], entry.Position[1], entry.Position[2]}] = entry.Handle
+		if entry.Handle > t.next {
```
HashimTheArab, line 113: `next` is rebuilt as max(handle), but observed-destruction deletes entries — so it reloads too low and reallocates a destroyed handle after a restart, pointing stored compasses at an unrelated lodestone. This breaks the "never reused" invariant the code states. Persist `next` (vanilla keeps `PositionTrackDB-LastId`).

server/session/handler_position_tracking_db.go
```diff
@@ -0,0 +1,53 @@
...
+// Handle responds with the tracked position, or marks the target as unavailable.
+func (*PositionTrackingDBHandler) Handle(p packet.Packet, s *Session, tx *world.Tx, _ Controllable) error {
+	pk, ok := p.(*packet.PositionTrackingDBClientRequest)
+	if !ok {
+		return fmt.Errorf("expected *packet.PositionTrackingDBClientRequest, got %T", p)
+	}
+	if pk.RequestAction != packet.PositionTrackingDBRequestActionQuery {
+		return fmt.Errorf("unknown position tracking request action %d", pk.RequestAction)
+	}
+	pos, dim, found := tx.World().TrackedPosition(pk.TrackingID)
```
HashimTheArab: No possession check: any client can walk sequential IDs to read every linked lodestone's coords + dimension (other players' bases), and queries mutate state — `found` observes handles the requester doesn't own, and an inactive hit deletes the owner's reuse association. PNX checks the player holds the item first and never mutates on a query. Verify possession and keep queries read-only.

server/world/mcdb/leveldat/data.go
```diff
@@ -272,4 +291,11 @@ func (d *Data) PutSettings(s *world.Settings) {
 	d.GameType = int32(mode)
 	difficulty, _ := world.DifficultyID(s.Difficulty)
 	d.Difficulty = int32(difficulty)
+	d.DragonflyPositionTracking = d.DragonflyPositionTracking[:0]
```
HashimTheArab, line 295: Data race: all three dimensions' `autoSave` goroutines call `PutSettings` on the same `db.ldat`. This now does `[:0]`+append on a shared slice (concurrent with `Close`'s `Marshal`) instead of scalar writes — a real race that can tear the entry list. Guard `db.ldat` with a mutex.

MEMOxiiii:
@HashimTheArab 

done

server/world/world.go
```diff
@@ -1389,7 +1399,11 @@ func (w *World) columnFrom(c *chunk.Column, _ ChunkPos) *Column {
 			w.conf.Log.Error("read column: block with nbt does not implement NBTer", "block", fmt.Sprintf("%#v", b))
 			continue
 		}
-		col.BlockEntities[be.Pos] = nb.DecodeNBT(be.Data).(Block)
+		decoded := nb.DecodeNBT(be.Data).(Block)
+		if tracked, ok := decoded.(PositionTrackingBlock); ok && tracked.TrackingHandle() != 0 {
+			decoded = tracked.WithTrackingHandle(w.TrackPosition(be.Pos, tracked.TrackingHandle()))
```
HashimTheArab, line 1407: The deadlock I flagged in `blockInChunk` moved here but is still live: `TrackPosition` takes `w.set.Lock()`, and this runs via `chunk → loadChunk → columnFrom`, which is reachable while `ticker.tick` already holds `w.set` (tick.go:50 → `highestObstructingBlock` → `w.chunk`). Same goroutine, non-reentrant mutex → hang. Trigger: an overworld with spawn Y at the max sentinel (fresh/imported worlds) whose spawn chunk holds a lodestone with a non-zero handle → freeze on first tick. The tracker just can't be mutated under `w.set` — give it its own mutex, or register lazily off any `w.set`-holding path.

Related: `setBlock` (line 311) calls `TrackPosition` with no `!= 0` guard (unlike here), so every placed lodestone — even unlinked — gets a handle and serializes a non-zero `trackingHandle` (vanilla unlinked = 0). That also widens the deadlock trigger to any persisted lodestone. Worth guarding on `TrackingHandle() != 0` for symmetry, but the mutex split is the real fix.

HashimTheArab (commented):
Thanks, the deadlock issue is still active:

server/world/world.go
HashimTheArab: give the tracker its own mutex instead of reusing w.set

MEMOxiiii: done

HashimTheArab (commented):
`setBlock` (world.go:311) still calls `TrackPosition` without the `!= 0` guard that `columnFrom` has, so every placed lodestone — even unlinked — gets a handle and serializes a non-zero `trackingHandle` (vanilla unlinked = 0), which also grows the tracker/level.dat unbounded.

MEMOxiiii:
> `setBlock` (world.go:311) still calls `TrackPosition` without the `!= 0` guard that `columnFrom` has, so every placed lodestone — even unlinked — gets a handle and serializes a non-zero `trackingHandle` (vanilla unlinked = 0), which also grows the tracker/level.dat unbounded.

done

server/item/compass.go
```diff
@@ -1,9 +1,58 @@
...
+	case int32:
+		c.TrackingHandle = handle
+	case int:
+		c.TrackingHandle = int32(handle)
+	}
+	return c
+}
+
+// LodestoneCompass is the registered item form of a compass linked to a lodestone.
+type LodestoneCompass struct{}
```
HashimTheArab: The `LodestoneCompass` shim isn't idiomatic — it exists only to claim the `minecraft:lodestone_compass` registry name, and its `DecodeNBT` morphs into a different type (`Compass`). Dragonfly's pattern for "same Go type, variant identity" is to register the type twice: see `Arrow` — `register.go` does `RegisterItem(Arrow{})` and `RegisterItem(Arrow{Tip: p})`, and `EncodeItem` switches name/meta on the field (arrow.go:14). The registry keys on `{name, meta}` (world/item.go), so `RegisterItem(Compass{TrackingHandle: 1})` claims `minecraft:lodestone_compass` directly. That lets you delete this whole type (4 methods) and the type-morphing `DecodeNBT`, registering `Compass{}` and `Compass{TrackingHandle: 1}` instead.

One caveat if you do: `Compass.DecodeNBT` (line 30) should reset `c.TrackingHandle = 0` before reading, so the registered sentinel `1` can't leak if the NBT tag is ever absent (defensive — real items always carry it).

HashimTheArab (commented):
- **`UseItemOnBlock` branch (low/med).** `player.go` `if useCtx.CountSub >= i.Count() && !useCtx.NewItem.Empty()` is necessary (creative skips `subtractItem`, so without it a creative relink dupes the stack) and doesn't regress the other 6 `NewItem` Activatables (none set `NewItem`). But it's a single-block special-case inferred via a `CountSub >= Count` heuristic on the shared hot path — a `UseContext` capability (e.g. an explicit in-place-replace field the block sets) handled uniformly would be more idiomatic, and the comment should spell out the real reason (creative no-subtract → dupe avoidance).

- **Design-placement, discuss (low).** (a) `Settings` now embeds a mutex+maps `positionTracker`; dragonfly otherwise keeps world runtime state on `World` (`scheduledUpdates`, `entities`, `viewers`) and `Settings` is a flat bag of saved scalars. The justification is real (the tracker must be shared across the 3 dimension `World`s and `*Settings` is the only shared object) — but a dedicated shared type referenced by all three `World`s may be preferred. (b) `DragonflyPositionTracking` is the only `Dragonfly`-prefixed level.dat key; vanilla stores this in separate LevelDB keys (`PositionTrackDB-LastId`, `PositionTrackDB-0x…`), and retained entries are never pruned so it grows unbounded — consider the leveldb keyspace + persisting `next` explicitly.

server/item/compass.go
MEMOxiiii: i pushed the changes but wanted to check the direction before committing

i removed the separate LodestoneCompass type and went back to the existing DF variant-registration style with Compass{} and Compass{TrackingHandle: 1}. i also made Compass.DecodeNBT reset TrackingHandle before reading the tag, and made compass decoding run even when the nbt is empty so the registration sentinel cannot leak

MEMOxiiii: for the UseItemOnBlock point, i replaced the CountSub >= Count heuristic with an explicit UseContext.ReplaceHeldItem flag, and lodestone only sets it for the relink case

MEMOxiiii: for the tracker placement point, i kept the data in Settings because it still needs to be shared across dimensions, but changed it into a dedicated shared *PositionTracker component with its own mutex, and persisted next explicitly through level.dat

MEMOxiiii: does this match the direction you wanted ??

PR #1302, opened 2026-07-07
world: add synchronous mode for deterministic, in-process worlds

HashimTheArab:
## Synchronous worlds

This PR adds `world.Config.Synchronous`. A synchronous `World` starts no background goroutines: transactions created through `World.Exec` are executed directly on the calling goroutine, the world is not saved automatically, unused chunks are not unloaded automatically, and no automatic ticking happens. Time only passes through the new `World.AdvanceTick` method, which runs exactly one tick synchronously.

## Why

Block, item and entity behaviour is currently very hard to unit test: constructing a `World` spins up a tick loop, autosave and transaction goroutines, and anything time-dependent (neighbour updates, scheduled ticks, random ticks) fires asynchronously at 20 TPS. As a result, `server/world` and `server/block` had no tests at all. A synchronous world is fully deterministic — set up state in one `Exec`, advance ticks explicitly, and assert the outcome, all in-process:

```go
w := world.Config{Synchronous: true, Entities: entity.DefaultRegistry}.New()
defer w.Close()

<-w.Exec(func(tx *world.Tx) {
    tx.SetBlock(support, block.Stone{}, nil)
    tx.SetBlock(torch, block.Torch{Facing: cube.FaceDown}, nil)
    tx.SetBlock(support, block.Air{}, nil)
})
w.AdvanceTick() // neighbour update fires deterministically here
```

`TestTorchBreaksWithoutSupport` (included) uses exactly this pattern; the other new tests cover the mode's contract (no goroutines started, inline `Exec`, no automatic ticking, prompt `Close`).

## Implementation notes

- `Exec` reuses `normalTransaction.Run` inline. `weakExec` runs the weak transaction body inline but skips the `cond.L`/`Broadcast` handshake: that handshake exists to wake a goroutine already waiting on `cond`, and in the inline case the caller only inspects the buffered channel after `weakExec` returns (taking `cond.L` inline would self-deadlock, as the caller in `EntityHandle` still holds it).
- `ticker.tick`'s "skip when no viewers" guard is bypassed for synchronous worlds, since their ticks only ever happen explicitly via `AdvanceTick`.
- Entities and viewers work unchanged; they simply don't advance between `AdvanceTick` calls.

### DIFF
--- a/server/block/block_behaviour_test.go
+++ b/server/block/block_behaviour_test.go
@@ -1,0 +1,32 @@
+package block_test
+
+import (
+	"testing"
+
+	"github.com/df-mc/dragonfly/server/block"
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/df-mc/dragonfly/server/entity"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// TestTorchBreaksWithoutSupport verifies that a torch is broken by a neighbour
+// update on the tick after its supporting block is removed, using a
+// synchronous World to make the tick deterministic.
+func TestTorchBreaksWithoutSupport(t *testing.T) {
+	w := world.Config{Synchronous: true, Entities: entity.DefaultRegistry}.New()
+	defer w.Close()
+
+	support, torch := cube.Pos{0, 0, 0}, cube.Pos{0, 1, 0}
+	<-w.Exec(func(tx *world.Tx) {
+		tx.SetBlock(support, block.Stone{}, nil)
+		tx.SetBlock(torch, block.Torch{Facing: cube.FaceDown}, nil)
+		tx.SetBlock(support, block.Air{}, nil)
+	})
+	w.AdvanceTick()
+
+	<-w.Exec(func(tx *world.Tx) {
+		if b := tx.Block(torch); b != (block.Air{}) {
+			t.Errorf("expected torch to break after removing its support, got %v", b)
+		}
+	})
+}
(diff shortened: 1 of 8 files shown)


RestartFU:
looks good but those comments are wayyy too verbose 💀

HashimTheArab:
I think they earn their spot, I read them and it's descriptive enough to describe what this does and make sure nobody misuses it

HashimTheArab:
I've trimmed the comments a bit

server/block/block_behaviour_test.go
RestartFU, line 1: maybe this file should be named something like torch_test.go so that it appears near the torch.go file?

server/world/world_test.go
RestartFU, line 1: same here I feel like this should be named something like sync_test.go

server/block/block_behaviour_test.go
HashimTheArab: the goal is to include future tests here and not clutter with files

RestartFU (dismissed):

HashimTheArab:
Pushed fixes for a few edge cases found in further testing:

- `ExecWorld` self-removal deadlocked on synchronous worlds — inline weak transactions now release the entity cond lock while running, matching the async path's locking contract
- viewerless synchronous worlds now actually tick entities, random blocks and block entities in loaded chunks (previously only time/weather/scheduled updates advanced)
- `strikeLightning` no longer panics when the config has no `Entities` registry (default sync worlds turn thunder on via the weather cycle)
- documented the single-goroutine contract for `Synchronous` and nested `Exec` behaviour

All covered by regression tests.

didntpot (approved):

RestartFU (approved):

PR #1305, opened 2026-07-09
Add owner-safe scheduling APIs replacing public blocking transactions

HashimTheArab:
## Summary

`World.Exec` and `EntityHandle.ExecWorld` make it easy for user code to self-deadlock: a blocking call that is safe from an outside goroutine deadlocks inside an event handler, command or packet callback that is already running on that world's owner (the single goroutine that runs all of a world's transactions), and it is often not obvious which of the two a given function is. This PR replaces them with owner-scoped scheduling as the public re-entry path (full design rationale with repository evidence: [HashimTheArab/dragonfly#33](https://github.com/HashimTheArab/dragonfly/issues/33)):

- **`world.Tx`** remains what it has always been — the handle a callback holds while it has exclusive, serialized access to the world — and owns all world operations; the ~250 existing `*world.Tx` gameplay signatures are unchanged and final. **`world.Context`** becomes a small cancellable event scope embedding `*Tx`, created per handler dispatch via `Tx.Event()`, so cancelling one event cannot affect another fired in the same tick — and non-cancellable events (`HandleEntitySpawn`, `HandleEntityDespawn`, `HandleClose`) take `*Tx` directly, making a no-op `Cancel()` unrepresentable.
- **Fire-and-forget scheduling**: `World.Do`/`DoAfter`, `EntityHandle.Do`/`DoAfter`, typed `world.EntityRef[T]`, `player.Ref`/`Player.Do`/`DoAfter` and `Context.Defer`. Entity tasks follow the entity across world changes and fail with defined errors (`ErrEntityClosed`, `ErrWorldClosed`, `ErrEntityType`) when the entity or world closes first.
- **Blocking request/response for off-owner code**: `world.Call`/`CallEntity`/`CallRef` return a typed result and a `*world.Task` handle (`Done`/`Err`/`Wait`/`OnDone`/`Cancel`). Callback panics are captured into `world.PanicError` (with stack), logged through the world logger, and re-panicked on session-facing paths so they still surface loudly.
- **Migrated internals**: sessions, server join and player iteration, respawn/disconnect and delayed actions (goat horn, death timing) run on the new paths with defined behaviour when a world or entity closes mid-flight. World close drains in-flight scheduled work in a defined order before saving. Commands may run with a nil `*world.Tx` when the source is not attached to a world.

## Migration

```go
// Before: blocks, and deadlocks if already inside this world's owner.
<-w.Exec(func(tx *world.Tx) {
    tx.SetBlock(pos, block.Stone{}, nil)
})

// After: fire-and-forget from anywhere.
w.Do(func(tx *world.Tx) {
    tx.SetBlock(pos, block.Stone{}, nil)
})

// Off-owner code that needs a result:
count, err := world.Call(ctx, w, func(tx *world.Tx) (int, error) { ... })

// Delayed entity work follows the entity across worlds:
handle.DoAfter(time.Second, func(tx *world.Tx, e world.Entity) { ... })

// Typed player re-entry from outside a callback:
player.Do(handle, func(tx *world.Tx, p *player.Player) {
    p.Message("hello")
})
```

Inside handlers and commands nothing changes: the callback context is already the owner, so world operations are used directly. `player.Context` now embeds `*world.Context`, so player event handlers can do world operations straight off the event context.

## Breaking changes

- `World.Exec` and `EntityHandle.ExecWorld` are removed (replaced by `Do`/`Call*` above).
- `world.Context`/`player.Context` become concrete event-scope types replacing the `event.Context` aliases; `ctx.Val()` becomes `ctx.Player()` for player events, and world operations are available directly on the event context. `HandleEntitySpawn`/`HandleEntityDespawn`/`HandleClose` now take `*world.Tx` (they were never cancellable).
- `session.Config.HandleStop` may now receive a nil `*world.Tx` for a controllable that could not be restored to any world during teardown.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race -count=2 ./server/world ./server/session ./server/player`.
- Regression tests are kept for the invariants that are easiest to silently break: owner-queue-full deadlocks (both the task and weak-transaction paths), close-drain ordering, weak-transaction cancellation, and the AddEntity/scheduled-callback handoff.

### DIFF
--- a/server/block/block_behaviour_test.go
+++ b/server/block/block_behaviour_test.go
@@ -1,6 +1,7 @@
 package block_test
 
 import (
+	"context"
 	"testing"
 
 	"github.com/df-mc/dragonfly/server/block"
@@ -17,16 +18,20 @@ func TestTorchBreaksWithoutSupport(t *testing.T) {
 	defer w.Close()
 
 	support, torch := cube.Pos{0, 0, 0}, cube.Pos{0, 1, 0}
-	<-w.Exec(func(tx *world.Tx) {
+	w.Do(func(tx *world.Context) {
 		tx.SetBlock(support, block.Stone{}, nil)
 		tx.SetBlock(torch, block.Torch{Facing: cube.FaceDown}, nil)
 		tx.SetBlock(support, block.Air{}, nil)
 	})
 	w.AdvanceTick()
 
-	<-w.Exec(func(tx *world.Tx) {
-		if b := tx.Block(torch); b != (block.Air{}) {
-			t.Errorf("expected torch to break after removing its support, got %v", b)
-		}
+	b, err := world.Call(context.Background(), w, func(tx *world.Context) (world.Block, error) {
+		return tx.Block(torch), nil
 	})
+	if err != nil {
+		t.Fatalf("read torch block: %v", err)
+	}
+	if b != (block.Air{}) {
+		t.Errorf("expected torch to break after removing its support, got %v", b)
+	}
 }
--- a/server/block/block_behaviour_test.go
+++ b/server/block/block_behaviour_test.go
@@ -18,14 +18,14 @@ func TestTorchBreaksWithoutSupport(t *testing.T) {
 	defer w.Close()
 
 	support, torch := cube.Pos{0, 0, 0}, cube.Pos{0, 1, 0}
-	w.Do(func(tx *world.Context) {
+	w.Do(func(tx *world.Tx) {
 		tx.SetBlock(support, block.Stone{}, nil)
 		tx.SetBlock(torch, block.Torch{Facing: cube.FaceDown}, nil)
 		tx.SetBlock(support, block.Air{}, nil)
 	})
 	w.AdvanceTick()
 
-	b, err := world.Call(context.Background(), w, func(tx *world.Context) (world.Block, error) {
+	b, err := world.Call(context.Background(), w, func(tx *world.Tx) (world.Block, error) {
 		return tx.Block(torch), nil
 	})
 	if err != nil {
--- a/server/block/explosion.go
+++ b/server/block/explosion.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/block/cube/trace"
-	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/item"
 	"github.com/df-mc/dragonfly/server/world"
 	"github.com/df-mc/dragonfly/server/world/particle"
@@ -134,7 +133,7 @@ func (c ExplosionConfig) Explode(tx *world.Tx, explosionPos mgl64.Vec3) {
 		}
 	}
 
-	ctx := event.C(tx)
+	ctx := tx.Event()
 	spawnFire := c.SpawnFire
 	itemDropChance := c.ItemDropChance
 	if tx.World().Handler().HandleExplosion(ctx, explosionPos, &affectedEntities, &affectedBlocks, &itemDropChance, &spawnFire); ctx.Cancelled() {
--- a/server/block/farmland.go
+++ b/server/block/farmland.go
@@ -4,7 +4,6 @@ import (
 	"math/rand/v2"
 
 	"github.com/df-mc/dragonfly/server/block/cube"
-	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/world"
 )
 
@@ -74,7 +73,7 @@ func (f Farmland) hydrated(pos cube.Pos, tx *world.Tx) bool {
 func (f Farmland) EntityLand(pos cube.Pos, tx *world.Tx, e world.Entity, _ *float64) {
 	if living, ok := e.(livingEntity); ok {
 		if fall, ok := living.(fallDistanceEntity); ok && rand.Float64() < fall.FallDistance()-0.5 {
-			ctx := event.C(tx)
+			ctx := tx.Event()
 			if tx.World().Handler().HandleCropTrample(ctx, pos); !ctx.Cancelled() {
 				tx.SetBlock(pos, Dirt{}, nil)
 			}
--- a/server/block/fire.go
+++ b/server/block/fire.go
@@ -4,7 +4,6 @@ package block
 
 import (
 	"github.com/df-mc/dragonfly/server/block/cube"
-	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/item"
 	"github.com/df-mc/dragonfly/server/item/enchantment"
 	"github.com/df-mc/dragonfly/server/world"
@@ -175,12 +174,12 @@ func (f Fire) tick(pos cube.Pos, tx *world.Tx, r *rand.Rand) {
 // this might end up not happening.
 func (f Fire) spread(from, to cube.Pos, tx *world.Tx, r *rand.Rand) {
 	if _, air := tx.Block(to).(Air); !air {
-		ctx := event.C(tx)
+		ctx := tx.Event()
 		if tx.World().Handler().HandleBlockBurn(ctx, to); ctx.Cancelled() {
 			return
 		}
 	}
-	ctx := event.C(tx)
+	ctx := tx.Event()
 	if tx.World().Handler().HandleFireSpread(ctx, from, to); ctx.Cancelled() {
 		return
 	}
--- a/server/block/lava.go
+++ b/server/block/lava.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/df-mc/dragonfly/server/block/cube"
-	"github.com/df-mc/dragonfly/server/event"
 	"github.com/df-mc/dragonfly/server/world"
 	"github.com/df-mc/dragonfly/server/world/sound"
 )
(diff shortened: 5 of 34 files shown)


HashimTheArab:
I've trimmed the tests I had in the original code before opening this PR, the remaining ones are for logic that could silently break and cause a deadlock, etc.

- 2 deadlock regressions — an owner callback calling World.Do while the transaction queue is full, for both the task path and the weak-transaction path. This is not hypothetical: the weak-transaction variant was a real world-freezing deadlock found during review, and the fix is a lock-ordering discipline (scheduleMu must never be held across a blocking queue send) that nothing in the type system enforces. Any future edit to weakExec/scheduleTask can silently reintroduce it.
- 3 close-drain ordering tests — work queued before close runs before HandleClose, entity work scheduled during the close transaction still drains before the queue shuts, and entity work blocked when close starts fails promptly instead of hanging. World close is a five-signal state machine (closed/closeStarted/closing/queueClosing + WaitGroups); reordering two lines in close() breaks these without any visible diff smell.
- 1 cancellation-poisoning test — cancelling a task whose weak transaction was invalidated mid-flight must not leave weakTxActive stuck and wedge every future transaction on that entity handle.
- 1 handoff test — scheduled callbacks can't observe an entity mid-AddEntity (the worldReady handshake); breaking it produces races, not errors.

HashimTheArab:
# Migrating from `Exec` to owner scheduling

The blocking transaction entry points `World.Exec` and `EntityHandle.ExecWorld` are removed. World work is now scheduled onto the world's owner: the single goroutine that runs all of a world's transactions, ticking included. Any code that receives a `*world.Tx` is already on the owner.

From any goroutine, work is scheduled with `Do` and `DoAfter` (fire-and-forget). From inside a callback, `tx.Defer` schedules follow-up work on the same owner. Code that is not on the owner and needs a result uses `world.Call`, `world.CallEntity` or `world.CallRef`. Every scheduled call returns a `*world.Task` that reports success or failure (see "Handling failures").

## What changes

- Awaiting `Exec` from inside a handler, command or packet callback could deadlock the world permanently, and whether a given function was "inside" was invisible at the call site. `Do` and `Defer` are safe from any goroutine.
- Work scheduled against a world or entity that closes first fails with a defined error (`ErrWorldClosed`, `ErrEntityClosed`, and so on) on the returned task, rather than hanging or silently not running.
- Delayed entity work follows the entity: a task scheduled on a player who travels through a portal runs in the world they arrive in.
- `player.Ref` and `world.EntityRef[T]` hand callbacks a typed value directly, removing the `e.(*player.Player)` casts.
- A panicking fire-and-forget callback is recovered, logged with its stack through the world's logger, and recorded on the task. The synchronous `Call` functions re-panic on the calling goroutine, like a normal function call.
- Each handler event dispatch gets its own cancel scope, so cancelling one event cannot affect another fired in the same tick.
- Existing `*world.Tx` signatures are unchanged. The one rename in handler bodies is `ctx.Val()` → `ctx.Player()`.

## Fire-and-forget world work

```go
// Before — blocks, and deadlocks if already on this world's owner:
<-w.Exec(func(tx *world.Tx) {
    tx.SetBlock(pos, block.Stone{}, nil)
})

// After — returns immediately, safe from anywhere:
w.Do(func(tx *world.Tx) {
    tx.SetBlock(pos, block.Stone{}, nil)
})
```

## Blocking calls with results (off-owner only)

For goroutines, startup code and tests — never from inside an owner callback, where it deadlocks.

```go
// Before — hand-rolled capture through a closed-over variable:
var count int
<-w.Exec(func(tx *world.Tx) { count = len(slices.Collect(tx.Entities())) })

// After:
count, err := world.Call(ctx, w, func(tx *world.Tx) (int, error) {
    return len(slices.Collect(tx.Entities())), nil
})
```

The same shape exists for entities and players: `world.CallEntity(ctx, handle, f)`, `world.CallRef(ctx, ref, f)` and `player.Call(ctx, handle, f)`.

## Delayed entity work

```go
// Before — manual re-entry that breaks if the entity changes worlds:
time.AfterFunc(time.Second, func() {
    handle.ExecWorld(func(tx *world.Tx, e world.Entity) { ... })
})

// After — follows the entity, fails with a defined error if it is gone:
handle.DoAfter(time.Second, func(tx *world.Tx, e world.Entity) { ... })
```

## Reaching a player from outside a callback

```go
// Before:
handle.ExecWorld(func(tx *world.Tx, e world.Entity) {
    e.(*player.Player).Message("hello")
})

// After:
player.Do(handle, func(tx *world.Tx, p *player.Player) {
    p.Message("hello")
})
```

## Storing references to players and entities

`*player.Player` and `world.Entity` values are only valid inside the callback that received them. Code that holds players across time — a minigame roster, an arena's participants — stores the stable reference and re-enters through it:

```go
type Arena struct {
    players []player.Ref // not []*player.Player
}

for _, r := range a.players {
    r.Do(func(tx *world.Tx, p *player.Player) { p.Message("round over") })
}
```

Refs are built with `player.NewRef(h)` and `world.NewEntityRef[T](h)`. A one-shot call does not need a ref: `player.Do(h, f)` works directly on the handle.

## Deferred work

The old workarounds for running something right after the current event were the un-awaited nested `Exec` and the goroutine paid purely to make the wait safe:

```go
// Before — fire-and-forget nested Exec (awaiting it would deadlock):
tx.World().Exec(func(tx *world.Tx) { ... })

// Before — or re-entry from a goroutine:
go func() {
    <-w.Exec(func(tx *world.Tx) { ... })
}()

// After:
tx.Defer(func(tx *world.Tx) { ... })
```

`Defer` runs on the same owner immediately after the current callback finishes, ahead of everything else in the world's queue, so it sees the world essentially as the callback left it. Deferred callbacks run FIFO in registration order — not Go `defer`'s LIFO — and each receives a fresh `*Tx`; the one from the current callback must not be captured. On a `player.Context`, `ctx.Defer` also re-resolves the player for that moment.

Typical uses are removing or adding entities found while iterating `tx.Entities()` (mutating the world mid-iteration is unsafe) and acting after an event's default behaviour has applied, since handlers run before it.

One behavioural difference from the old un-awaited `Exec`: that went to the back of the world's queue, so other queued work ran in between. `Defer` jumps the queue. Code that relied on back-of-queue ordering can use `w.Do` from inside the callback instead.

| Situation | Use |
|---|---|
| Right after this callback, same world | `tx.Defer` |
| Soon, from anywhere | `w.Do` |
| Later | `DoAfter` |
| Wherever the entity is by then | `handle.Do` |

## Iterating players

`Server.Accept` and `Server.Players` run the loop body on the player's world owner. Blocking there stalls that world, and calling `world.Call*` or `Task.Wait` from the loop body deadlocks. With a non-nil `tx`, `Players` yields players in other worlds by blocking on those owners one at a time — keep it out of latency-sensitive code, and note that two worlds whose handlers iterate each other can deadlock.

Players yielded are only valid inside the loop body. To act on them later, collect handles and fan out:

```go
var handles []*world.EntityHandle
for p := range srv.Players(nil) {
    handles = append(handles, p.H())
}
for _, h := range handles {
    player.Do(h, func(tx *world.Tx, p *player.Player) { p.Message("hi") })
}
```

## Handlers and commands

Code inside handlers and commands is already on the owner and keeps using the context it is given. The one rename for player events is `ctx.Val()` → `ctx.Player()`, and world operations are available directly on the event context:

```go
func (h MyHandler) HandleBlockBreak(ctx *player.Context, pos cube.Pos, ...) {
    ctx.Player().Message("broken")
    ctx.SetBlock(pos.Side(cube.FaceUp), block.Air{}, nil)
}
```

For `world.Handler` implementors: cancellable events still take `ctx *world.Context`, which embeds the transaction. `HandleEntitySpawn`, `HandleEntityDespawn` and `HandleClose` now take `tx *world.Tx` — they were never cancellable, and the signature now says so.

For command authors: `Runnable.Run`'s `tx` is nil when the source is not attached to a world, such as a console source. Commands that use `tx` must nil-check it; target selectors already fail politely.

## Handling failures

Every `Do`, `DoAfter` and `Defer` returns a `*world.Task`:

```go
task := handle.DoAfter(time.Second, func(tx *world.Tx, e world.Entity) {
    // ...
})
task.OnDone(func(err error) {
    switch {
    case err == nil:                            // ran fine
    case errors.Is(err, world.ErrEntityClosed): // entity despawned or closed before it fired
    case errors.Is(err, world.ErrWorldClosed):  // its world shut down first
    case errors.Is(err, world.ErrTaskPanicked): // the callback panicked (stack in *world.PanicError)
    }
})
```

Other errors a task can report: `world.ErrEntityType` (a typed ref's task ran but the entity is no longer that type), `world.ErrEntityNotInWorld` (a `player.Context.Defer` fired after the player moved to another world — the player is alive, just not here; follow it with `player.Do` if that is the intent) and `world.ErrTaskCancelled` (the task was stopped with `Cancel()`).

There are three ways to consume a task. `OnDone(f)` registers a completion hook that always runs on a fresh goroutine, so it is safe to register anywhere — but the hook itself runs off-owner: schedule again rather than touching world state from it. `Wait(ctx)`, `Err()` and `Done()` are for code already off the owner, such as background goroutines, shutdown paths and tests; calling `Wait` from inside an owner callback blocks the owner on itself. `Cancel()` aborts a task that has not started yet.

Ignoring the task is also fine: a panicking callback is still logged with its stack through the world's logger. The synchronous `Call` functions re-panic on the waiting goroutine automatically. To escalate a `Do` task's captured panic manually:

```go
if pe, ok := errors.AsType[*world.PanicError](task.Err()); ok {
    panic(pe.Value) // original stack is in pe.Stack and already logged
}
```

RestartFU (approved):

PR #1122, opened 2025-10-01
Implement Nether Portals

xNatsuri:
This PR is built off of #426 
image
image

PR #1308, opened 2026-07-10
server: Implement End portals

HashimTheArab:
## What

Implements End portals on top of the nether portal machinery from #1122:

- **`block.EndPortalFrame`** — placeable frame with `Facing` and `Eye` states, following the Bedrock `cardinal_direction` convention (frames face opposite the placing player, so building a ring from the centre yields the valid inward-facing configuration). Indestructible, like bedrock.
- **`item.EnderEye`** — using it on an empty frame inserts the eye; when the twelfth frame of a valid ring is filled, the 3x3 interior fills with `block.EndPortal`. Eyes cannot be removed once inserted.
- **`block.EndPortal`** — light level 15, no item form, teleports entities to and from the End. Breaking a frame of an active portal removes the connected portal blocks; breaking a portal block itself leaves the rest in place, since the ring is still complete.
- **`world/portal`** — `EndPortalFromPos`/`ActivateEndPortal` validate a complete twelve-frame ring from any frame on it; `GenerateEndSpawnPlatform` rebuilds the 5x5 obsidian arrival platform at (100, 48, 0) and clears the three air layers above it on every entry, matching vanilla's unconditional regeneration.

## Travel behaviour

- End travel is always instant, in both directions, regardless of game mode.
- Arrival in the End is on the spawn platform: players at y=49, other entities one block higher at y=50 (Bedrock values).
- Returning from the End leads to the player's spawn point (bed), falling back to the world spawn; non-player entities go to the world spawn. Custom `PortalDestination` pairings other than End→Overworld keep the normal portal search.

## API changes

- `entity.PortalTravelComputer.Instantaneous` changed from `func() bool` to `func(source, target world.Dimension) bool` so the End can be instant while nether portals keep the timer — breaking for anyone constructing the computer directly.
- New optional `entity.PortalTravelComputer.SpawnPoint` hook for the End-return arrival position.

## Known gaps

- The eye of ender projectile (stronghold locating throw) is not implemented; the item currently only fills frames.
- Blocks destroyed by platform regeneration are removed silently rather than dropped as items.
- The arriving player's rotation is not set (vanilla faces west).
- Frames emit no comparator signal; dragonfly has no analog comparator output yet (vanilla outputs 15 when the eye is inserted).


server/block/model/end_portal_frame.go
```diff
@@ -0,0 +1,27 @@
...
+	// Eye is true if an eye of ender has been inserted into the frame.
+	Eye bool
+}
+
+// BBox ...
+func (f EndPortalFrame) BBox(cube.Pos, world.BlockSource) []cube.BBox {
+	boxes := []cube.BBox{cube.Box(0, 0, 0, 1, 0.8125, 1)}
+	if f.Eye {
+		boxes = append(boxes, cube.Box(0.3125, 0.8125, 0.3125, 0.6875, 1, 0.6875))
+	}
```
TwistedAsylumMC: This is not accurate to bedrock, the eye does not get included in the hitbox

HashimTheArab: The extracted Bedrock block-state data includes an eye collision volume, but live vanilla testing shows arrows pass through the eye. Dragonfly uses the model BBox for projectile intersection as well as entity collision, so retaining that volume caused observable projectile behaviour to diverge from vanilla. Removed the eye box; the frame now remains 13/16 high in both states.

PR #1314, opened 2026-07-11
world/chunk: Reject invalid network subchunk bounds

HashimTheArab:
## Summary

Prevent malformed network chunk data from panicking `NetworkDecodeBuffer` by validating both the declared subchunk count and version 9 encoded subchunk indices before indexing the destination chunk.

## Root cause

`LevelChunk.SubChunkCount` controls the decode loop, while version 9 subchunks may replace the loop index with an absolute Y value from the payload. Both values could reach `c.sub[index]` without being checked against the chunk range, producing an out-of-range panic.

## Changes

- reject negative or oversized declared subchunk counts before decoding
- validate the decoded version 9 index before assigning the subchunk

## Validation

- `go test ./...`


HashimTheArab:
Encountered a ton of panics from these missing guards when decoding chunks from various servers

PR #1315, opened 2026-07-11
server: enable IPv6 LAN discovery

HashimTheArab:
## Summary
- listen on `[::]:19133` by default alongside the IPv4 listener on `:19132`, using the server's existing multi-listener support (one `Config.Listeners` entry per address)
- allow disabling or overriding the IPv6 listener through `Network.AddressV6`
- preserve one server-wide player limit and advertised player count across both listeners via a shared gophertunnel `ListenerGroup` (Sandertv/gophertunnel#465)
- if the IPv6 bind fails, log the error and keep serving on IPv4 instead of failing startup (`conf.New` previously appended a nil listener when a listener factory errored — fixed with `continue`)

PocketMine documents UDP 19133 as the IPv6 port used by Bedrock LAN discovery: https://github.com/pmmp/PocketMine-MP/blob/stable/changelogs/4.0.md#general

Note: the `go.mod`/`go.sum` churn comes from a temporary `replace` pointing at my gophertunnel fork with the `ListenerGroup` support; it will be dropped once gophertunnel tags a release including Sandertv/gophertunnel#465.

## Validation
- `go test ./...`
- `git diff --check`


HashimTheArab:
Blocked on https://github.com/Sandertv/gophertunnel/pull/465. The temporary `replace` will be removed once the shared listener state is available upstream.

RestartFU (approved):

go.mod
```diff
@@ -13,22 +13,46 @@ require (
...
 	golang.org/x/exp v0.0.0-20250103183323-7d7fa50e5329
-	golang.org/x/mod v0.22.0
-	golang.org/x/text v0.23.0
-	golang.org/x/tools v0.28.0
+	golang.org/x/mod v0.32.0
+	golang.org/x/text v0.34.0
+	golang.org/x/tools v0.41.0
 )
 
+replace github.com/sandertv/gophertunnel => github.com/HashimTheArab/gophertunnel v1.25.3-0.20260711143608-4724458c7a08
```
Silentmussle913, line 21: shouldnt this not be here? like till the gophertunnel pr gets accepted and then just use that? rather than a fork

PR #1321, opened 2026-07-12
Keep quick regen feature

FDUTCH:


FDUTCH:
tested, everything works

PR #1322, opened 2026-07-12
Feature/tinted glass

MEMOxiiii:
adds tinted glass block with correct vanilla properties

before: light passes through

image

after: light is blocked

image


MEMOxiiii:
ready for review

PR #1331, opened 2026-07-15
server/block: Implement bubble columns

HashimTheArab:
## Summary

- add and register bubble column block states
- form and update columns above soul sand and magma while preserving source water
- apply upward and downward entity movement, including surface behavior
- harden lava next to waterlogged bubble columns and let fire burn indefinitely on magma

## Testing

- `go test ./...`
- `go test -race ./server/block`
- `go vet ./...`


NopeNotDark (approved):
LGTM

PR #1332, opened 2026-07-16
Allow specifying private key for NetherNet listener

lactyy2:
In future versions (probably .40), clients will allow connecting to plain HTTP servers, but will prompt players to decide whether to trust the server's public key. That key is also used to sign the WebRTC connection's fingerprint (even when connecting to HTTPS server), which is required to secure the connection because Minecraft does not enable encryption even after the S2C/C2S handshake packets have been exchanged.

image

currently `go-nethernet` generates that key randomly, which means players may see that prompt every time the server restarts.

This PR adds a config field specifying the file path where the private key is stored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 
 # NetherNet server identity key
 identity.pem
+nethernet_identity.pem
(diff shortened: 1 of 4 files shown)


HashimTheArab:
Thanks for the PR!

PR #1334, opened 2026-07-19
implemet custom dimensions

FDUTCH:


HashimTheArab:
[P1] Dimension bounds are reversed. [server.go:449](https://github.com/df-mc/dragonfly/blob/b2f51ef2cdf952d7280c10f343774eb995424ca5/server/server.go#L449) sends {Max, Min}, while gophertunnel requires {Min, Max}. Every advertised custom dimension therefore has an invalid vertical range.

[P1] Custom dimension IDs cannot round-trip through the default player provider. [dimension.go:78](https://github.com/df-mc/dragonfly/blob/b2f51ef2cdf952d7280c10f343774eb995424ca5/server/world/dimension.go#L78) assigns IDs above 1000, but playerdb stores the dimension as uint8. Saving ID 1001 truncates it to 233, so the player reloads into the wrong dimension.

[P1] IDs are unstable and collisions are unchecked. IDs depend on registration order and are persisted in MCDB chunk keys. Reordering registrations can silently associate stored chunks with another dimension. Registering the same dimension—or world.Overworld—also overwrites its reverse mapping. Accept an explicit stable ID ≥1000 and reject duplicate IDs/dimensions.

[P2] (optional) Adding String() breaks existing implementations. [dimension.go:105](https://github.com/df-mc/dragonfly/blob/b2f51ef2cdf952d7280c10f343774eb995424ca5/server/world/dimension.go#L105) adds a method to the exported world.Dimension interface, breaking downstream types even when they do not use registration. Pass the protocol name as registration metadata or use a narrower interface.

FDUTCH:
> [P1] Dimension bounds are reversed. [server.go:449](https://github.com/df-mc/dragonfly/blob/b2f51ef2cdf952d7280c10f343774eb995424ca5/server/server.go#L449) sends {Max, Min}, while gophertunnel requires {Min, Max}. Every advertised custom dimension therefore has an invalid vertical range.
> 
looks like gopher docs are wrong, if I do min first and then max the client crashes

FDUTCH:
actually, I'm disappointed that instead of a real review, I'm getting a comment generated by AI, I don't want to waste my time arguing with AI

HashimTheArab:
.. they're literally valid findings? your code starts dimension ids at 1000+ while the db saves them as a uint8. the unstable/collisions one is valid too. the protocol was worth validating because gophertunnel claimed the opposite.

I reviewed it myself before i posted that comment and didn't notice any immediate issues so I ran an AI review and it listed many things i missed

you can dislike ai but this is a valid code review whats the issue here

FDUTCH:
> .. they're literally valid findings? your code starts dimension ids at 1000+ while the db saves them as a uint8. the unstable/collisions one is valid too. the protocol was worth validating because gophertunnel claimed the opposite.
> 
> I reviewed it myself before i posted that comment and didn't notice any immediate issues so I ran an AI review and it listed many things i missed
> 
> you can dislike ai but this is a valid code review whats the issue here

sure some of findings are valid, others... nvm... there is too many things to explain

HashimTheArab:
they are all valid except the last one which is debatable which is why i labeled it as optional

PR #1337, opened 2026-07-21
server/block: Fix outdated custom block components

MymaQc:
Dragonfly still sends custom blocks using components that are deprecated in current Bedrock versions. As a result, full-cube custom blocks may not render correctly, and their light properties may not be applied by the client.

Update the generated components to their modern equivalents:

- Replace [`minecraft:block_light_filter`](https://wiki.bedrock.dev/blocks/block-format-history#_1-19-10) with `minecraft:light_dampening`.
- Replace [`minecraft:block_light_emission`](https://wiki.bedrock.dev/blocks/block-format-history#_1-19-20) with `minecraft:light_emission` and encode the `0-15` value returned by `LightEmissionLevel()` as an NBT integer.
- Replace [`minecraft:unit_cube`](https://wiki.bedrock.dev/blocks/block-format-history#_1-20-60) with `minecraft:geometry` using the `minecraft:geometry.full_block` identifier.

These changes only affect the internal block components sent to the client. The public custom-block API remains unchanged.

### DIFF
--- a/server/internal/blockinternal/components.go
+++ b/server/internal/blockinternal/components.go
@@ -14,7 +14,7 @@ func Components(identifier string, b world.CustomBlock, blockID int32) map[strin
 	builder := NewComponentBuilder(identifier, components, blockID)
 	if emitter, ok := b.(block.LightEmitter); ok {
 		builder.AddComponent("minecraft:light_emission", map[string]any{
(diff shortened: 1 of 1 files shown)


HashimTheArab (approved):
LGTM, matches ms docs

RestartFU (approved):

PR #1342, opened 2026-07-22
Implement tridents

Superomarking:
This PR aims to fully implement tridents, needs testing.


server/session/handler_player_auth_input.go
```diff
@@ -125,6 +125,12 @@ func (h PlayerAuthInputHandler) handleInputFlags(flags protocol.Bitset, s *Sessi
 	if flags.Load(packet.InputFlagStartJumping) {
 		c.Jump()
 	}
+	if flags.Load(packet.InputFlagStartSpinAttack) {
+		c.StartSpinning()
```
RestartFU: [P1] `InputFlagStartSpinAttack` is client-controlled, so this lets a crafted packet enter or indefinitely refresh `spinning` without the ten-tick charge, water/rain check, or durability cost enforced by `Trident.Release`. Gate this transition on the server-validated release rather than authorizing it directly from the input flag.

server/entity/trident.go
```diff
@@ -0,0 +1,310 @@
...
+func (tridentType) DecodeNBT(m map[string]any, data *world.EntityData) {
+	conf := TridentBehaviourConfig{
+		Damage:            float64(nbtconv.Float32(m, "Damage")),
+		Item:              nbtconv.MapItem(m, "Trident"),
+		DisablePickup:     !nbtconv.Bool(m, "player"),
+		CollisionPosition: nbtconv.Pos(m, "StuckToBlockPos"),
+	}
+	if conf.Item.Empty() {
+		conf.Item = item.NewStack(item.Trident{}, 1)
+	}
```
RestartFU, line 289: [P1] Persisted tridents are rebuilt without `Owner` or the `returning` state, so a chunk unload/restart breaks Loyalty; a stuck loyalty trident then reaches `b.conf.Owner.Entity(...)` with a nil handle and panics, while a returning one no longer returns. Persist a resolvable owner/return state and make the owner lookups nil-safe.

server/item/trident.go
```diff
@@ -0,0 +1,139 @@
...
+type Trident struct{}
+
+// MaxCount ...
+func (Trident) MaxCount() int {
+	return 1
+}
+
+// AttackDamage ...
+func (Trident) AttackDamage() float64 {
+	return 9
```
RestartFU: [P2] `Stack.AttackDamage()` adds the base 1 damage to every `Weapon.AttackDamage()`, so returning 9 here makes trident melee hits deal 10 rather than Bedrock's 9. Return 8, matching the convention used by the existing weapon implementations.

server/entity/trident.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// Returning returns true if the trident is currently returning to its owner
+// as a result of the loyalty enchantment.
+func (b *TridentBehaviour) Returning() bool {
+	return b.returning
+}
+
+// Glint returns true if the trident stack carried by the entity is enchanted.
+func (b *TridentBehaviour) Glint() bool {
```
RestartFU, line 84: [P2] `Glint()` is defined on `TridentBehaviour`, but `parseEntityMetadata` only tests `e.H().Type()` for the `glint` interface and `addSpecificMetadata` never tests behaviours for it. Enchanted thrown tridents therefore never receive the enchanted flag; expose this through the metadata path that evaluates the behaviour.

RestartFU (commented):
Inline findings attached.

server/entity/trident.go
Superomarking: If anyone could help me with this would be appreciated!

PR #1346, opened 2026-07-23
server/entity: Match Bedrock arrow damage

HashimTheArab:
## Summary

- use Bedrock arrow impact damage instead of combining Bedrock launch velocity with Java damage scaling
- apply the Bedrock Power multiplier after critical damage
- preserve Power and critical state through arrow NBT
- correct crossbow arrow damage and launch velocity

## Root cause

Dragonfly launches bow arrows with Bedrock velocity (`force * 5`) but calculated impact damage using Java-style `ceil(base damage * velocity)`. A fully charged unenchanted bow therefore dealt 10-15 raw damage instead of Bedrock’s 9-10 critical range. Crossbows were affected more severely because their configured damage of 9 was multiplied by their launch velocity, and their launch speed was incorrectly set to 5.15 instead of Bedrock’s dedicated crossbow arrow power of 3.15.

This changes all arrow entities to one consistent Bedrock damage model: base damage plus 97% of current velocity, the 9-10 critical range, then Power scaling. Non-arrow projectiles keep their existing calculation.

The crossbow velocity was verified against the official BDS 1.26.36.1 binary: the crossbow constructor initializes `mDefaultArrowPower` to `3.15f`, and the arrow firing path passes that field directly to the projectile shoot call.

Reference: https://learn.microsoft.com/en-us/minecraft/creator/reference/source/vanillabehaviorpack_snippets/entities/arrow

## Validation

- `go test ./...`
- `git diff --check`


xSuperr:
LGTM

ethaniccc (dismissed):

PR #1347, opened 2026-07-23
Add HandleSetOnFire player handler

xSuperr:
Allows modification/cancellation of SetOnFire ticks called from any source (DF supports Lava/Fire blocks, Lightning, Arrows, Fire Aspect)

Verification:

go test ./...
git diff --check

### DIFF
--- a/server/player/handler.go
+++ b/server/player/handler.go
@@ -196,6 +196,7 @@ func (NopHandler) HandleAttackEntity(*Context, world.Entity, *float64, *float64,
 func (NopHandler) HandleExperienceGain(*Context, *int)                                     {}
 func (NopHandler) HandlePunchAir(*Context)                                                 {}
 func (NopHandler) HandleHurt(*Context, *float64, bool, *time.Duration, world.DamageSource) {}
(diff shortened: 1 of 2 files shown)


server/player/handler.go
```diff
@@ -50,6 +50,10 @@ type Handler interface {
 	// the original cause of the immunity frame. In this case, the damage is
 	// reduced but the player is still knocked back.
 	HandleHurt(ctx *Context, damage *float64, immune bool, attackImmunity *time.Duration, src world.DamageSource)
+	// HandleSetOnFire handles the player being set on fire by any source.
+	// The fire duration passed is after fire protection modifiers and may be changed
+	// by assigning to *ticks.
+	HandleSetOnFire(ctx *Context, ticks *int64)
```
RestartFU: shouldn't we translate to a time.Duration before and then translate to ticks again after (to use)?

it would be a much better UX

RestartFU: I don't like the idea of passing raw ticks

HashimTheArab (approved):

ethaniccc (approved):

PR #1344, opened 2026-07-22
Implement End Crystals

HashimTheArab:
This PR was split from https://github.com/df-mc/dragonfly/pull/1247 so End crystals can be reviewed independently from respawn anchors.

- Adds End crystal placement on obsidian and bedrock with vanilla air-space and entity-collision checks.
- Adds the End crystal entity, registration, metadata, persistence, direct/projectile damage handling, explosion chaining, and void-damage behaviour.
- Adds configurable explosion size through `entity.EndCrystalConfig`, retaining the vanilla radius of 6 by default.
- Prevents End-crystal explosions from affecting entities through liquid layers, matching the Bedrock `allow_underwater: false` default.

Verification:
- `go test ./...`
- `make lint`
- `git diff --check`

PR #1349, opened 2026-07-23
server: fix End crystal fire and placement parity

HashimTheArab:
Follow-up to #1344 and #1350.

## Summary
- only generate End crystal fire in the End when the crystal position is air and the block below is non-air
- preserve the normal scheduled fire lifecycle after placing crystal-generated fire
- match vanilla End crystal placement collision by checking the volume from the clicked support block through the air block above, including the placing player

## Verification
- go test ./...
- make lint
- git diff --check
- codex review against current upstream master (no actionable findings)


RestartFU (approved):

NopeNotDark (approved):
LGTM, Thanks for the PR!

ethaniccc (approved):

PR #1350, opened 2026-07-24
fix endcrystal behavior

FDUTCH:


JavierLeon9966:
End crystals in vanilla dont explode anything below if there is a explosion proof block below them

FDUTCH:
> End crystals in vanilla dont explode anything below if there is a explosion proof block below them

no way, let's think for a second, how does vanilla achives such behavior?
If you really think this is achieved by cliping explosion below some Y level, then I have some bad news for you

didntpot (approved):

HashimTheArab (approved):

PR #1358, opened 2026-07-25
server/player: Export NewEventContext function

FDUTCH:


RestartFU (dismissed):
just curious, what do you need that for?

FDUTCH:
> just curious, what do you need that for?

one of projects that I'm working on heavily relies on player.Context (for handling some stuff), it is useful cause you can call handler from other places

Dasciam (approved):

HashimTheArab:
can you show an example of this usecase

FDUTCH:
> can you show an example of this usecase

seriously, what the hell are you even asking? didn't I explain clearly how this could be useful? or are you trying to teach me how to code?
or do you want me to start explaining my architectural decisions to you?

why am I the only one who gets questions like this? am I really that bad?

HashimTheArab:
dude you need to calm down i just wanted to see if this is something that can be solved a better way instead of exposing that internal method

if you would rather have me not review your prs then ill let them sit, why are you being hostile over a simple question

FDUTCH:
> dude you need to calm down i just wanted to see if this is something that can be solved a better way instead of exposing that internal method
> 
> if you would rather have me not review your prs then ill let them sit, why are you being hostile over a simple question

okay simple answer: 
I want to call player handler methods

FDUTCH:
> why are you being hostile over a simple question?
because I get asked it too often, and I've already answered it above

HashimTheArab:
okay this might need a different api which is why i asked for an example
>  I've already answered it above
my question was different from the one above

https://github.com/df-mc/dragonfly/blob/95c383e0bb42336f8cbf94ce16fbc3b7b8673978/server/player/context.go#L7-L18

NewContext captures the transaction of the player, exposing it introduces footguns such as capturing the transaction of a stale player which will only panic once you call something like ctx.Block(..), and conflicts with "player.Context is valid only during its callback" which we documented. This was an explicit part of the new api design.

A better design to expose might be `NewEventContext(tx *world.Tx, p *Player) *Context`

FDUTCH:
> A better design to expose might be `NewEventContext(tx *world.Tx, p *Player) *Context`

we might call it `NewEventContext`, but passing world.Tx looks silly to me

FDUTCH:
okay, I kinda need this... @HashimTheArab do I need to change the docs or something?

HashimTheArab (dismissed):

HashimTheArab (dismissed):

didntpot (dismissed):

HashimTheArab (dismissed):

HashimTheArab (approved):

didntpot (approved):

PR #1363, opened 2026-07-28
run PlayerProvider.Save() asyncroniusly

FDUTCH:
This pr adds async player data save, so if you have slow player provider it won't cause lag spikes on your server

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -58,6 +58,9 @@ type Server struct {
 	// p holds a map of all players currently connected to the server. When they
 	// leave, they are removed from the map.
 	p map[uuid.UUID]*onlinePlayer
+	// saving holds the UUIDs of players whose data is still being saved. New
+	// sessions are rejected until their previous data has been persisted.
+	saving map[uuid.UUID]struct{}
 	// pwg is a sync.WaitGroup used to wait for all players to be disconnected
 	// before server shutdown, so that their data is saved properly.
 	pwg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -541,25 +541,32 @@ func (srv *Server) dimension(dimension world.Dimension) *world.World {
...
 		// might not be added to the players map yet. This is expected, but we
 		// need to be careful not to crash when this happens.
 		return
 	}
 
-	if tx != nil {
-		if err := srv.conf.PlayerProvider.Save(c.UUID(), c.(*player.Player).Data(), tx.World()); err != nil {
-			srv.conf.Log.Error("Save player data: " + err.Error())
+	data := c.(*player.Player).Data()
+	w := tx.World()
```
NopeNotDark, line 572: tx can be nil https://github.com/df-mc/dragonfly/blob/daad82a123a5d335bfcc1e83503e8593788b6b89/server/server.go#L555-L561

server/server.go
```diff
@@ -541,25 +541,32 @@ func (srv *Server) dimension(dimension world.Dimension) *world.World {
...
+	go func() {
+		if tx != nil {
```
RestartFU: this check is useless here, it should be done before ```	w := tx.World()``` (line 553)

server/server.go
```diff
@@ -541,25 +541,32 @@ func (srv *Server) dimension(dimension world.Dimension) *world.World {
...
+		} else {
+			srv.conf.Log.Error("Save player data: player's worlds closed before teardown; data not saved", "uuid", c.UUID())
 		}
-	} else {
-		srv.conf.Log.Error("Save player data: player's worlds closed before teardown; data not saved", "uuid", c.UUID())
-	}
-	srv.pwg.Done()
+
+		// removing player from the list after saving their data,
+		// to prevent possible race condition exploits.
```
RestartFU:
```suggestion
```
pretty self explainatory already

RestartFU (changes requested):

server/server.go
FDUTCH: good point

NopeNotDark (approved):
LGTM, Thanks for the PR!

PR #1365, opened 2026-08-01
server: match vanilla shulker box drop behaviour

FDUTCH:
This pr fixes shulkerbox breaking behavior.

### DIFF
--- a/server/block/shulker_box.go
+++ b/server/block/shulker_box.go
@@ -264,6 +264,12 @@ func (s ShulkerBox) CreativeDrops() []item.Stack {
 	return []item.Stack{item.NewStack(s, 1)}
 }
 
+// Explode destroys the shulker box and drops it with its contents intact.
+func (s ShulkerBox) Explode(_ world.ExplosionSource, pos cube.Pos, tx *world.Tx) {
+	tx.SetBlock(pos, nil, nil)
+	dropItem(tx, item.NewStack(s, 1), pos.Vec3Centre())
+}
+
 func (s ShulkerBox) MaxCount() int {
 	return 1
 }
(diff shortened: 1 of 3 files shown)


HashimTheArab (dismissed):

FDUTCH:
@HashimTheArab  shulker should be droped even if explosion block drop chance is 0

HashimTheArab (dismissed):

server/block/shulker_box.go
```diff
@@ -254,6 +254,22 @@ func (s ShulkerBox) BreakInfo() BreakInfo {
 	return newBreakInfo(2, alwaysHarvestable, pickaxeEffective, oneOf(s))
 }
 
+// CreativeDrops returns the drops produced when a creative player breaks the shulker box. Filled shulker boxes retain
```
DaPigGuy: Why can't this be done with a break handler

RestartFU: pretty sure it's to match bedrock natively in dragonfly

FDUTCH: good question

HashimTheArab: because it bypassed HandleBlockBreak

DaPigGuy: This should be inside of BreakInfo instead of being its own thing. 


> because it bypassed HandleBlockBreak

Then we should also switch over all the container blocks that drop their content on break. So don't call this `CreativeDrops`, and make explosions drop those items too instead of what's happening under here.

HashimTheArab: We'll keep this pr consistent with the existing behavior then and PR a separate fix for all the blocks

HashimTheArab (approved):
Restored the PR to the original break-handler implementation; the generalized drop pipeline will follow separately.

FDUTCH:
bump

PR #1368, opened 2026-08-04
server: don't send empty custom item entries

HashimTheArab:
`makeItemComponents` sizes the slice with `make([]protocol.ItemEntry, len(custom))` and then appends to it, so the item registry sent to clients begins with one zero-valued entry per custom item — empty name, runtime ID 0 — before the real ones.

```go
srv.customItems = make([]protocol.ItemEntry, 0, len(custom))
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -413,9 +413,9 @@ func (srv *Server) makeBlockEntries() {
 // at startup
 func (srv *Server) makeItemComponents() {
 	custom := world.CustomItems()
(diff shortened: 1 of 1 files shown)


DaPigGuy (dismissed):
could also do what `makeBlockEntries` is doing

HashimTheArab:
resolved

DaPigGuy (approved):

PR #1370, opened 2026-08-04
dragonfly: update for Minecraft 1.26.40

HashimTheArab:
## Summary

- Update Dragonfly session integration for Minecraft 1.26.40 / protocol 2168.
- Migrate crafting data to gophertunnel typed recipe vectors and name-based item descriptors.
- Adopt the new optional PlayerAuthInput payloads and InputFlags API.
- Move player-list and scoreboard discriminators to their per-entry wire representations.
- Encode LevelChunk request mode with a zero sub-chunk count and the new optional request limit; the legacy count sentinels are outside the protocol 2168 schema's `0..64` range.
- Refresh the embedded block registry, vanilla item registry, creative inventory, crafting data, and potion data for 1.26.40.
- Temporarily replace gophertunnel with HashimTheArab/gophertunnel pending the upstream protocol update.

## Data sources

- `block_states.nbt` was converted from CloudburstMC/Data commit `619483eb88140f46b8933506c6263861c0d8fa43`, the v2168 snapshot.
- `vanilla_items.nbt`, `creative_items.nbt`, `crafting_data.nbt`, and `potion_data.nbt` were captured from the exact Bedrock Dedicated Server 1.26.40.8 release using HashimTheArab/datagen.
- The BDS archive matched the SHA-256 published by EndstoneMC/bedrock-server-data.
- Smithing data was unchanged. Furnace data was retained because furnace recipes are no longer present in the 1.26.40 CraftingData packet.

## Dependency

This PR consumes the protocol update from Sandertv/gophertunnel#481 through a temporary fork replace. The replace should be removed once that dependency is available upstream.

## Verification

- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`
- Generated NBT loads through the Dragonfly block, item, creative, recipe, and session packages.
- No test files added.


EmirhanOlgn:
Heads up: with this branch a 1.26.40 client connects, spawns, and is then immediately disconnected with `ClientDisconnection-90` (error word `Block`), with nothing logged server-side.

The cause is not in this PR — it is a pre-existing bug that 1.26.40 turns fatal. `PlayerListEntry.BuildPlatform` is never set anywhere in Dragonfly, so every entry goes out with `0`, which is not a valid `protocol.DeviceOS` (the constants start at `DeviceAndroid = 1`, and gophertunnel already rejects a login outside 1-15). Older clients ignored it; 1.26.40 validates the field while decoding and drops the connection.

I opened #1376 against `master` with the two-line fix, so it can land independently of this PR.

For what it's worth, while tracking this down I verified the protocol side of the update is correct: I captured a `PlayerList` packet from an official BDS 1.26.40.8 and round-tripped it through `hashimthearab/gophertunnel` — byte-for-byte identical, including the per-entry variant tag, the `SerializedSkin` layout, `BEARGB` for `Player Color` and the lowercase `TrustedSkinFlag` string. Bisecting the join packets pointed at `PlayerList` add entries, and the only meaningful difference from the BDS capture was `Build Platform` (BDS echoed the client's device, Dragonfly sent `0`).

One other thing: this PR pins `hashimthearab/gophertunnel` at `e8804e953f72`, but the branch has two later commits — `5accc560` (fix sub-chunk result values) and `f5b1e240` (remove obsolete sub-chunk request sentinels). I did my testing on `f5b1e240` plus #1376, and with that combination the client spawns, streams chunks and plays normally.

PR #1372, opened 2026-08-05
server/item: export RepeatStacks over any countable stack

HashimTheArab:
`repeatStacks` splits a repeated crafting output across as many stacks as the item's max count needs. Nothing about that is specific to `item.Stack` — it needs a count, a max count, and a way to grow.

Being unexported and typed to `[]item.Stack`, a consumer holding its own stack type has to retype the arithmetic. In our Bedrock proxy the inventory overlay has exactly this function, line for line, over a type that embeds `item.Stack` alongside a network ID and NBT.

Converting instead of copying does not work: the split means outputs do not correspond 1:1 to inputs, so there is nothing to map the extra state back onto after the call.

```go
type Repeatable[T any] interface {
	Count() int
	MaxCount() int
	Grow(int) T
}

func RepeatStacks[T Repeatable[T]](items []T, repetitions int) []T
```

`item.Stack` satisfies it unchanged. It lives in `stack.go` next to the three methods it is built on. `handler_crafting` and `handler_stonecutter` call the exported form, so behaviour is unchanged.

## Validation
`go build ./...`, `go test ./server/... -count=1`. Happy to add tests here if you would like them — left out to keep the diff to the change itself.

### DIFF
--- a/server/item/stack.go
+++ b/server/item/stack.go
@@ -3,6 +3,7 @@ package item
 import (
 	"fmt"
 	"maps"
+	"math"
 	"reflect"
 	"slices"
 	"sort"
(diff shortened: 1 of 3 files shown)


HashimTheArab:
i can see a fair point of closing this one and keeping it on my fork

PR #1373, opened 2026-08-05
server/block/customblock: expose ambient occlusion intensity

HashimTheArab:
Trix’s #1426 fixed `ambient_occlusion` encoding and defaults on `master`. This PR now builds on that implementation and exposes the rest of the Bedrock API: callers can choose an ambient-occlusion intensity instead of only toggling the vanilla `0`/`1` values.

Microsoft documents `ambient_occlusion` as a float from `0.0` through `10.0` inclusive for format version 1.26.20. `WithAmbientOcclusionIntensity(float32)` accepts that complete range and rejects out-of-range or NaN values. Existing `WithAmbientOcclusion()` and `WithoutAmbientOcclusion()` behavior remains unchanged.

## Validation

- `go test ./... -count=1`
- `go vet ./...`


server/block/customblock/material.go
```diff
@@ -35,24 +39,35 @@ func (m Material) WithoutFaceDimming() Material {
...
+// vanilla strength. Use WithAmbientOcclusionIntensity to pick a different one.
 func (m Material) WithAmbientOcclusion() Material {
-	m.ambientOcclusion = true
+	m.ambientOcclusion = 1
+	return m
+}
+
+// WithAmbientOcclusionIntensity returns a copy of the Material with ambient occlusion applied
+// at the intensity given, where 0 disables it and 1 is the vanilla strength. The client
+// expects an intensity between 0 and 10.
```
DaPigGuy: between 0 and 1, not 0 and 10 right? should probably enforce this too

server/block/customblock/material.go
```diff
@@ -35,24 +39,35 @@ func (m Material) WithoutFaceDimming() Material {
...
 // Encode returns the material encoded as a map that can be sent over the network to the client.
 func (m Material) Encode() map[string]any {
 	return map[string]any{
-		"texture":           m.texture,
-		"render_method":     m.renderMethod.String(),
-		"face_dimming":      m.faceDimming,
+		"texture":       m.texture,
+		"render_method": m.renderMethod.String(),
+		"face_dimming":  m.faceDimming,
+		// The client reads this as an intensity, not a flag: a bool here is silently
```
DaPigGuy: AI context not needed

server/block/customblock/material.go
HashimTheArab: Enforced the 0–1 range in 073277e8. Out-of-range values now panic, matching the existing convention for bounded fluent APIs.

server/block/customblock/material.go
HashimTheArab: Removed the implementation-context comment in 073277e8.

HashimTheArab:
wait fixing, bedrock docs say 0-10, old code comment was wrong or outdated

server/block/customblock/material.go
HashimTheArab: Checked the current first-party Bedrock reference: as of format version 1.26.20, this must be a float from 0.0 through 10.0 inclusive. Updated the guard to 0–10 in e4fe3639. [Microsoft documentation](https://learn.microsoft.com/en-us/minecraft/creator/reference/content/blockreference/examples/blockcomponents/minecraftblock_material_instances?view=minecraft-bedrock-stable)

PR #1377, opened 2026-08-05
server/session: Fix client disconnect on player teleport

THEBOSS9345:
MovePlayer was sent with MoveModeTeleport but no TeleportData, which is written unconditionally. The client gets a teleport with no cause and errors out.

### DIFF
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -278,6 +278,12 @@ func (s *Session) ViewEntityTeleport(e world.Entity, position mgl64.Vec3) {
 			Yaw:             float32(yaw),
 			HeadYaw:         float32(yaw),
 			Mode:            packet.MoveModeTeleport,
+			// TeleportData is written whether or not it is set, so a
+			// MoveModeTeleport with no data leaves the client with a teleport
+			// it has no cause for.
+			TeleportData: protocol.Option(protocol.TeleportData{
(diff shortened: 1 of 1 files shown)


server/session/world.go
```diff
@@ -278,6 +278,12 @@ func (s *Session) ViewEntityTeleport(e world.Entity, position mgl64.Vec3) {
 			Yaw:             float32(yaw),
 			HeadYaw:         float32(yaw),
 			Mode:            packet.MoveModeTeleport,
+			// TeleportData is written whether or not it is set, so a
+			// MoveModeTeleport with no data leaves the client with a teleport
+			// it has no cause for.
```
TwistedAsylumMC: I don't think this comment is necessary, if you remove this then it's good to merge!

THEBOSS9345: Oh, okay i removed it

THEBOSS9345:
@TwistedAsylumMC, I'm not sure which `Cause` type to use since this function is used in so many different contexts, so I just used `Unknown`.

TwistedAsylumMC:
yeah I'm sure it's fine. not really sure what the client uses this for, hopefully just telemetry. If we see any issue with this then we can update it appropriately in the future 👍🏻

PR #1378, opened 2026-08-05
server/world: Add fall damage world setting

THEBOSS9345:
Adds Settings.FallDamage, mapped to the falldamage game rule already parsed from level.dat but previously unused. Player fall damage and ender pearl damage now respect it. Defaults to true.

### DIFF
--- a/server/entity/ender_pearl.go
+++ b/server/entity/ender_pearl.go
@@ -42,7 +42,9 @@ func teleport(e *Ent, tx *world.Tx, target trace.Result) {
 	if user, ok := owner.(teleporter); ok {
 		tx.PlaySound(user.Position(), sound.Teleport{})
 		user.Teleport(target.Position())
-		user.Hurt(5, FallDamageSource{})
+		if tx.World().FallDamage() {
+			user.Hurt(5, FallDamageSource{})
+		}
 	}
 }
 
--- a/server/player/player.go
+++ b/server/player/player.go
@@ -582,6 +582,9 @@ func (p *Player) fall(distance float64) {
 	if h, ok := b.(block.EntityLander); ok {
 		h.EntityLand(pos, p.tx, p, &distance)
 	}
+	if !p.tx.World().FallDamage() {
+		return
+	}
 	dmg := distance - 3
 	if boost, ok := p.Effect(effect.JumpBoost); ok {
 		dmg -= float64(boost.Level())
--- a/server/world/mcdb/leveldat/data.go
+++ b/server/world/mcdb/leveldat/data.go
@@ -247,6 +247,7 @@ func (d *Data) Settings() *world.Settings {
 		DefaultGameMode: mode,
 		Difficulty:      difficulty,
 		TickRange:       d.ServerChunkTickRange,
+		FallDamage:      d.FallDamage,
 	}
 }
 
@@ -268,6 +269,7 @@ func (d *Data) PutSettings(s *world.Settings) {
 	}
 	d.CurrentTick = s.CurrentTick
 	d.ServerChunkTickRange = s.TickRange
+	d.FallDamage = s.FallDamage
 	mode, _ := world.GameModeID(s.DefaultGameMode)
 	d.GameType = int32(mode)
 	difficulty, _ := world.DifficultyID(s.Difficulty)
--- a/server/world/settings.go
+++ b/server/world/settings.go
@@ -44,6 +44,9 @@ type Settings struct {
 	// TickRange is the radius in chunks around a Viewer that has its blocks and entities ticked when the world is
 	// ticked. If set to 0, blocks and entities will never be ticked.
 	TickRange int32
+	// FallDamage specifies if entities take damage from falling in this World. If set to false, falls are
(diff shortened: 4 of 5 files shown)


HashimTheArab (approved):

FDUTCH:
imo fall damage should be handled by the developer

THEBOSS9345:
> imo fall damage should be handled by the developer

You are not wrong, but we should give them basic access to turn it off or on without them editing the codebase.

RestartFU (changes requested):
I don't think we should have this, it should be handled through the player handler

THEBOSS9345:
> I don't think we should have this, it should be handled through the player handler

There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has

THEBOSS9345:
We have fall damage turn on by default we should then also have something for them to turn it off

RestartFU:
> > I don't think we should have this, it should be handled through the player handler
> 
> There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has

then perhaps the FallDamage source should contain some information about where it comes from

THEBOSS9345:
> > > I don't think we should have this, it should be handled through the player handler
> > 
> > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> 
> then perhaps the FallDamage source should contain some information about where it comes from 

For ex

THEBOSS9345:
> > > > I don't think we should have this, it should be handled through the player handler
> > > 
> > > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> > 
> > then perhaps the FallDamage source should contain some information about where it comes from 
> 
> For example?

RestartFU:
> > > > I don't think we should have this, it should be handled through the player handler
> > > 
> > > 
> > > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> > 
> > 
> > then perhaps the FallDamage source should contain some information about where it comes from
> 
> For ex

source entity

THEBOSS9345:
> > > > > I don't think we should have this, it should be handled through the player handler
> > > > 
> > > > 
> > > > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> > > 
> > > 
> > > then perhaps the FallDamage source should contain some information about where it comes from
> > 
> > For ex
> 
> source entity

You like me added more comments so it explains better where tge damage is coming from

RestartFU:
> > > > > > I don't think we should have this, it should be handled through the player handler
> > > > > 
> > > > > 
> > > > > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> > > > 
> > > > 
> > > > then perhaps the FallDamage source should contain some information about where it comes from
> > > 
> > > For ex
> > 
> > source entity
> 
> You like me added more comments so it explains better where tge damage is coming from

what?

what I meant is that the way you've done this is wrong, it should all be handled through the player handler, you're complaining that we can't know where the fall damage comes from, instead of adding a world method, we should add that missing information to the FallDamage source, just like AttackEntity's attacker field

THEBOSS9345:
Oh I get what you mean I am sorry.
Just woke up I was thinking should we add it in the player hanlder a new func for this won't that be much better or add a new field in player damage

RestartFU:
> Oh I get what you mean I am sorry. Just woke up I was thinking should we add it in the player hanlder a new func for this won't that be much better or add a new field in player damage

a new field in the Fall Damage damage source is the appropriate place for this

THEBOSS9345:
> > > > > > > I don't think we should have this, it should be handled through the player handler
> > > > > > 
> > > > > > 
> > > > > > There is no proper way for them to handle it they can't cancel fall damage from a ender pearl and it's something basic that every server software has
> > > > > 
> > > > > 
> > > > > then perhaps the FallDamage source should contain some information about where it comes from
> > > > 
> > > > For ex
> > > 
> > > source entity
> > 
> > You like me added more comments so it explains better where tge damage is coming from
> 
> what?
> 
> what I meant is that the way you've done this is wrong, it should all be handled through the player handler, you're complaining that we can't know where the fall damage comes from, instead of adding a world method, we should add that missing information to the FallDamage source, just like AttackEntity's attacker field

But should there not be a field in world settings for them to just have it off if we have it on mu default we dont know if tge user wants FallDamage on or off

RestartFU:
no, as I said, it should be handled in the player handler

THEBOSS9345:
Okay I'll fix it but I really think if we have something on my default the user should be able to turn it off

RestartFU:
> Okay I'll fix it but I really think if we have something on my default the user should be able to turn it off

no

THEBOSS9345:
Should we get a second opinion before I recoded it

HashimTheArab:
It does make more sense to just be a field in the fall damage damage-source, i approved it too quickly, the rest of the damage sources in dragonfly are handled this way too. You can add a field there to have full control over what you want to cancel in the player handler

PR #1379, opened 2026-08-06
server: Allow setting rotation when teleporting

THEBOSS9345:
Teleport only moved an entity, with no way to set the direction it faces on arrival. Adds an optional rotation to Player.Teleport and Ent.Teleport, applied before viewers are notified so it rides along in the existing movement packet.

### DIFF
--- a/server/entity/ent.go
+++ b/server/entity/ent.go
@@ -69,8 +69,14 @@ func (e *Ent) SetVelocity(v mgl64.Vec3) {
 
 // Teleport teleports the entity to the position given.
 func (e *Ent) Teleport(pos mgl64.Vec3) {
+	e.TeleportRotated(pos, e.Rotation())
+}
+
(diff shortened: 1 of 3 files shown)


HashimTheArab:
I would do an explicit rotation instead of making it variadic and document that callers who don't want to rotate should pass e.Rotation()

THEBOSS9345:
well they already do that i think better if we make so it they dont need to ad rotation its only option

HashimTheArab:
using variadic as optional in go is considered an anti pattern, im not completely against it but another option is to keep the existing teleport and add TeleportRotated

THEBOSS9345:
@HashimTheArab  i fixed it can you double check if it's right

HashimTheArab (dismissed):

THEBOSS9345:
also i am fixing a deadlock in chunk

HashimTheArab (approved):

PR #1380, opened 2026-08-06
server/player: Add HandleArmourEquip

THEBOSS9345:
Adds HandleArmourEquip(ctx *Context, slot int, before, after item.Stack) to
player.Handler. There is currently no way to react to a player equipping armour, so restrictions on what a player may wear are not possible without polling the armour inventory every tick.
It is called before the item is placed, so ctx.Cancel() prevents the equip rather than reverting it. The armour inventory already routes through inventory.Handler.HandlePlace on both paths a player can equip by, so no new
call sites are introduced.

### DIFF
--- a/server/player/armour_handler.go
+++ b/server/player/armour_handler.go
@@ -9,17 +9,30 @@ import (
 // Handler.HandleArmourEquip before an item is placed into one of the armour slots, so that equipping may be
 // prevented.
 type armourHandler struct {
-	inventory.NopHandler
-	p *Player
+	p    *Player
+	next inventory.Handler
+}
+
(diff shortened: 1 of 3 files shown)


HashimTheArab:
This is already possible in dragonfly using p.Armour().Handle(myHandler)

THEBOSS9345:
Oh, yeah, I made my own on my fork. I didn't notice the handler on the item, so I thought you guys had forgotten to add one.

FDUTCH:
even though there isn't much point to this pr, it still has some:
HandleArmourEquip captures tx  (which maybe useful, when working with inventory)

THEBOSS9345:
i was thinking the same thing
but i am not sure if dragonfly it self needs it
i do use this same type of system on my codebase for stuff

PR #1381, opened 2026-08-06
server/world: Fix deadlock when a chunk callback loads another chunk

THEBOSS9345:
`Loader.viewChunk` held `l.mu` while calling into viewer code. `addViewer` reads entity metadata, and `Player.Breathing` reads a block through `Tx.Liquid` if that block fell in a chunk with a pending request, `Tx.chunk` ran that request's callbacks inline, re entering `viewChunk` on the same `Loader` and goroutine and deadlocking on `l.mu`, which is not reentrant.

```
world.(*Loader).viewChunk      loader.go:127   <- l.mu.Lock, blocks
world.(*chunkRequest).signal   chunk_request.go:141
world.(*Tx).chunk              world.go:1363
player.(*Player).Breathing     player.go:3325
world.(*World).addViewer       world.go:1305
world.(*Loader).viewChunk      loader.go:145   <- already holds l.mu
```

The victim is the world owner, so the transaction queue is never drained again: no viewer receives chunks, neither `autoSave` nor `World.close` can run, and the process must be killed, losing everything since the last save. Introduced with #1216, so it affects master. A player rejoining at a border position hits it reliably.

Chunk callbacks are now dispatched through `Tx.Defer` in `signal` and in both inline paths of `loadChunkAsync`, so one can never run nested inside another, and `viewChunk` releases `l.mu` before calling into viewer code. `abort` also marks the request so `loadChunkAsync` discards it instead of appending callbacks to a request that will never complete.

Tested with `go build ./...`, `go vet`, and `go test ./server/world/`, plus a live server that reproduced the hang before the change and runs clean after it.

image

### DIFF
--- a/server/session/world.go
+++ b/server/session/world.go
@@ -278,6 +278,12 @@ func (s *Session) ViewEntityTeleport(e world.Entity, position mgl64.Vec3) {
 			Yaw:             float32(yaw),
 			HeadYaw:         float32(yaw),
 			Mode:            packet.MoveModeTeleport,
+			// TeleportData is written whether or not it is set, so a
+			// MoveModeTeleport with no data leaves the client with a teleport
+			// it has no cause for.
+			TeleportData: protocol.Option(protocol.TeleportData{
+				TeleportCause: packet.TeleportCauseUnknown,
+			}),
 		})
 		return
 	}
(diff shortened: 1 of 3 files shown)


HashimTheArab (dismissed):

THEBOSS9345:
@didntpot @RestartFU Can you guys also accept this one it's a major bug in rendering

PR #1384, opened 2026-08-07
server: Use the same NetherNet identity key path as vanilla BDS

lactyy2:
Bedrock Dedicated Server version 1.26.50.24 has added support for saving server identity keys via the serveridentity command.

This PR changes the default key path to `keys/server_identity_key.pem`, matching the path used by BDS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,5 @@
 /players/
 /resources/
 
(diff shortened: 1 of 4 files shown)


HashimTheArab:
Thanks!

PR #1388, opened 2026-08-07
Remove repanic

FDUTCH:
do anyone really likes this behavior? repanic happens on the critical path where server handles player's packets and instead of kicking the player who caused the panic, repanic just crashes the entire server


HashimTheArab:
maybe it should be configurable? this current behavior matches what dragonfly did before the new api refactor, so i dont think we should change it to silently consume panics which never happened before

FDUTCH:
> maybe it should be configurable? this current behavior matches what dragonfly did before the new api refactor, so i dont think we should change it to silently consume panics which never happened before

maybe you have an advice how to handle it? I agree that it would be better to make it configurable...

PR #1389, opened 2026-08-08
Remove GameTypeSurvivalSpectator

AkmalFairuz:


Dasciam:
Isn't this means,that we should send a native "spectator" game mode, since "survival spectator" is making the client forcibly disconnect?

AkmalFairuz:
Native spectator mode is broken, like we can't see the inventory HUD or use items.

Dasciam:
> Native spectator mode is broken, like we can't see the inventory HUD or use items.

Spectator mode wasn't intended for using items or interacting with items

Silentmussle913:
> Native spectator mode is broken, like we can't see the inventory HUD or use items.

isint that how spectator mode is supposed to work like literally or am i wrong

AkmalFairuz:
> > Native spectator mode is broken, like we can't see the inventory HUD or use items.
> 
> 
> 
> Spectator mode wasn't intended for using items or interacting with items
> 
> 

Yeah, it wasn't. The PR just makes the spectator gamemode have the same behavior as GameTypeSurvivalSpectator (you can see the inventory HUD)

Dasciam:
In spectator mode, the player cannot use items, interact with blocks, and in my opinion it would be the correct approach to make any world interaction getting cancelled by the client too

AkmalFairuz:
> In spectator mode, the player cannot use items, interact with blocks, and in my opinion it would be the correct approach to make any world interaction getting cancelled by the client too

It was already cancelled by the client since we sent AbilityData that doesn’t allow interacting with the world.

Dasciam:
> > In spectator mode, the player cannot use items, interact with blocks, and in my opinion it would be the correct approach to make any world interaction getting cancelled by the client too
> 
> It was already cancelled by the client since we sent AbilityData that doesn’t allow interacting with the world.

But still, when the player enters spectator mode, it sees that itself it's a floating head

Dasciam:
> > > In spectator mode, the player cannot use items, interact with blocks, and in my opinion it would be the correct approach to make any world interaction getting cancelled by the client too
> > 
> > 
> > It was already cancelled by the client since we sent AbilityData that doesn’t allow interacting with the world.
> 
> But still, when the player enters spectator mode, it sees that itself it's a floating head

Should I say, that the spectator sees other spectator players the same?

AkmalFairuz:
> But still, when the player enters spectator mode, it sees that itself it's a floating head

What do you think if we use native spectator gamemode when sending it to other players, but survival gamemode for the player itself?

AkmalFairuz:
By the way, the point of this PR is to provide a quick fix for the invalid gamemode enum error in v1.26.40, and the fix should be backward compatible with GameTypeSurvivalSpectator. We should support native spectator gamemode (maybe in another PR)

HashimTheArab (approved):

didntpot (approved):

PR #1391, opened 2026-08-11
server/session: bound the title and author a book is signed with

schphe:
### What happens and why

The page count, the page number and the text of a page are all checked against
the limits the game uses, but the title and author a book is signed with were
copied into the item unchecked. Both come from the client, and the only bound
on them is the maximum length of a string on the wire, so a client may sign a
book with a megabyte in each. The book is then kept in the player's data and
sent to every viewer of the item.

### Verification

Regression test fails without the fix: `signing was accepted with a title of 33 and an author of 5 bytes, want an error`.

**This is a remote crash, not only a validation gap.** Driven from a real client against a real server, signing a book with a megabyte of title takes the whole process down: the resulting `InventorySlot` packet reaches gophertunnel's NBT writer, which panics rather than returning an error for a string longer than 32767 bytes, on the session's write goroutine. Any client can do it.


xRookieFight:
I don't think this needs a new test

PR #1399, opened 2026-08-11
server/player: fix fall damage dealt in mid-air when moving past an overhang

schphe:
### What happens

Jump off a high platform and hug its side as you fall. At the moment you pass the platform's bottom face you take fall damage, in mid-air, touching nothing, and then keep falling.

### Why

`checkOnGround` extends the player's bounding box back over the movement of the tick, so ground moved past within a single tick is not missed:

```go
box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(deltaPos.Mul(-1.0))
```

`Extend` grows each axis on its own, which makes that the *union* of the boxes before and after the movement rather than the volume swept between them. While falling the movement is downwards, so the box grows **upwards by a full tick of falling** — and the check does not require the block it finds to be below the player at all.

Holding into the wall matters: the client resolves Y before X/Z, so on the tick the player's head clears the platform's bottom face the sideways movement stops being obstructed and the player slides into the platform's column. The grown box then covers a corner the player never occupied, inside the platform.

```
player box  [-0.4 7.7 0.2]..[0.2 9.5 0.8]    entirely below y=10, mid-air
check box   [-0.6 7.65 0.2]..[0.2 10.5 0.8]  onGround = true
```

The accumulated fall distance is turned into damage and reset, which is why the fall continues afterwards.

### What changed

The box is no longer extended upwards. Extending it back over horizontal movement, which is what this was added for in #1044, is unchanged.

### Verification

- Reported from live play on a real client.
- Regression test fails without the fix: `checkOnGround([0.2 -1 0]) at [-0.1 7.7 0.5] = true, want false`. It includes the sprint-down-stairs case from #1044 as a control, which still reports ground with the fix in place.


server/player/player.go
```diff
@@ -3001,7 +3001,13 @@ func (p *Player) checkEntitySteppers() {
 
 // checkOnGround checks if the player is currently considered to be on the ground.
 func (p *Player) checkOnGround(deltaPos mgl64.Vec3) bool {
-	box := Type.BBox(p).Translate(p.Position()).Extend(mgl64.Vec3{0, -0.05}).Extend(deltaPos.Mul(-1.0))
+	// The box is extended back over the movement of this tick, so that ground moved past within a single tick is not
+	// missed. It is never extended upwards: Extend grows every axis on its own, so a box extended back over a fall
+	// would cover a corner the player never occupied, and a block moved past sideways, such as the bottom edge of an
+	// overhang, would be read as ground.
```
TwistedAsylumMC: Could you shorten/remove this comment?

HashimTheArab (approved):

PR #1400, opened 2026-08-11
server/world: fix items and entities being duplicated by a chunk that is not saved

schphe:
### What happens and why

A chunk is only written to disk when it is marked as modified, and two things that change a chunk never mark it.

The contents of a container live behind a pointer its block entity holds, so moving items in or out of a chest never
touches the Column and never sets the flag. Taking items out of a chest and letting the chunk be saved and unloaded
leaves the chest on disk still holding them: the items come back when the chunk is loaded again, and the player still
has the ones they took. No crash or disconnect is needed, an ordinary save is enough.

An entity that crosses a chunk border is appended to the entities of the new chunk and deleted from those of the old
one, marking neither. The new chunk is written with the entity while the old chunk keeps the copy it was saved with, so the entity exists twice after a reload. Both chunks are marked now.

### What changed

A block entity that holds its contents behind a pointer now reports its own changes, through a new optional interface
alongside `NBTer`:

```go
type NBTChanger interface {
      NBTer
      Changed() bool
      ResetChanged()
}
```

`saveChunk` writes a Column when `modified` is set or any block entity reports a change, and clears both afterwards. The six containers that hold an `*inventory.Inventory` set the flag from the `SlotFunc` they already have: chest, barrel, hopper, shulker box, brewing stand and the smelters.

This also answers the question about `modified` never being set back to false. It cannot be cleared on its own: without the tracking above, a chunk would go clean while a chest's contents changed and then never be written again, which turns this from wasteful into lost items. The two have to land together.

### Verification

Reproduced against a running server with a real client attached and a leveldb provider: a chest filled, unloaded, read
back, emptied and unloaded again still held its contents on disk while the player kept the items they took.

Measured after the change: a chunk modified once is written once rather than on every save, a Column holding an
unchanged block entity is not written at all, and one holding a changed block entity is written exactly once.
Previously 25 chunks modified once produced 250 `StoreColumn` calls.

No tests are included, matching the other changes in this series.


server/world/block.go
```diff
@@ -169,6 +169,17 @@ type NBTer interface {
...
 
+// NBTChanger is an NBTer that keeps track of whether it changed since it was last written. The contents of a block
+// entity such as a container live behind a pointer it holds, so they are changed without the Column holding it being
+// touched, which would otherwise leave the Column unaware that it has to be written again.
+type NBTChanger interface {
+	NBTer
+	// Changed reports whether the block entity changed since ResetChanged was last called.
+	Changed() bool
+	// ResetChanged records the block entity as unchanged. It is called once it has been written.
+	ResetChanged()
```
TwistedAsylumMC: I'd probably prefix these methods with NBT otherwise looking a them in a block you have no idea what it's detecting as being changed

schphe: makes sense, done the changes

TwistedAsylumMC (dismissed):

server/world/world.go
```diff
@@ -1193,12 +1193,34 @@ func (w *World) save(f func(*Tx, ChunkPos, *Column)) execFunc {
...
-			w.conf.Log.Error("save chunk: "+err.Error(), "X", pos[0], "Z", pos[1])
+	if w.conf.ReadOnly || !c.changed() {
+		return
+	}
+	c.Compact()
+	if err := w.conf.Provider.StoreColumn(pos, w.conf.Dim, w.columnTo(c, pos)); err != nil {
+		w.conf.Log.Error("save chunk: "+err.Error(), "X", pos[0], "Z", pos[1])
+		return
+	}
+	c.modified = false
```
HashimTheArab: After a successful save clears `c.modified`, mutations to an entity that remains in the same chunk—such as movement, age, velocity, health, or entity-specific state—do not mark the column modified; `tickEntities` only does so when the entity crosses a chunk boundary. A later save or unload can therefore skip the chunk and retain the earlier entity snapshot. Could we add entity-level change tracking, or otherwise avoid clearing the flag while mutable entities are present?

server/block/smelter.go
```diff
@@ -23,12 +24,16 @@ type smelter struct {
...
 	experience        int
+
+	dirty atomic.Bool
 }
 
 // newSmelter initialises a new smelter with the given remaining, maximum, and cook durations and XP, and returns it.
 func newSmelter() *smelter {
 	s := &smelter{viewers: make(map[ContainerViewer]struct{})}
 	s.inventory = inventory.New(3, func(slot int, _, item item.Stack) {
+		s.dirty.Store(true)
```
HashimTheArab: This dirty flag only tracks inventory callbacks, but `tickSmelting` also changes encoded pointer-backed state such as cook/burn durations and experience without necessarily changing a slot or block state. The brewer has the same issue for duration and fuel, and `ResetExperience` is another path. After a save resets the flag, unloading mid-cycle can restore stale NBT. Could every encoded state mutation mark the block entity dirty?

server/block/barrel.go
```diff
@@ -31,14 +32,18 @@ type Barrel struct {
...
 }
 
 // NewBarrel creates a new initialised barrel. The inventory is properly initialised.
 func NewBarrel() Barrel {
 	m := new(sync.RWMutex)
 	v := make(map[ContainerViewer]struct{}, 1)
+	d := new(atomic.Bool)
 	return Barrel{
 		inventory: inventory.New(27, func(slot int, _, item item.Stack) {
+			d.Store(true)
```
HashimTheArab: `DecodeNBT` hydrates inventories through `InvFromNBT`, which calls `SetItem` and therefore fires this callback. A populated container loaded from disk is immediately considered dirty even when untouched, so its chunk is rewritten on the first save or unload. This applies to the other containers using the same pattern too. Could the flag be reset after hydration, or could hydration avoid reporting changes?

HashimTheArab (commented):
Three persistence and dirty-tracking concerns from a complete-branch review:

schphe:
done with the changes

PR #1408, opened 2026-08-11
server/block,server/session: fix a chest handing out a second copy of its contents

schphe:
### What happens and why

Pairing two chests into a double chest, and unpairing them again, replaced the
inventory of both with a copy of it. A player with either chest open is still
holding the inventory it was opened with, which is then no longer part of any
chest but is still live and still full: taking the items out of it gives a
second set of everything the chest held. Placing a chest beside an open one is
enough, no breaking is needed. The inventories are kept now, and unpairing
only replaces the function called when a slot changes.

Placing a chest that pairs with another also told the other chest it was
paired before knowing whether the placement succeeded. When it did not, the
neighbour was left paired with a position holding no chest, and resolving that
pair later wrote a chest block back into the world at a position a player had
already broken.

The window opened for a container is also checked against the block it was
opened for before it is used, as the other cases in that switch already are.

### Verification

Regression tests fail without the fix: `pairing replaced the inventory of both chests, leaving an open window holding a copy`, the same on unpairing, and `the window of a chest that was broken is still usable, so its contents can be taken twice`.

The window check closes two further cases that were confirmed against a real session driving real ItemStackRequest packets: a shulker box broken while open, and one half of a double chest broken while open. Both leave the inventory the window holds detached from the world but still full, and `ContainerLevelEntity` was the one branch of that switch with no check that the block was still there. Nothing sets the container as closed when the block is broken, so the window remained usable and everything in it could be taken a second time.


TwistedAsylumMC:
Using the live inventories in pair() means left/right keep their original SlotFuncs while the merged double also notifies, so viewers will now get ViewSlotChange twice per change under different slot indices

PR #1409, opened 2026-08-11
server/world: fix a World with no provider raining from tick 1 and dereferencing a nil Biome

schphe:
### What happens and why

The default Settings leave the rain and thunder counters at zero. Those
counters are decremented every weather tick and toggle the weather they
control when they reach zero, so the first tick, which Config.New runs itself,
turns both rain and thunder on. The world is then in a thunderstorm from the
moment it is created, and lightning starts striking within seconds.

This is not limited to worlds built directly from world.Config. The server
package substitutes world.NopProvider when Config.WorldProvider is nil, and
that provider returns these same default Settings, so a server built from a
server.Config without a world provider starts every one of its dimensions in a
storm. A server backed by mcdb is unaffected, but only by accident: a fresh
level.dat records rain and lightning as already falling with the same zero
timers, so its first tick toggles them off instead. Vanilla resets these
counters to 12,000-179,999 ticks whenever weather stops, so they now start in
that range.

Biomes register themselves in the init of server/world/biome, which is only
reached by importing the server package. A World built straight from a
world.Config therefore has an empty biome registry, and both the biome lookup
and the out of bounds fast path returned a nil Biome that every caller then
dereferenced. rainingAt does so on the lightning path, which the world tick
reaches at random, killing the world's goroutine within half a minute. Code
that imports the server package registers every vanilla biome and only sees
this for a chunk holding a biome ID dragonfly does not know, such as one saved
by a newer version of the game. Both now fall back to an unknown biome, which
the lookup already logs.

### Verification

Regression tests fail without the fix: a newly created world reports `Raining = true`, and `biome(...) = nil, want a biome` followed by a nil dereference on the lightning path.


server/world/biome.go
```diff
@@ -60,8 +60,24 @@ func Biomes() []Biome {
...
+type unknownBiome struct{}
+
+func (unknownBiome) Temperature() float64    { return 0.5 }
+func (unknownBiome) Rainfall() float64       { return 0 }
+func (unknownBiome) Depth() float64          { return 0.1 }
+func (unknownBiome) Scale() float64          { return 0.1 }
+func (unknownBiome) WaterColour() color.RGBA { return color.RGBA{R: 0x44, G: 0xaf, B: 0xf5, A: 0xff} }
+func (unknownBiome) Tags() []string          { return nil }
+func (unknownBiome) String() string          { return "unknown" }
+func (unknownBiome) EncodeBiome() int        { return 0 }
```
TwistedAsylumMC: Is this meant to have the same ID as ocean?

server/world/settings.go
```diff
@@ -46,14 +46,18 @@ type Settings struct {
 	TickRange int32
 }
 
-// defaultSettings returns the default Settings for a new World.
+// defaultSettings returns the default Settings for a new World. The rain and thunder times are not left at 0: those
+// counters toggle the weather they control when they reach 0, so a World would start in a thunderstorm on its first
+// tick.
```
TwistedAsylumMC: Additional comments are not necessary

server/world/settings.go
```diff
@@ -46,14 +46,18 @@ type Settings struct {
...
 func defaultSettings() *Settings {
 	return &Settings{
 		Name:            "World",
 		DefaultGameMode: GameModeSurvival,
 		Difficulty:      DifficultyNormal,
 		TimeCycle:       true,
 		WeatherCycle:    true,
+		RainTime:        12000,
+		ThunderTime:     12000,
```
TwistedAsylumMC: Should these not be randomised from the start?

TwistedAsylumMC (dismissed):

server/world/biome.go
schphe: ah no, now it keeps the ids same

server/world/settings.go
schphe: right, it's randomised now, following same expression weather.go re-seeds with when weather stops

RestartFU (approved):

HashimTheArab (approved):

PR #1411, opened 2026-08-11
server/session: fix closing a session without a transaction panicking

schphe:
### What happens and why

Closing a session without a transaction skips removing the controllable from
its world, as there is no transaction to remove it with, but still closed the
handle of that entity. Closing a handle that is still in a world panics, so a
session closed this way takes the goroutine that closed it down with it.

The handle is only closed when the entity was removed. Leaving it open is what
the rest of that branch already does with the chunk loader and the container.

### Verification

Reproduced in process against a real session: `Session.Close(nil, c)` panics with `cannot add entity to new world before removing from old world` when the controllable is still in a world.

### DIFF
--- a/server/session/session.go
+++ b/server/session/session.go
@@ -332,8 +332,7 @@ func (s *Session) close(tx *world.Tx, c Controllable) {
 	// early.
 	if tx != nil {
 		tx.RemoveEntity(c)
-		// The handle is only closed once the entity has been removed from its world: closing one that is still in a
-		// world panics. Without a transaction the entity cannot be removed, so its handle is left alone.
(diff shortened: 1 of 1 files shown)


server/session/session.go
```diff
@@ -332,8 +332,10 @@ func (s *Session) close(tx *world.Tx, c Controllable) {
 	// early.
 	if tx != nil {
 		tx.RemoveEntity(c)
+		// The handle is only closed once the entity has been removed from its world: closing one that is still in a
```
HashimTheArab: shorten or remove the comment

HashimTheArab (approved):

PR #1415, opened 2026-08-11
server/entity: fix a firework damaging players through walls

schphe:
### What happens and why

The check for a clear line between an exploding firework and a victim treats
any hit as a clear line. trace.Perform reports a hit for the block it stopped
at as well as for an entity, so a wall between the two satisfied it and the
damage was dealt through the wall. Only a hit that is not a block counts now.

### Verification

Reproduced in process: a firework exploding on the far side of a wall damages a player through it. `trace.Perform` reports a hit for the block it stopped at (`server/block/cube/trace/result.go:32`) as well as for an entity, so the guard that was meant to require a clear line accepted the wall as one.

No unit test: the fix is a type check on a result the surrounding code already has, and reaching it needs a firework entity, a victim and a wall in a live world.

### DIFF
--- a/server/entity/firework_behaviour.go
+++ b/server/entity/firework_behaviour.go
@@ -1,6 +1,7 @@
 package entity
 
 import (
+	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/block/cube/trace"
 	"github.com/df-mc/dragonfly/server/item"
 	"github.com/df-mc/dragonfly/server/world"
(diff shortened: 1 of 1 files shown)


TwistedAsylumMC:
I don't think the fix is quite right. Perform truncates the ray at the block it hits (end = hit.Position()) before it tests entities, so an entity standing between the firework and the wall still produces an EntityResult and the victim behind the wall takes damage anyway. Perform is the wrong tool here: it answers "what did this ray hit first", not "is there a block in between". Could this test block occlusion only, via TraverseBlocks/BlockIntercept? ExplosionConfig.exposure already does block-only line-of-sight for the same purpose, so there's a pattern to follow.

PR #1417, opened 2026-08-11
server/item/inventory: fix data race between a slot change and SlotFunc or Close

schphe:
### What happens and why

setItem returns a closure that its callers run after the inventory has been
unlocked, so that the function called on a slot change does not run while the
inventory is locked. That closure read inv.f when it ran rather than when it
was built, while SlotFunc and Close write inv.f under the lock, so a slot
being written concurrently with either was a data race on a type documented as
safe for concurrent use.

The function is now read while the lock is still held. A change already in
progress therefore calls the function that was installed when it started,
which is the same guarantee the lock gives every other field.

### Verification

Regression test reports `WARNING: DATA RACE` without the fix, under `go test -race`. Note the repository's CI does not run with the race detector, so this test only earns its place if run that way.


server/item/inventory/inventory.go
```diff
@@ -376,8 +376,11 @@ func (inv *Inventory) setItem(slot int, it item.Stack) func() {
 	}
 	before := inv.slots[slot]
 	inv.slots[slot] = it
+	// The function is read here rather than in the closure below: the closure runs after the inventory is unlocked,
```
HashimTheArab: shorten this pls or remove

HashimTheArab (approved):

PR #1424, opened 2026-08-11
server/block: fix a composter using the held item when its bone meal is taken

schphe:
### What happens and why

Collecting the bone meal from a full composter left `Activate` returning false, so the click that emptied it also went on to use or place whatever the player was holding. One click did two things.

Separately, the item a composter drops was built from the block, so the stack carried the level the composter had reached rather than describing an empty one. An empty composter is dropped now, as in vanilla.

### Verification

Regression test fails without the fix: `composter at level 8 dropped one at level 8, want 0`.

**The drop half is not exploitable end to end, and that is worth stating plainly.** As with the other blocks whose drops carried their state, `entity.NewItem` round trips the dropped stack through NBT and `EncodeItem` writes the empty form, so a player never receives a composter that is still full; a measured run produced one bone meal per refill. What is wrong is the `Drops` implementation itself, which anything reading it directly would act on.

The `Activate` half is unaffected by that round trip and is reachable in ordinary play.


server/block/composter.go
```diff
@@ -82,7 +82,8 @@ func (c Composter) SideClosed(cube.Pos, cube.Pos, *world.Tx) bool {
 
 // BreakInfo ...
 func (c Composter) BreakInfo() BreakInfo {
-	return newBreakInfo(0.6, alwaysHarvestable, axeEffective, oneOf(c)).withBreakHandler(func(pos cube.Pos, tx *world.Tx, u item.User) {
+	// The composter dropped is empty: an item carrying the level would place a composter that is still full.
+	return newBreakInfo(0.6, alwaysHarvestable, axeEffective, oneOf(Composter{})).withBreakHandler(func(pos cube.Pos, tx *world.Tx, u item.User) {
```
TwistedAsylumMC: Is this change actually necessary? I thought we strip block data anyways on items. Either way, please remove this comment

schphe: BreakBlock passes &drops to HandleBlockBreak before any of that happens, so if a server collects drops straight into the inventory there, the composter comes back at level 8 and places down still full.

DaPigGuy (approved):

PR #1425, opened 2026-08-11
Fix custom block ambient occlusion encoding

TrixNEW:
﻿## What changed

Encode custom block `minecraft:material_instances` ambient occlusion as `float32(0)` or `float32(1)` while retaining the existing boolean-facing `Material` API.

Adds a regression test covering enabled and disabled ambient occlusion.

## Why

Since block format version 1.26.20, Bedrock no longer accepts boolean values for `ambient_occlusion`; the component expects a float in the 0.0-10.0 range. Sending the old boolean value can disconnect current clients with `ClientDisconnection-90` and the `Block` codeword while processing custom block definitions.

Reference: https://learn.microsoft.com/en-us/minecraft/creator/reference/content/blockreference/examples/blockcomponents/minecraftblock_material_instances?view=minecraft-bedrock-stable

## Validation

- `go test ./server/block/customblock`
- Built a Dragonfly application with six registered custom blocks against this checkout
- Application custom-block tests and full `go build ./...` pass


TrixNEW:
Closing this to clean up the branch. I'll open a replacement shortly.

PR #1430, opened 2026-08-13
session.go: Implement batch packet reading

AkmalFairuz:
This PR improves world transaction performance by batching inbound packets before processing them.

Previously, each packet required a separate world transaction. Packets are now read into a buffered channel, and all immediately available packets are processed in order within a single transaction. This reduces transaction scheduling and synchronisation overhead when multiple packets arrive close together.

### DIFF
--- a/server/session/session.go
+++ b/server/session/session.go
@@ -418,41 +418,40 @@ func (s *Session) handlePackets() {
 		}
 	}()
 	for {
-		select {
-		case first, ok := <-readPackets:
-			if !ok {
-				return
-			}
-			packets := []packet.Packet{first}
-			for {
-				var exit bool
-				select {
-				case pk, ok := <-readPackets:
-					if !ok {
-						return
-					}
-					packets = append(packets, pk)
-				default:
-					exit = true
-				}
-				if exit {
-					break
-				}
-			}
-			if err := s.withControllable(context.Background(), func(tx *world.Tx, c Controllable) error {
-				for _, pk := range packets {
-					if err := s.handlePacket(pk, tx, c); err != nil {
-						return err
-					}
-				}
-				return nil
-			}); err != nil {
-				if sessionOwnerStopped(err) {
+		first, ok := <-readPackets
+		if !ok {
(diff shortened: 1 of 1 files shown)


server/session/session.go
```diff
@@ -406,20 +406,53 @@ func (s *Session) handlePackets() {
...
-			if sessionOwnerStopped(err) {
+	}()
+	for {
+		select {
+		case first, ok := <-readPackets:
+			if !ok {
+				return
+			}
+			packets := []packet.Packet{first}
+			for {
```
FDUTCH: why not procsess all packets in this loop, instead of collecting them in a slice?

FDUTCH: no, I mean just put this loop inside withControllable

AkmalFairuz: Mainly to avoid the channel synchronization overhead while holding the transaction. I collect the packets first, then process the whole batch inside withControllable

FDUTCH: I bet that creating & growing upprepared slice creates more overhead than reading channel

AkmalFairuz: Reading from the channel inside a transaction would block other transactions. I agree that creating and growing unprepared slices adds overhead, I just fixed that.

FDUTCH: > Reading from the channel inside a transaction would block other transactions. I agree that creating and growing unprepared slices adds overhead, I just fixed that.

okay, I where dumb, comparing slice gown outside transaction and channel reading inside... 
I'm okay with the implementation after optimisation

FDUTCH: > I'm okay with the implementation after optimisation
but I'm not okay with the timer...

Dasciam:
This batch reading isn't quite synchronised to a real client's batching. I think instead of internal batching, we would need to modify the gophertunnel library to handle client's batches with the for example `ReadBatch` func. I think it's more convenient

TrippleAWap:
Duplicate of https://github.com/Sandertv/gophertunnel/issues/300.

PR #1436, opened 2026-08-16
Implement mycelium

MEMOxiiii:
Adds mycelium, which spreads onto nearby dirt and reverts to dirt in low light through a spreadDirt helper extracted unchanged from Grass.RandomTick — vanilla implements both blocks with the same class.

636669734-2a6c6d2f-28f3-47f8-a0fb-09561246c416


server/block/dirt.go
```diff
@@ -59,6 +62,36 @@ func (d Dirt) EncodeBlock() (string, map[string]any) {
...
+		spreadPos := pos.Add(cube.Pos{x - 1, y - 3, z - 1})
+		if tx.Light(spreadPos.Side(cube.FaceUp)) < 4 {
+			continue
+		}
+		if dirt, ok := tx.Block(spreadPos).(Dirt); !ok || dirt.Coarse {
+			continue
+		}
+		tx.SetBlock(spreadPos, b, nil)
+	}
+}
```
HashimTheArab: Parameterising the vertical offset keeps the extraction while preserving each block's Bedrock range (grass -3, mycelium -2).

This also restores the four inline comments the move dropped, reworded away from "grass" - now that the thresholds are read outside `Grass`, the reason for 4 vs 9 is less obvious from the call site.

```suggestion
// spreadDirt turns the block passed into dirt in low light and otherwise spreads it onto nearby dirt. It is
// the random tick behaviour shared by grass blocks and mycelium. minY is the lowest vertical offset a spread
// attempt may target, spanning five blocks upwards from there.
func spreadDirt(b world.Block, pos cube.Pos, tx *world.Tx, r *rand.Rand, minY int) {
	aboveLight := tx.Light(pos.Side(cube.FaceUp))
	if aboveLight < 4 {
		// The light above the block is too low: The block turns to dirt.
		tx.SetBlock(pos, Dirt{}, nil)
		return
	}
	if aboveLight < 9 {
		// Don't attempt to spread if the light level is lower than 9.
		return
	}

	// A single uint32 is enough as only 28 bits are needed, 7 per iteration.
	n := r.Uint32()

	// Four attempts to spread to another block.
	for range 4 {
		x, y, z := int(n)%3, int(n>>2)%5, int(n>>5)%3
		n >>= 7

		spreadPos := pos.Add(cube.Pos{x - 1, y + minY, z - 1})
		// Don't spread to locations where dirt is exposed to hardly any light.
		if tx.Light(spreadPos.Side(cube.FaceUp)) < 4 {
			continue
		}
		if dirt, ok := tx.Block(spreadPos).(Dirt); !ok || dirt.Coarse {
			continue
		}
		tx.SetBlock(spreadPos, b, nil)
	}
}
```

server/block/grass.go
```diff
@@ -47,36 +47,7 @@ func (g Grass) SoilFor(block world.Block) bool {
...
-		if tx.Light(spreadPos.Side(cube.FaceUp)) < 4 {
-			continue
-		}
-		b := tx.Block(spreadPos)
-		if dirt, ok := b.(Dirt); !ok || dirt.Coarse {
-			continue
-		}
-		tx.SetBlock(spreadPos, g, nil)
-	}
+	spreadDirt(g, pos, tx, r)
```
HashimTheArab:
```suggestion
	spreadDirt(g, pos, tx, r, -3)
```

server/block/mycelium.go
```diff
@@ -0,0 +1,47 @@
...
+	switch block.(type) {
+	case ShortGrass, Fern, DoubleTallGrass, Flower, DoubleFlower, NetherSprouts, PinkPetals, SugarCane, DeadBush, BambooSapling, Bamboo:
+		return true
+	}
+	return false
+}
+
+// RandomTick ...
+func (m Mycelium) RandomTick(pos cube.Pos, tx *world.Tx, r *rand.Rand) {
+	spreadDirt(m, pos, tx, r)
```
HashimTheArab: `nextInt(5) - 2` in `MyceliumBlock::randomTick`, against `nextInt(5) - 3` in `GrassBlockBase::randomTick`.

```suggestion
	spreadDirt(m, pos, tx, r, -2)
```

server/block/mycelium.go
```diff
@@ -0,0 +1,47 @@
+package block
+
+import (
+	"math/rand/v2"
+
+	"github.com/df-mc/dragonfly/server/block/cube"
+	"github.com/df-mc/dragonfly/server/world"
+)
+
+// Mycelium is a variant of dirt that generates naturally in mushroom fields biomes.
```
HashimTheArab: Nit: there is a single mushroom fields biome, and mycelium is what the biome's surface is made of rather than something that generates within it.

```suggestion
// Mycelium is a variant of dirt that covers the surface of the mushroom fields biome.
```

server/block/dirt.go
```diff
@@ -59,6 +62,36 @@ func (d Dirt) EncodeBlock() (string, map[string]any) {
 	return "minecraft:dirt", nil
 }
 
+// spreadDirt turns the block passed into dirt in low light and otherwise spreads it onto nearby dirt. It is
+// the random tick behaviour shared by grass blocks and mycelium.
+func spreadDirt(b world.Block, pos cube.Pos, tx *world.Tx, r *rand.Rand) {
```
HashimTheArab: The vertical range is the one thing the two blocks do not share, so it has to be a parameter.

```suggestion
// spreadDirt turns the block passed into dirt in low light and otherwise spreads it onto nearby dirt. It is
// the random tick behaviour shared by grass blocks and mycelium. minY is the lowest vertical offset a spread
// attempt may target.
func spreadDirt(b world.Block, pos cube.Pos, tx *world.Tx, r *rand.Rand, minY int) {
```

server/block/dirt.go
```diff
@@ -59,6 +62,36 @@ func (d Dirt) EncodeBlock() (string, map[string]any) {
...
+	}
+
+	// A single uint32 is enough as only 28 bits are needed, 7 per iteration.
+	n := r.Uint32()
+
+	for range 4 {
+		x, y, z := int(n)%3, int(n>>2)%5, int(n>>5)%3
+		n >>= 7
+
+		spreadPos := pos.Add(cube.Pos{x - 1, y - 3, z - 1})
```
HashimTheArab:
```suggestion
		spreadPos := pos.Add(cube.Pos{x - 1, y + minY, z - 1})
```

server/block/grass.go
HashimTheArab: One up, three down.

```suggestion
	spreadDirt(g, pos, tx, r, -3)
```

server/block/mycelium.go
HashimTheArab: Two up, two down.

```suggestion
	spreadDirt(m, pos, tx, r, -2)
```

server/block/mycelium.go
HashimTheArab: Nit: there is one mushroom fields biome, and mycelium is its surface rather than something generating in it.

```suggestion
// Mycelium is a variant of dirt that covers the surface of the mushroom fields biome.
```

server/block/mycelium.go
MEMOxiiii: thx for the review !

applied all but the mycelium range, the wiki says "one space above, one sideways, or three down" for it, so do you have a source for two up, two down ?

HashimTheArab: the wiki is wrong there, those are the values for grass, for mycelium i can confirm its 2 up 2 down

MEMOxiiii: done

HashimTheArab (approved):

PR #1452, opened 2026-08-30
 server/entity: share the living-entity foundation with the player

EinBexiii:
This is the substance of #1450, with the bug fixes and the UI change pulled out into their own PRs.

A player is an entity, but almost nothing that makes an entity alive sits on the entity side. Damage, armour, effects, air, fall handling, fire ticks, knockback and attack immunity all live on Player, and it's the same logic anything else alive would need, so anything that isn't a player has to duplicate it and the two copies drift apart. This moves that into shared pieces and has the player use them like anything else would: LivingEnt and LivingState as a base, AirSupply and AttackImmunity as components to embed, and FinalDamage, KnockBackVector, ExplosionDamage and Fall as plain functions. Living splits into Damageable and EffectBearer, and player.go gets a good deal shorter along the way.

HandleEntityHurt comes with it because it can't go in alone. There's no single place entity damage passes through upstream, since Hurt sits on Living and every entity implements it itself, so the hook would have nothing to hook into. The shared damage path is what gives it one.

### DIFF
--- a/server/entity/doc.go
+++ b/server/entity/doc.go
@@ -1,0 +1,72 @@
+// Package entity implements the entities that inhabit a world, along with the machinery to build new ones,
+// both inside this package and in external modules.
+//
+// # Architecture
+//
+// Three types work together to form an entity:
+//
+//   - A world.EntityType is the identity of an entity kind. It creates the entity's live
+//     form through its Open method and encodes/decodes its data to NBT.
+//   - A world.EntityHandle is the persistent form of one entity. It owns the entity's world.EntityData and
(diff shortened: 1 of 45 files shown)


server/entity/doc.go
Silentmussle913, line 1: needed?

razzmanpoo: yes bro

EinBexiii: theres already a doc.go in block, cmd, event and a couple of others, so its not like im adding something the repo doesnt already do.

PR #1454, opened 2026-08-30
server/entity: Push an item out of a block it is stuck in

EinBexiii:
see title

### DIFF
--- a/server/entity/push.go
+++ b/server/entity/push.go
@@ -1,29 +1,70 @@
 package entity
 
 import (
+	"math"
+
 	"github.com/df-mc/dragonfly/server/block/cube"
 	"github.com/df-mc/dragonfly/server/world"
 	"github.com/go-gl/mathgl/mgl64"
 )
 
-// pushSpeed is how fast an entity inside a block rises out of it, in blocks per
-// tick. Measured against BDS.
+// pushSpeed is how fast an entity inside a block is pushed out of it, in
+// blocks per tick. Measured against BDS.
 const pushSpeed = 0.21
 
-// pushOutOfBlock lifts an entity that ended up inside a block out of it and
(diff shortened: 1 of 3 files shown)


server/entity/push.go
```diff
@@ -0,0 +1,49 @@
...
+	up, prevVel := mgl64.Vec3{0, pushSpeed, 0}, e.data.Vel
+	return &Movement{v: tx.Viewers(pos), e: e,
+		pos: pos.Add(up), vel: up, dpos: up, dvel: up.Sub(prevVel),
+		rot: e.data.Rot, onGround: false,
+	}
+}
+
+// obstructed reports whether any block intersects the bounding box passed.
+func obstructed(tx *world.Tx, box cube.BBox) bool {
+	min, max := cube.PosFromVec3(box.Min()), cube.PosFromVec3(box.Max())
```
FDUTCH: use cube.Range3D

server/entity/push.go
```diff
@@ -0,0 +1,49 @@
...
+// entity at the first block it is not yet inside. BDS lifts it straight up
+// through whatever is above it instead.
+func pushOutOfBlock(e *Ent, tx *world.Tx) *Movement {
+	pos := e.Position()
+	if !obstructed(tx, e.H().Type().BBox(e).Translate(pos)) {
+		return nil
+	}
+	// The velocity is kept so the entity's own physics carry it on once it is
+	// out, which is how BDS has it leave on an arc rather than stopping dead.
+	up, prevVel := mgl64.Vec3{0, pushSpeed, 0}, e.data.Vel
```
FDUTCH: in vanilla, items can be pushed to the side (it is probably gets pushed to the nearest side)

PR #1456, opened 2026-09-10
Fix registry cloning panic in compound block hashes

HashimTheArab:
Finalizing `world.DefaultBlockRegistry.Clone()` or `world.NewBlockRegistry()` before the default registry panics with duplicate block hashes. Generated stairs, slab, and wall hashes encode their material through `world.BlockHash()`, which reads the default registry's uninitialized bit width. Oak and spruce stairs then collide even without calling `block.NextHash()`.

Generate a registry-independent material hash with separate 16-bit type and state fields and checked bounds. This also replaces the wall-only state bound.

Validation:
- Reproduced the reported panic before the fix and verified the standalone reproduction succeeds afterward.
- `go test ./...`
- `make lint` (0 issues)
- `go generate ./server/block/` followed by `git diff --exit-code`

### DIFF
--- a/cmd/blockhash/main.go
+++ b/cmd/blockhash/main.go
@@ -142,7 +142,8 @@ func (b *hashBuilder) writeBlockHash(w io.Writer) {
 func hashBlock(b world.Block) uint64 {
 	base, state := b.Hash()
 	if base > 0xffff || state > 0xffff {
-		panic("material block hash exceeds 16-bit type or state")
+		name, _ := b.EncodeBlock()
+		panic("hash of block " + name + " exceeds 16-bit type or state")
 	}
 	return base | state<<16
 }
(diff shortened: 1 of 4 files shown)


AkmalFairuz (dismissed):

cmd/blockhash/main.go
```diff
@@ -133,6 +134,24 @@ func (b *hashBuilder) writeNextHash(w io.Writer) {
...
 
+// writeBlockHash writes the registry-independent hash used for material blocks.
+func (b *hashBuilder) writeBlockHash(w io.Writer) {
+	const helper = `
+// hashBlock packs a material block's type and state into the 32 bits reserved by generated hashes.
+// Each half uses 16 bits so the result does not depend on a registry's size or initialization.
+func hashBlock(b world.Block) uint64 {
+	base, state := b.Hash()
+	if base > 0xffff || state > 0xffff {
+		panic("material block hash exceeds 16-bit type or state")
```
TwistedAsylumMC: Can we print the block name here like before?

didntpot (approved):

PR #1459, opened 2026-09-14
server/listener.go: Implement UDP port range for NetherNet connections

lactyy2:
This PR adds a config field to limit the range of UDP ports used by the server to establish ICE connections with players.

Currently, the server allocates one UDP port for each connection, which might not be ideal for production servers. This PR changes the default to use only UDP port 19132 with ICE muxing.

Additionally, server operators can configure a range of UDP ports, such as `19132-20000`.


AkmalFairuz (approved):
LGTM!

